PR #2, opened 2019-04-17
README tweak

skaegi:


skaegi:
thanks this was also just a test to make sure the OWNERS file worked correctly -- which it did

PR #7, opened 2019-04-29
Update .gitignore

vtereso:
Removes dependencies to prevent huge diffs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@
 *.so
 *.dylib
 
+# dependencies
+node_modules
+vendor
+
 # Test binary, build with `go test -c`
 *.test
 


mnuttall:
Hi @vtereso please get your CLA sorted, and delete the leading /s as per @vdemeester's comment.

vtereso:
On it. Should we also remove the leading slashes on dashboard as well then? https://github.com/tektoncd/dashboard/blob/master/.gitignore#L14

vtereso:
I think it was an issue with the owners file previously because I didn't do anything as far as the CLA stuff, when I initially made this PR, everything seemed fine, I believe it was something with the owners file. Works for me.

PR #9, opened 2019-04-30
Add releasename as optional param (used with Helm pipeline)

a-roberts:
# Changes

For our webhook extension we should be able to support Helm pipelines too. This gets us a step closer by being able to do helm upgrades. 

The behaviour before this PR is that we generate Pipeline parameters which includes a release name that is passed into a Helm pipeline. The release name is built up of both the repository name and the short commit ID.

This means that subsequent commits result in entirely new Helm releases instead of upgrading the existing Helm release: because the names differ.

In this PR I've added a new optional field, `releasename`, with character length validation as Helm releases have a maximum length.

I've added a test for such validation, documented that it's a new field one can use, and tested it with a Helm pipeline and deploy task (I want to contribute said pipeline and task once I make the Tiller secret an optional parameter).

# Submitter Checklist

review them:

for more details._

Test code too 😄 

I've also removed two entirely used params: registrysecret and the repositorysecretname.

I'm changing it to just `releasename` based on feedback 😉

### DIFF
--- a/webhooks-extension/README.md
+++ b/webhooks-extension/README.md
@@ -21,6 +21,12 @@ To initiate this installation, run `development_install.sh`.
 
 _Note: Your git provider must be able to reach the address of this extension's sink. The sink is deployed with a Knative service, so you may need to configure Knative serving. We recommend [setting up a custom domain](https://knative.dev/v0.3-docs/serving/using-a-custom-domain/) with the extension `.nip.io`._
 
+## Running tests
+
+```bash
+docker build -f cmd/extension/Dockerfile_test .
+```
+
 ## API Definitions
 
 - [Extension API definitions](cmd/extension/README.md)
@@ -57,7 +63,12 @@ Reference the [Knative eventing GitHub source sample](https://knative.dev/docs/e
 - All knative event sources are created in the namespace into which the dashboard and this extension are installed.
 - Currently the docker registry to which built images are pushed is hard coded from the registry you specified at install time, there is work underway to change this restriction.
 - Only one webhook can be created for each git repository, so each repository will only be able to trigger a PipelineRun from one webhook.
-- Only the [simple-pipeline](https://github.com/pipeline-hotel/example-pipelines/blob/master/config/pipeline.yaml) Pipeline definition is currently supported.
+
+- Three pipeline definitions are currently supported.
+
+  - [simple-pipeline](https://github.com/pipeline-hotel/example-pipelines/blob/master/config/pipeline.yaml)
+  - [simple-helm-pipeline](https://github.com/pipeline-hotel/example-pipelines/blob/master/config/helm-pipeline.yaml) (requires a secret to talk to a secure Tiller)
+  - [simple-helm-pipeline-insecure](https://github.com/pipeline-hotel/example-pipelines/blob/master/config/helm-insecure-pipeline.yaml.yaml)
 
 ## Architecture information
 
--- a/webhooks-extension/endpoints/sink.go
+++ b/webhooks-extension/endpoints/sink.go
@@ -134,6 +134,8 @@ func createPipelineRunFromWebhookData(buildInformation BuildInformation, r Resou
 	pipelineTemplateName := webhook.Pipeline
 	pipelineNs := webhook.Namespace
 	saName := webhook.ServiceAccount
+	requestedReleaseName := webhook.ReleaseName
+
 	if saName == "" {
 		saName = "default"
 	}
@@ -186,7 +188,17 @@ func createPipelineRunFromWebhookData(buildInformation BuildInformation, r Resou
 
 	imageTag := buildInformation.SHORTID
 	imageName := fmt.Sprintf("%s/%s", dockerRegistry, strings.ToLower(buildInformation.REPONAME))
-	releaseName := fmt.Sprintf("%s-%s", strings.ToLower(buildInformation.REPONAME), buildInformation.SHORTID)
+
+	releaseName := ""
+
+	if requestedReleaseName != "" {
+		log.Printf("Release name based on input: %s", requestedReleaseName)
+		releaseName = requestedReleaseName
+	} else {
+		releaseName = fmt.Sprintf("%s", strings.ToLower(buildInformation.REPONAME))
+		log.Printf("Release name based on repository name: %s", releaseName)
+	}
+
 	repositoryName := strings.ToLower(buildInformation.REPONAME)
 	params := []v1alpha1.Param{{Name: "image-tag", Value: imageTag},
 		{Name: "image-name", Value: imageName},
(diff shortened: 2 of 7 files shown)


a-roberts:
@ncskier please take a look 😄

webhooks-extension/README.md
```diff
@@ -50,6 +56,10 @@ curl -d "${data}" -H "Content-Type: application/json" -X POST http://localhost:9
 
 Reference the [Knative eventing GitHub source sample](https://knative.dev/docs/eventing/samples/github-source/) to properly create the `accesstoken` secret. This is the secret that is used to create GitHub webhooks.
 
+If using Helm, you can also specify a Helm release name that enables you to do `helm install --upgrade` as expected, otherwise our sink will include the short Git commit ID in your Helm release name.
+
+Provide `releasename` (which must be less than 64 characters in length) in the POST request. This is an optional field.
```
ncskier: Since this sentence refers to the API definition, please move this sentence to the extension's api definition README: https://github.com/tektoncd/experimental/blob/master/webhooks-extension/cmd/extension/README.md#post-endpoints.

webhooks-extension/endpoints/webhook_test.go
```diff
@@ -96,6 +99,34 @@ func testGitHubSource(expectedName string, expectedOwnerAndRepo string, expected
...
+		RepositorySecretName: "secretName1",
+		ReleaseName:          "1234567891234567891234567891234567891234567891234567891234567890", // 0 brings us to 64 char
+	}
+
+	b, _ := json.Marshal(data)
+	httpReq := dummyHTTPRequest("POST", "http://wwww.dummy.com:8080/webhooks-extension/webhook/", bytes.NewBuffer(b))
+	req := dummyRestfulRequest(httpReq, "", "")
+	httpWriter := httptest.NewRecorder()
+	resp := dummyRestfulResponse(httpWriter)
+	r.createWebhook(req, resp)
```
ncskier: Please use the helper `createWebhook(webhook, resource)` function, instead of re-writing the code to create webhooks with dummy http requests & responses. This will minimize duplicate code.

Since your tests use the `resp`, you can update the `createWebhook` function to return the `resp`, and the tests that don't need it don't have to worry about what's being returned.

webhooks-extension/README.md
```diff
@@ -50,6 +56,10 @@ curl -d "${data}" -H "Content-Type: application/json" -X POST http://localhost:9
 
 Reference the [Knative eventing GitHub source sample](https://knative.dev/docs/eventing/samples/github-source/) to properly create the `accesstoken` secret. This is the secret that is used to create GitHub webhooks.
 
+If using Helm, you can also specify a Helm release name that enables you to do `helm install --upgrade` as expected, otherwise our sink will include the short Git commit ID in your Helm release name.
```
ncskier: Can you also please add a link to the helm pipeline that people can use with this extension. This could go in the [Limitations](https://github.com/tektoncd/experimental/tree/master/webhooks-extension#limitations) section with the [simple-pipeline](https://github.com/pipeline-hotel/example-pipelines/blob/master/config/pipeline.yaml) link?

a-roberts:
@ncskier thanks for the detailed review! Please have a look at https://github.com/pipeline-hotel/example-pipelines/pull/1/files as well where I've added the pipeline and task definitions for Helm. 

I would like to reduce duplication here (can we have optional parameters e.g. for the Helm secret?) and they are very specific for "straight through" deployment scenarios and hence the comments. Have a look at my PR for the restrictions so far.

Happy to work on improving that PR as well as this one, and we should merge these two together 😄

a-roberts:
Docs and pipeline PR at https://github.com/pipeline-hotel/example-pipelines/pull/1, this and the sink changes to be merged together (Brandon's looking at both)

webhooks-extension/endpoints/sink.go
```diff
@@ -187,17 +188,23 @@ func createPipelineRunFromWebhookData(buildInformation BuildInformation, r Resou
...
 	imageName := fmt.Sprintf("%s/%s", registryURL, strings.ToLower(buildInformation.REPONAME))
-	releaseName := fmt.Sprintf("%s-%s", strings.ToLower(buildInformation.REPONAME), buildInformation.SHORTID)
+	releaseName := ""
+
+	if wantedReleaseName != "" {
+		log.Printf("A specified release name %s has been provided", wantedReleaseName)
+		releaseName = wantedReleaseName
+	} else {
+		log.Printf("Generating release name based on repository name and short commit iD")
+		releaseName = fmt.Sprintf("%s-%s", strings.ToLower(buildInformation.REPONAME), buildInformation.SHORTID)
```
dibbles: So the question I would have here is an architectural one ..... if no release name is provided should it be the case that:

a) project name + commit is used so multiple helm installs are attempted (probably service name clashes)
b) project name is used without commit so upgrade works ... i.e default to a fixed name based on repo

Additional note - I would log the generated name, rather then logging one will be generated and not logging it

webhooks-extension/endpoints/sink.go
```diff
@@ -129,11 +129,12 @@ func createPipelineRunFromWebhookData(buildInformation BuildInformation, r Resou
 		log.Printf("Error getting github webhook: %s.", err.Error())
 		return
 	}
-	registrySecret := webhook.RegistrySecret
 	helmSecret := webhook.HelmSecret
 	pipelineTemplateName := webhook.Pipeline
 	pipelineNs := webhook.Namespace
 	saName := webhook.ServiceAccount
+	wantedReleaseName := webhook.ReleaseName
```
dibbles: dislike wantedReleaseName - would prefer requestedReleaseName

webhooks-extension/endpoints/webhook.go
```diff
@@ -42,6 +42,16 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
 		RespondError(response, err, http.StatusBadRequest)
 		return
 	}
+
+	if webhook.ReleaseName != "" {
+		if len(webhook.ReleaseName) > 63 {
+			err := errors.New("Wanted release name is too long, must be less than 64 characters")
```
dibbles: "Requested" rather than "Wanted" and possibly ad the "it"

webhooks-extension/endpoints/webhook.go
```diff
@@ -42,6 +42,16 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+			log.Printf("Error: %s.", err.Error())
```
dibbles: Wondering if we need an issue so we have different logging levels .... errors would normally start lower cased.

webhooks-extension/endpoints/webhook.go
dibbles: Lowercase at start of error message

webhooks-extension/endpoints/sink.go
a-roberts: I've added the generated name log, for the architectural bit, yeah I like b). Indeed with a) you're going to have service name clashes. I'll implement that.

a-roberts:
@dibbles thanks, I've just merged the latest in including your changes and will test 😄

a-roberts:
Ready for another review please, I tested this just now with another Helm repository and merged in the latest changes, I also documented the `dockerregistry` required field on the API

PR #18, opened 2019-05-03
Install Tekton dashboard, webhook-extension and prereqs

mnuttall:
# Changes

See test/README.md for details. This sets everything up ready for testing. Actual tests to follow in subsequent PRs. 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/test/cleanup.sh
+++ b/webhooks-extension/test/cleanup.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+export tekton_repo_dir=$(git rev-parse --show-toplevel)
+export test_dir="${tekton_repo_dir}/webhooks-extension/test"
+
+source ${test_dir}/config.sh
+source ${test_dir}/util.sh
+
+if [ $DASHBOARD_INSTALL_NS != "default" ]; then
+  kubectl delete ns ${DASHBOARD_INSTALL_NS}
+else
+  kubectl delete deployment tekton-dashboard -n $DASHBOARD_INSTALL_NS
+fi
+kubectl delete ns knative-eventing
+kubectl delete ns knative-serving
+kubectl delete ns istio-system
--- a/webhooks-extension/test/install_prereqs.sh
+++ b/webhooks-extension/test/install_prereqs.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# This script is for use on Docker Desktop on a Mac. It installs Istio, Knative and Tekton. It requires:
+#  - kubectl
+#  
+
+export tekton_repo_dir=$(git rev-parse --show-toplevel)
+export test_dir="${tekton_repo_dir}/webhooks-extension/test"
+
+source ${test_dir}/config.sh
+source ${test_dir}/util.sh
+  
+# Required for devops-back-end: istio, knative eventing, knative eventing sources, knative serving, and tekton
+install_istio_nodeport ${KNATIVE_VERSION}
+install_knative_serving_nodeport ${KNATIVE_VERSION}
+install_knative_eventing ${KNATIVE_VERSION}
+install_knative_eventing_sources ${KNATIVE_VERSION}
+install_tekton ${TEKTON_VERSION}
+
+# Docker desktop: cluster IP = host IP. Obviously not true for other types of cluster. 
+# For minikube, ip=$(minikube ip)
+ip=$(ifconfig | grep netmask | sed -n 2p | cut -d ' ' -f2) 
+kubectl patch configmap config-domain --namespace knative-serving --type='json' \
+  --patch '[{"op": "add", "path": "/data/'"${ip}.nip.io"'", "value": ""}]'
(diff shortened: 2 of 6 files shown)


webhooks-extension/test/cleanup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+export tekton_repo_dir=$(git rev-parse --show-toplevel)
+export test_dir="${tekton_repo_dir}/webhooks-extension/test"
+
+source ${test_dir}/config.sh
+source ${test_dir}/util.sh
+
+if [ $DASHBOARD_INSTALL_NS != "default" ]; then
```
vtereso: Is it ok to potentially delete the namespace with the deployment inside of it?

webhooks-extension/test/install_prereqs.sh
```diff
@@ -0,0 +1,24 @@
...
+# Required for devops-back-end: istio, knative eventing, knative eventing sources, knative serving, and tekton
+install_istio_nodeport ${KNATIVE_VERSION}
+install_knative_serving_nodeport ${KNATIVE_VERSION}
+install_knative_eventing ${KNATIVE_VERSION}
+install_knative_eventing_sources ${KNATIVE_VERSION}
+install_tekton ${TEKTON_VERSION}
+
+# Docker desktop: cluster IP = host IP. Obviously not true for other types of cluster. 
+# For minikube, ip=$(minikube ip)
+ip=$(ifconfig | grep netmask | sed -n 2p | cut -d ' ' -f2)
```
vtereso: We could maybe check `kubectl config current-context` to determine which ip to grab.

webhooks-extension/test/util.sh
```diff
@@ -0,0 +1,215 @@
...
+        echo "Missing [version]"
+        exit 1
+    fi
+    version="$1"
+    kubectl apply --filename https://github.com/knative/eventing/releases/download/${version}/release.yaml
+    # Wait until all the pods come up
+    wait_for_ready_pods knative-eventing 180 20
+}
+
+# Note that eventing-sources.yalm was renamed from release.yaml in the v0.5.0 release, so this won't work for earlier releases as-is.
```
vtereso: eventing-sources.yalm -> eventing-sources.yaml nit

webhooks-extension/test/util.sh
```diff
@@ -0,0 +1,215 @@
...
+    npm run build_ko
+    dep ensure -v
+    ko apply -f config -n $namespace
+    popd
+  fi
+}
+
+# Install https://github.com/tektoncd/experimental/tree/master/webhooks-extension
+function install_webhooks_extension() { 
+  pushd $GOPATH/src/github.com/tektoncd/experimental/webhooks-extension
```
vtereso: Didn't know about pushd/popd that's a nice utility

webhooks-extension/test/util.sh
```diff
@@ -0,0 +1,215 @@
...
+  dep ensure -v
+  ko apply -f config -n $namespace
+  popd
+  wait_for_ready_pods $namespace 60 10
+}
+
+# Wait until all pods in a namespace are Running or Complete
+# wait_for_ready [namespace] [timeout] <sleepTime>
+# <sleepTime> is optional
+function wait_for_ready_pods() {
```
vtereso: Would something like this be sufficient? https://github.com/openshift-cloud-functions/knative-operators/blob/master/etc/scripts/installation-functions.sh#L32
To wait until there are no pods that are NOT running/completed/terminating? Seems like a big function, but no real qualms otherwise.

webhooks-extension/test/util.sh
```diff
@@ -0,0 +1,215 @@
...
+    kubectl get pods --namespace $namespace
+}
+
+function check() { 
+  if [ -z "$1" ]; then
+    echo "Usage ERROR for function: check [varname] (example-value)"
+    echo "Missing [varname]"
+    exit 1
+  fi
+  vToCheck=${!1}
```
vtereso: Indirection is nicer than using eval that I have done in the past. I will be using this too.

webhooks-extension/test/cleanup.sh
mnuttall: Deleting the namespace deletes everything in that namespace as well.

webhooks-extension/test/install_prereqs.sh
mnuttall: I've added a comment about how we could use `kubectl cluster-info`. It returns 'localhost' for Docker Desktop which is a nuisance. If it's ok with you I'd rather we extended this logic as we use this code to support testing on other platforms, rather than speculatively.

webhooks-extension/test/util.sh
mnuttall: Yes they're great.

webhooks-extension/test/util.sh
mnuttall: Nice - thank you, I've added that in.

mnuttall: @vtereso thanks for your review, I've made the changes I think necessary so please lgtm if you're happy, thank you.

PR #22, opened 2019-05-09
Add delete endpoint for webhooks, optionally deleting PipelineRuns

a-roberts:
# Changes

Adds the ability to delete webhooks by name and if &deletepipelineruns=true is set, delete the PipelineRuns that are for the same Git repository URL. This is only going to work for PipelineRuns we create through the extension as we label them. These labels are what I use to do the comparison with.

I've also made the images Alpine to reduce their size (for our tests Dockerfiles), added tests and API docs.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/cmd/extension/README.md
+++ b/webhooks-extension/cmd/extension/README.md
@@ -55,6 +55,24 @@ Example POST
 }
 ```
 
+### DELETE endpoint
+
+```
+DELETE /webhooks/<webhookid>?namespace=<my namespace>
+```
+
+You can optionally add &deletepipelineruns=true to remove all PipelineRuns associated with the same repository.
+
+Returns HTTP code 201 if the webhook was deleted successfully
+Returns HTTP code 400 if an error occurred with the request body
+Returns HTTP code 404 if the webhook wasn't found
+Returns HTTP code 405 if a query parameter alone was provided
+Returns HTTP code 500 if any other errors occurred
+
+Deletes the GithubSource (therefore the webhook from the repository) and optionally deletes all PipelineRuns for the configured repository. 
+The ConfigMap used to maintain a list of configured webhooks to Pipelines is also updated.
+
+
 These endpoints can be accessed through the dashboard.
 
 If using Helm, you can also specify a Helm release name. If no Helm release name is provided, your Helm release name will default to be the repository name.
--- a/webhooks-extension/cmd/extension/README.md
+++ b/webhooks-extension/cmd/extension/README.md
@@ -59,7 +59,6 @@ Example POST
 
 ```
 DELETE /webhooks/<webhookid>?namespace=<my namespace>
-```
 
 You can optionally add &deletepipelineruns=true to remove all PipelineRuns associated with the same repository.
 
@@ -71,6 +70,7 @@ Returns HTTP code 500 if any other errors occurred
 
 Deletes the GithubSource (therefore the webhook from the repository) and optionally deletes all PipelineRuns for the configured repository. 
 The ConfigMap used to maintain a list of configured webhooks to Pipelines is also updated.
+```
 
 
 These endpoints can be accessed through the dashboard.
--- a/webhooks-extension/endpoints/webhook.go
+++ b/webhooks-extension/endpoints/webhook.go
@@ -29,8 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var statusLock sync.Mutex
-var createLock sync.Mutex
+var modifyingConfigMapLock sync.Mutex
 
 // RegisterEndpoints registers the endpoints for a container
 // Do it this way so we can create our own test server too
@@ -52,8 +51,8 @@ func (r Resource) RegisterEndpoints(container *restful.Container) {
 
 // Creates a webhook for a given repository and populates (creating if doesn't yet exist) a ConfigMap storing this information
 func (r Resource) createWebhook(request *restful.Request, response *restful.Response) {
-	createLock.Lock()
-	defer createLock.Unlock()
+	modifyingConfigMapLock.Lock()
+	defer modifyingConfigMapLock.Unlock()
 
 	logging.Log.Infof("Creating webhook with request: %+v.", request)
 	installNs := r.Defaults.Namespace
@@ -162,6 +161,8 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
 
 // Removes from ConfigMap, removes the actual GitHubSource, removes the webhook
 func (r Resource) deleteWebhook(request *restful.Request, response *restful.Response) {
+	modifyingConfigMapLock.Lock()
+	defer modifyingConfigMapLock.Unlock()
 	logging.Log.Debug("In deleteWebhook")
 	name := request.PathParameter("name")
 	namespace := request.QueryParameter("namespace")
@@ -248,8 +249,6 @@ func (r Resource) findRepoURLFromConfigMap(webhookName, namespace string) string
 func (r Resource) deleteWebhookFromConfigMapByName(webhookName, namespace string) error {
 	// We don't want any other requests going through here so use a mutex
 	// otherwise we'll end up with operations performed on old configmaps so we'll have invalid data
-	statusLock.Lock()
-	defer statusLock.Unlock()
 	logging.Log.Debug("Deleting webhook info from ConfigMap")
 
 	configMapClient := r.K8sClient.CoreV1().ConfigMaps(namespace)
--- a/webhooks-extension/endpoints/webhook.go
+++ b/webhooks-extension/endpoints/webhook.go
@@ -247,8 +247,6 @@ func (r Resource) findRepoURLFromConfigMap(webhookName, namespace string) string
 
 // Delete the webhook information from our ConfigMap
 func (r Resource) deleteWebhookFromConfigMapByName(webhookName, namespace string) error {
-	// We don't want any other requests going through here so use a mutex
-	// otherwise we'll end up with operations performed on old configmaps so we'll have invalid data
 	logging.Log.Debug("Deleting webhook info from ConfigMap")
 
 	configMapClient := r.K8sClient.CoreV1().ConfigMaps(namespace)
--- a/webhooks-extension/endpoints/webhook.go
+++ b/webhooks-extension/endpoints/webhook.go
@@ -212,7 +212,7 @@ func (r Resource) deleteWebhook(request *restful.Request, response *restful.Resp
 		RespondError(response, err, http.StatusBadRequest)
 		return
 	}
-	response.WriteHeader(201)
+	response.WriteHeader(200)
 }
 
 // Find the repository URL for a webhook. We store webhook in a ConfigMap
--- a/webhooks-extension/endpoints/webhook.go
+++ b/webhooks-extension/endpoints/webhook.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	logging "github.com/tektoncd/experimental/webhooks-extension/pkg/logging"
 	"net/http"
+	"strconv"
 	"strings"
 
 	restful "github.com/emicklei/go-restful"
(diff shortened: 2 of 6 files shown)


a-roberts:
@ncskier @dibbles can you please take a look?

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+
+	var toDeletePipelineRuns = false
+	var err error
+
+	if deletePipelineRuns != "" {
+		toDeletePipelineRuns, err = strconv.ParseBool(deletePipelineRuns)
+		if err != nil {
+			theError := errors.New("bad request information provided, cannot handle deletepipelineruns query (should be set to true or not provided)")
+			logging.Log.Error(theError)
+			RespondError(response, err, http.StatusInternalServerError)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+		}
+	}
+
+	if namespace == "" {
+		theError := errors.New("no namespace was provided")
+		logging.Log.Error(theError)
+		RespondError(response, theError, http.StatusBadRequest)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	}
+
+	if name != "" {
+		foundRepoURL := r.findRepoURLFromConfigMap(name, namespace)
+		// This will remove any PipelineRuns too if specified to, hence the need for the repository URL
+		err = r.deleteGitHubWebhookByName(name, namespace, foundRepoURL, toDeletePipelineRuns)
+		if err != nil {
+			if strings.Contains(err.Error(), "could not find webhook with name") {
+				logging.Log.Errorf("webhook (name %s) not found in namespace %s", name, namespace)
+				RespondError(response, err, http.StatusNotFound)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+			} else {
+				logging.Log.Errorf("error deleting the webhook (name %s), error: %s", name, err)
+				RespondError(response, err, http.StatusInternalServerError)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+			}
+		} else {
+			logging.Log.Info("Deleted the webhook OK, deleting from ConfigMap next")
+		}
+
+		err = r.deleteWebhookFromConfigMapByName(name, namespace)
+		if err != nil {
+			logging.Log.Errorf("error deleting the webhook information (name %s) from the ConfigMap, error: %s", name, err)
+			RespondError(response, err, http.StatusInternalServerError)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+		}
+	} else {
+		logging.Log.Error("no name was provided")
+		RespondError(response, err, http.StatusBadRequest)
```
ncskier: I think this should `return` after responding with an error message

webhooks-extension/endpoints/webhook_test.go
```diff
@@ -194,13 +209,273 @@ func testGetAllWebhooks(expectedWebhooks []webhook, r *Resource, t *testing.T) {
...
+	httpReq, _ := http.NewRequest(http.MethodDelete, server.URL+"/webhooks/?namespace=foo", nil)
+	response, _ := http.DefaultClient.Do(httpReq)
+	if response.StatusCode != 405 {
+		t.Errorf("Status code not set to 400 when deleting without a name, it's: %d", response.StatusCode)
+	}
+}
+
+func TestDeleteByNameNoNamespaceBadRequest(t *testing.T) {
+	setUpServer()
+	httpReq, _ := http.NewRequest(http.MethodDelete, server.URL+"/webhooks/name=foo", nil)
```
ncskier: I think `webhooks/name=foo` should be `webhooks/foo` because the name is a path parameter.

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	if namespace == "" {
+		theError := errors.New("no namespace was provided")
+		logging.Log.Error(theError)
+		RespondError(response, theError, http.StatusBadRequest)
+	}
+
+	if name != "" {
+		foundRepoURL := r.findRepoURLFromConfigMap(name, namespace)
+		// This will remove any PipelineRuns too if specified to, hence the need for the repository URL
+		err = r.deleteGitHubWebhookByName(name, namespace, foundRepoURL, toDeletePipelineRuns)
```
ncskier: If I'm not mistaken, all of the Github webhooks are created in the install namespace (this is because they need to be in the same namespace as their sink). So, I think this `deleteGitHubWebhookByName` should exclusively delete from the install namespace.

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	/* Contains for example, base64 encoded:
+	{
+		confiMap.BinaryData[GitHubSource]
+		"myhook":
+		{
+			"name":"myhook",
+			"namespace":"default",
+			"gitrepositoryurl":"https://myrepourl",
+		  "accesstoken":"my-secret-for-hooks",
+		  "pipeline":"simple-helm-pipeline-insecure",
```
ncskier: nit: please indent these two lines to match the others.

webhooks-extension/cmd/extension/README.md
```diff
@@ -55,6 +55,24 @@ Example POST
 }
 ```
 
+### DELETE endpoint
+
+```
+DELETE /webhooks/<webhookid>?namespace=<my namespace>
```
dibbles: interesting design question here - do we want namespace as a query param or as a path param ..... namespace/{namespace}/webhooks/{webhookid} ...... although we quite like having the /webhooks start ... so maybe /webhooks/namespace/{namespace}/webhooks/{webhook} although that seems long winded ... maybe query param is the nicest.

we get into quite a deep conversation about differences between the dashboard and the extension ... the dashboard being designed so that you choose a namespace and then all controls are on that namespace .... vs the extension taking namespace as selection in many places.  Possibly we should discuss as a wider group

webhooks-extension/endpoints/webhook.go
dibbles: I would partly argue for leaving in the namespace arg as giving more future potential ..... it also allows for deleting when a sink is in a different namespace to the install namespace??  Depends on what install namespace is defined as .... presumably the install namespace of the dashboard

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+		if err != nil {
+			if strings.Contains(err.Error(), "could not find webhook with name") {
+				logging.Log.Errorf("webhook (name %s) not found in namespace %s", name, namespace)
+				RespondError(response, err, http.StatusNotFound)
+			} else {
+				logging.Log.Errorf("error deleting the webhook (name %s), error: %s", name, err)
+				RespondError(response, err, http.StatusInternalServerError)
+			}
+		} else {
+			logging.Log.Info("Deleted the webhook OK, deleting from ConfigMap next")
```
dibbles: %s webhook name to identify which one was deleted ..... might also want to add similar to method entry after reading values.

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	err = json.Unmarshal(data, &itsANestedMap)
+	if err != nil {
+		logging.Log.Errorf("invalid data format in the ConfigMap, can't modify it, error: %s", err.Error())
+	}
+
+	delete(itsANestedMap, webhookName)
+
+	itsANestedMapAsBytes, err := json.Marshal(itsANestedMap)
+	if err != nil {
+		logging.Log.Errorf("Couldn't go back to bytes from the nested map, error is: %s", err.Error())
```
dibbles: nit: lower case error message

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +148,142 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	}
+
+	configMap.BinaryData["GitHubSource"] = itsANestedMapAsBytes
+
+	_, err = configMapClient.Update(configMap)
```
dibbles: discussion: does there need to be some locking around updates to the config map?

webhooks-extension/cmd/extension/README.md
a-roberts: I'll leave in the query namespace for now for future proofing, we can behave a little differently from the main dashboard with our own API definitions and, like you, I'm quite a fan of the namespace being a query parameter, so I'm going to leave it as is. I don't really want to change everything to be, for example, /v1/namespaced, at least not yet, until we have a proper namespace and consistency story going forward

webhooks-extension/endpoints/webhook.go
a-roberts: I'll leave in for future potential as we want to work with various namespaces eventually

webhooks-extension/endpoints/webhook.go
a-roberts: I agree, added a test case, need to fix implementation next.

If two users are attempting to update the configmap at the same time, we may end up with the wrong data (let's say Adam does a delete to remove his webhook as Duane does the same thing to remove theirs, and we're behaving particularly slow). Duane's update is going to work on the old ConfigMap's data.

https://golang.org/src/net/http/client.go the http client itself that we use mentions being thread safe...

I've added a simple test case using a WaitGroup with two goroutines, both doing requests to delete different webhooks, and then we check they were really both deleted OK (that is, one didn't act upon some invalid data). This should simulate two HTTP requests interacting with the same resource. I thought all was well but on wrapping said test in a for loop and doing it 100 times, there's a 1/10 chance (observed so far!) that sometimes not all of the data is removed successfully. So I'll add a lock.

a-roberts:
@dibbles @ncskier please take another look, approve and lgtm if you're happy, I've addressed the feedback - keeping in the existing namespace behaviour for future proofing, adding a test to ensure it's thread safe (it wasn't, added a mutex)

ncskier:
Good catch with the lock @dibbles. @a-roberts I think we'll also need to lock here: https://github.com/tektoncd/experimental/blob/master/webhooks-extension/endpoints/webhook.go#L122-L129 to protect the read, modify, write code.

a-roberts:
Added, and I've also made the response be http 200 instead of 201 created

webhooks-extension/endpoints/webhook.go
```diff
@@ -19,17 +19,43 @@ import (
...
+	"sync"
 
 	restful "github.com/emicklei/go-restful"
 	eventapi "github.com/knative/eventing-sources/pkg/apis/sources/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var statusLock sync.Mutex
+var createLock sync.Mutex
```
ncskier: Thanks for adding the extra lock, but I think there has to be one lock for both of them instead of two locks 😄 Otherwise, we still have the same problem if a CREATE and DELETE operation happen at the same time.

a-roberts: I agree, so I've renamed it to modifyingConfigMapLock and I reference the same lock in both the delete and create functions now (I'm happy `defer` is a thing and hoping I needn't manually manage the lock with this approach).

I added another two test cases.

One for doing lots of creations at once with multiple routines to check the ConfigMap integrity.

Another is for create and delete.

@ncskier please take another look.

ncskier: Oh, wow, that is a lot of logs 😅 I'm not sure how to limit the logs either. If it's not worth removing that `Debugf` call, then I think it's alright. I like that the tests are covering all of these sync issues 👍

webhooks-extension/endpoints/webhook.go
```diff
@@ -126,12 +154,153 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
...
+	}
+
+	foundURL = itsANestedMap[webhookName]["gitrepositoryurl"]
+	return foundURL
+}
+
+// Delete the webhook information from our ConfigMap
+func (r Resource) deleteWebhookFromConfigMapByName(webhookName, namespace string) error {
+	// We don't want any other requests going through here so use a mutex
+	// otherwise we'll end up with operations performed on old configmaps so we'll have invalid data
```
ncskier: I think these comments are leftover from before you moved the lock code. They can probably be moved/removed now?

webhooks-extension/endpoints/webhook.go
a-roberts: IKR, it's good to know what the webhooks are at the exact point of doing this call though.

I'd really prefer it if we could just turn off the logger and I don't want less iterations for obvious reasons (sometimes the tests will pass by chance).

Alternatively, sending stdout to /dev/null for these go tests as I've tried here. We test with `go test -v` and if I remove that verbose flag it's like you never ran the tests.

ncskier:
@a-roberts I just re-tested your code, and it worked well 🙂Just have that last request about the stale comments, and should be good to go!

webhooks-extension/endpoints/webhook.go
a-roberts: Yep can go, done

ncskier:
Thanks Adam!
/lgtm

PR #24, opened 2019-05-14
Fully scripted webhooks test

mnuttall:
# Changes

Scripting and documentation for an end-to-end test of webhooks support. 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/test/README.md
+++ b/webhooks-extension/test/README.md
@@ -20,3 +20,92 @@ This is a work in progress and will take a while to settle down.
 - check that GOPATH is set
 - `export KO_DOCKER_REPO=docker.io/[your_dockerhub_id]`
 - `test/install_dashboard_and_extension.sh`
+
+## Test with a webhook
+
+As we state in config.sh, 
+```
+# Note that to receive webhooks, your github must be able to http POST to your Tekton installation. 
+# Our initial testing has used Docker Desktop and GitHub Enterprise. 
+```
+
+Webhooks are outbound HTTP requests from (in this case GitHub) to your Kubernetes environment. If you are behind a firewall it's unlikely that github.com will be able to reach you. So the two most common testing scenarios are, 
+- An in-house GitHub Enterprise to your Docker Desktop
+- github.com to your internet-facing kubernetes cluster in a commercial public cloud environment
+
+The checked-in defaults current reflect the first of these two options. To test with a webhook you need to do some setup work. 
+
+### Credentials 
+Edit `credentials.sh` from 
+```
+DOCKERHUB_USERNAME=[your-dockerhub-id]
+DOCKERHUB_PASSWORD=[your-docker-hub-password]
+GITHUB_USERNAME=[your-github-login-id]
+GITHUB_TOKEN=[github-token-with-wehooks-permissions]
+```
+to something of the form, 
+```
+DOCKERHUB_USERNAME=mnuttall
+DOCKERHUB_PASSWORD=thisIsNotMyPassword
+GITHUB_USERNAME=mnuttall
+GITHUB_TOKEN=fbufliwufbe4wliufiuwebfliseubfweiluf
+```
+
+### Configuration
+Edit the GITHUB settings in `config.sh`: 
+```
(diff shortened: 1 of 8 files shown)


webhooks-extension/test/README.md
```diff
@@ -19,4 +19,93 @@ This is a work in progress and will take a while to settle down.
...
+DOCKERHUB_USERNAME=mnuttall
+DOCKERHUB_PASSWORD=thisIsNotMyPassword
+GITHUB_USERNAME=mnuttall
+GITHUB_TOKEN=fbufliwufbe4wliufiuwebfliseubfweiluf
+```
+
+### Configuration
+Edit the GITHUB settings in `config.sh`: 
+```
+# Set this to your github - used to create wehooks
```
a-roberts: Nit, webhooks

webhooks-extension/test/cleanup.sh
```diff
@@ -9,8 +9,11 @@ source ${test_dir}/util.sh
 if [ $DASHBOARD_INSTALL_NS != "default" ]; then
   kubectl delete ns ${DASHBOARD_INSTALL_NS}
 else
-  kubectl delete deployment tekton-dashboard -n $DASHBOARD_INSTALL_NS
+  kubectl delete deployment tekton-dashboard -n default
+  kubectl delete deployment webhooks-extension -n default
```
a-roberts: Do the kservices go too? How about the configmap called githubwebhook or any existing githubsource CRDs, or is this a "run me once" type thing?

a-roberts: Discussed with Mark, he found it useful while writing this - while it's not the most thorough of cleanups (e.g. you'd end up with lingering webhooks perhaps if you kept running it), nobody else is actually calling this script and it's more there for convenience. I'd rather we keep it as it is (so it's partially useful at least albeit not 100% rock solid), so resolving this convo

webhooks-extension/test/cleanup.sh
```diff
@@ -9,8 +9,11 @@ source ${test_dir}/util.sh
 if [ $DASHBOARD_INSTALL_NS != "default" ]; then
   kubectl delete ns ${DASHBOARD_INSTALL_NS}
 else
-  kubectl delete deployment tekton-dashboard -n $DASHBOARD_INSTALL_NS
+  kubectl delete deployment tekton-dashboard -n default
```
vtereso: `$DASHBOARD_INSTALL_NS` rather than default?

webhooks-extension/test/cleanup.sh
vtereso: Same as above.

webhooks-extension/test/setup_webhook_simple_test.sh
```diff
@@ -0,0 +1,89 @@
...
+
+check GOPATH
+if [ ! -f $GOPATH/src/github.com/tektoncd/experimental/webhooks-extension/test/credentials.sh ]; then 
+  echo "${GOPATH}/src/github.com/tektoncd/experimental/webhooks-extension/test must exist and contain adequate config. "
+  exit 1
+fi
+pushd $GOPATH/src/github.com/tektoncd/experimental/webhooks-extension/test
+. ./credentials.sh
+
+kubectl apply -f dashboard-service.yaml
```
vtereso: For this and the other 3 apply, should we add `-n $DASHBOARD_INSTALL_NS`?

webhooks-extension/test/cleanup.sh
mnuttall: No it's explicitly default because it's in the 'else if namespace==default' section.

webhooks-extension/test/cleanup.sh
mnuttall: As above, ns is explicitly default. The whole cleanup script is not called from anywhere, and not documented: it was for my convenience only.

webhooks-extension/test/setup_webhook_simple_test.sh
mnuttall: yes, definitely - thank you

PR #30, opened 2019-05-22
REST API for accesstokens

mnuttall:
# Changes

As part of https://github.com/tektoncd/dashboard/issues/160 I'm moving support for secrets that are access tokens out of dashboard into webhooks-extension. This PR adds the support to webhooks extension. A second PR will prune down credentials.go in dashboard. 

# Submitter Checklist

review them:

- [] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/webhooks-extension/endpoints/credentials.go
+++ b/webhooks-extension/endpoints/credentials.go
@@ -1,0 +1,236 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+		http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+	"fmt"
+	"math/rand"
+	"net/http"
+	"time"
+
+	restful "github.com/emicklei/go-restful"
+	logging "github.com/tektoncd/experimental/webhooks-extension/pkg/logging"
+	"github.com/tektoncd/experimental/webhooks-extension/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// 'credentials' from the webhooks-extension's point of view, are access tokens. That's the only sort we handle right now.
+type credential struct {
+	Name        string `json:"name"`
+	AccessToken string `json:"accesstoken"`
+	SecretToken string `json:"secrettoken,omitempty"`
+}
+
+/*--------------------------------------
+This file implements three endpoints from webhooks.go:
+	ws.Route(ws.POST("/credentials").To(r.createCredential))
+	ws.Route(ws.GET("/credentials").To(r.getAllCredentials))
+	ws.Route(ws.DELETE("/credentials/{name}").To(r.deleteCredential))
+---------------------------------------*/
+
+func (r Resource) createCredential(request *restful.Request, response *restful.Response) {
+	cred := credential{}
+	if err := getQueryEntity(&cred, request, response); err != nil {
+		return
+	}
+	if !r.verifyCredentialParameters(cred, response) {
+		return
+	}
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace == "" {
+		targetNamespace = r.Defaults.Namespace
+	}
+	if !r.namespaceExists(targetNamespace, response) {
+		return
+	}
+	secret := credentialToSecret(cred, targetNamespace, response)
+	if _, err := r.K8sClient.CoreV1().Secrets(targetNamespace).Create(secret); err != nil {
+		errorMessage := fmt.Sprintf("error creating secret in K8sClient: %s", err.Error())
+		utils.RespondMessageAndLogError(response, err, errorMessage, http.StatusBadRequest)
+		return
+	}
+	writeResponseLocation(request, response, cred.Name)
+}
+
+func (r Resource) deleteCredential(request *restful.Request, response *restful.Response) {
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace == "" {
+		targetNamespace = r.Defaults.Namespace
+	}
+	if !r.namespaceExists(targetNamespace, response) {
+		return
+	}
+	credName := request.PathParameter("name")
+	if !r.verifySecretExists(credName, targetNamespace, response) {
+		return
+	}
+	err := r.K8sClient.CoreV1().Secrets(targetNamespace).Delete(credName, &metav1.DeleteOptions{})
+	if err != nil {
+		errorMessage := fmt.Sprintf("error deleting secret from K8sClient: %s.", err.Error())
+		utils.RespondMessageAndLogError(response, err, errorMessage, http.StatusInternalServerError)
+		return
+	}
+}
+
+func (r Resource) getAllCredentials(request *restful.Request, response *restful.Response) {
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace == "" {
+		targetNamespace = r.Defaults.Namespace
+	}
+	if !r.namespaceExists(targetNamespace, response) {
+		return
+	}
+
+	// Get secrets from the resource K8sClient
+	secrets, err := r.K8sClient.CoreV1().Secrets(targetNamespace).List(metav1.ListOptions{})
+
+	if err != nil {
+		errorMessage := fmt.Sprintf("error getting secrets from K8sClient: %s.", err.Error())
+		response.WriteErrorString(http.StatusInternalServerError, errorMessage)
+		logging.Log.Error(errorMessage)
+		return
+	}
+
+	// Parse K8s secrets to credentials
+	creds := []credential{}
+	emptyCred := credential{}
+	for _, secret := range secrets.Items {
+		cred := secretToCredential(&secret, true)
+		if cred != emptyCred {
+			creds = append(creds, cred)
+			logging.Log.Infof("getAllCredentials Found credential %+v\n", cred)
+		}
+	}
+
+	logging.Log.Infof("getAllCredentials returning +%v", creds)
+
+	// Write the response
+	response.AddHeader("Content-Type", "application/json")
+	response.WriteEntity(creds)
+}
+
+// Sends error message 404 if the secret does not exist in the resource K8sClient
+func (r Resource) verifySecretExists(secretName string, namespace string, response *restful.Response) bool {
+	_, err := r.K8sClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
+	if err != nil {
+		errorMessage := fmt.Sprintf("error getting secret from K8sClient: '%s'.", secretName)
+		utils.RespondMessageAndLogError(response, err, errorMessage, http.StatusNotFound)
+		return false
+	}
+	return true
+}
+
+// Convert credential struct into K8s secret struct
+func credentialToSecret(cred credential, namespace string, response *restful.Response) *corev1.Secret {
+	// Create new secret struct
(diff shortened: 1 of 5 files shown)


mnuttall:
@dibbles please can you take a look since I know you're waiting for this?

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+func (r Resource) createCredential(request *restful.Request, response *restful.Response) {
+	cred := credential{}
+	if err := getQueryEntity(&cred, request, response); err != nil {
+		return
+	}
+	if !r.verifyCredentialParameters(cred, response) {
+		return
+	}
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace != "" {
```
dibbles: this should be if equal rather than not equal

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+		if targetNamespace == "" {
+			targetNamespace = "default"
+		}
+	}
+	if !r.verifyNamespaceExists(targetNamespace, response) {
+		return
+	}
+	secret := credentialToSecret(cred, targetNamespace, response)
+	if _, err := r.K8sClient.CoreV1().Secrets(targetNamespace).Create(secret); err != nil {
+		errorMessage := fmt.Sprintf("Error creating secret in K8sClient: %s", err.Error())
```
dibbles: go errors should start with a lowercase letter as they might get prefixed.

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/* Delete a Secret by name.
+ * Required path parameters:
+ *  - Namespace
+ *  - Name
+ */
+func (r Resource) deleteCredential(request *restful.Request, response *restful.Response) {
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace != "" {
```
dibbles: again - this should be an equals check

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+	if !r.verifyNamespaceExists(targetNamespace, response) {
+		return
+	}
+	credName := request.PathParameter("name")
+	if !r.verifySecretExists(credName, targetNamespace, response) {
+		return
+	}
+	_, err := r.K8sClient.CoreV1().Secrets(targetNamespace).Get(credName, metav1.GetOptions{})
+	if err != nil {
+		errorMessage := fmt.Sprintf("Error getting secret from K8sClient: %s.", err.Error())
```
dibbles: error message : lowercase start

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+	if !r.verifyNamespaceExists(targetNamespace, response) {
+		return
+	}
+	credName := request.PathParameter("name")
+	if !r.verifySecretExists(credName, targetNamespace, response) {
+		return
+	}
+	_, err := r.K8sClient.CoreV1().Secrets(targetNamespace).Get(credName, metav1.GetOptions{})
```
dibbles: I'm wondering if we need to check the revision ..... someone might delete and recreate with the same name before i press the delete button in the ui?  Might be slight overkill right now and would mean the front end needs to pass the revision.  Open to discussion on this one.

Additionally....

1. why would we do this after you have just done a verifySecretExists? surely we are just repeating ourselves.
2. If you had a reason to do it, call the secretExists function

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		errorMessage := fmt.Sprintf("Error getting secret from K8sClient: %s.", err.Error())
+		utils.RespondMessageAndLogError(response, err, errorMessage, http.StatusInternalServerError)
+		return
+	}
+	err = r.K8sClient.CoreV1().Secrets(targetNamespace).Delete(credName, &metav1.DeleteOptions{})
+	if err != nil {
+		errorMessage := fmt.Sprintf("Error deleting secret from K8sClient: %s.", err.Error())
```
dibbles: error message : lowercase start

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+		utils.RespondMessageAndLogError(response, err, errorMessage, http.StatusInternalServerError)
+		return
+	}
+}
+
+// Sends error messages if the namespace does not exist in the resource K8sClient
+func (r Resource) verifySecretExists(secretName string, namespace string, response *restful.Response) bool {
+	if err := r.secretExists(secretName, namespace); err != nil {
+		errorMessage := fmt.Sprintf("Error getting secret from K8sClient: '%s'.", secretName)
```
dibbles: error message : lowercase start

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+
+// Returns true if the secret exists in the resource K8sClient and false if it does not exist
+func (r Resource) secretExists(secretName string, namespace string) error {
+	_, err := r.K8sClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
+	return err
+}
+
+func (r Resource) getAllCredentials(request *restful.Request, response *restful.Response) {
+	targetNamespace := request.QueryParameter("namespace")
+	if targetNamespace != "" {
```
dibbles: this should be an equals rather than not equals

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if !r.verifyNamespaceExists(targetNamespace, response) {
+		return
+	}
+
+	// Get secrets from the resource K8sClient
+	secrets, err := r.K8sClient.CoreV1().Secrets(targetNamespace).List(metav1.ListOptions{})
+
+	if err != nil {
+		errorMessage := fmt.Sprintf("Error getting secrets from K8sClient: %s.", err.Error())
```
dibbles: error message : lowercase start

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+	if secret.Data["accessToken"] != nil {
+		cred = credential{
+			Name:            secret.GetName(),
+			AccessToken:     string(secret.Data["accessToken"]),
+			SecretToken:     string(secret.Data["secretToken"]),
+			ResourceVersion: secret.GetResourceVersion(),
+		}
+		if mask {
+			cred.AccessToken = "********"
+		}
```
dibbles: would we also want to mask the secret token??

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+
+// Convert K8s secret struct into credential struct
+func secretToCredential(secret *corev1.Secret, mask bool) credential {
+	var cred credential
+	if secret.Data["accessToken"] != nil {
+		cred = credential{
+			Name:            secret.GetName(),
+			AccessToken:     string(secret.Data["accessToken"]),
+			SecretToken:     string(secret.Data["secretToken"]),
+			ResourceVersion: secret.GetResourceVersion(),
```
dibbles: if we aren't doing updates do we need resource version on credential?

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	return nil
+}
+
+func (r Resource) verifyCredentialParameters(cred credential, response *restful.Response) bool {
+	errorMessage := ""
+	if cred.Name == "" {
+		errorMessage = fmt.Sprintf("Error: Name must be specified")
+	} else if cred.AccessToken == "" {
+		errorMessage = fmt.Sprintf("Error: AccessToken must be supplied")
```
dibbles: inconsistency in this message and that on 225 .... I'd use specified in both cases.

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+func writeResponseLocation(request *restful.Request, response *restful.Response, identifier string) {
+	location := request.Request.URL.Path
+	if request.Request.Method == http.MethodPost {
+		location = location + "/" + identifier
+	}
+	response.AddHeader("Content-Location", location)
+	response.WriteHeader(201)
+}
+
+// Returns true if the namespace exists in the resource K8sClient and false if it does not exist
```
dibbles: change the comment - as this isn't what the function returns

webhooks-extension/endpoints/credentials.go
```diff
@@ -0,0 +1,261 @@
...
+func (r Resource) namespaceExists(namespace string) error {
+	_, err := r.K8sClient.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+	return err
+}
+
+// Sends error messages if the namespace does not exist in the resource K8sClient
+func (r Resource) verifyNamespaceExists(namespace string, response *restful.Response) bool {
+	if err := r.namespaceExists(namespace); err != nil {
+		errorMessage := fmt.Sprintf("Namespace provided does not exist: '%s'.", namespace)
```
dibbles: lowercase error message?

webhooks-extension/endpoints/credentials_test.go
```diff
@@ -0,0 +1,243 @@
...
+		Name:        "accesstokenToDelete",
+		AccessToken: "sdkfhighregusfbliusbbbwhfwiehw8hwefhw938hf497fhw97b47",
+	}
+	createAndCheckCredential(namespace, accessTokenToDelete, "", r, t)
+
+	httpReq := dummyHTTPRequest("GET", "http://wwww.dummy.com:8383/credentials/accesstokenToDelete?namespace=ns2", bytes.NewBuffer(nil))
+	req := dummyRestfulRequest(httpReq, namespace, "")
+	httpWriter := httptest.NewRecorder()
+	resp := dummyRestfulResponse(httpWriter)
+	r.deleteCredential(req, resp)
```
dibbles: do you want to add in a get and check the secret is no longer listed?

webhooks-extension/endpoints/credentials_test.go
```diff
@@ -0,0 +1,243 @@
...
+	expectedError := fmt.Sprintf("Error: AccessToken must be supplied")
+	createAndCheckCredential(namespace, badAccessToken, expectedError, r, t)
+
+	// Verify no credentials have been created
+	checkCredentials(namespace, []credential{}, "", r, t)
+}
+
+func TestAccessTokenWithSecret(t *testing.T) {
+	r := dummyResource()
+	namespace := "default"
```
dibbles: I would suggest adding in another test similar but with non default namespace to ensure namespaces are as expected.

webhooks-extension/endpoints/credentials.go
mnuttall: http DELETE doesn't take a body so we'd have to pass the revision as a query param. Shall we have that conversation in dashboard first, and then add ?revision here if we also decide to do it in dashboard?

webhooks-extension/endpoints/credentials.go
mnuttall: deleted for now.

webhooks-extension/endpoints/credentials.go
mnuttall: I thought we might want to display it, particularly in the case where we generated it, in case it's needed for debugging. I also wondered about not hiding the first few characters of each accessToken to help users know which was which. It's a shame it's called 'secretToken' by the knative GitHub event source and not 'handshake token' : )

webhooks-extension/endpoints/credentials.go
mnuttall: I've merged the functions.

webhooks-extension/endpoints/credentials_test.go
mnuttall: good catch, done, found another test code bug in doing so :)

webhooks-extension/endpoints/credentials_test.go
mnuttall: doh! Done.

PR #52, opened 2019-06-05
Rough User Interface for Creating and Listing Webhooks

dibbles:
# Changes

User interface for the webhooks extension - admittedly this is quite rough but better to start with something and fix it up.

# Submitter Checklist

review them:

- [] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._


dibbles:
@jessm12 please look at this ..... I know it all needs squashing and bringing up to date etc etc... but would be good to have a look over for glaring issues right now.

I'm still working on the tests and also have this list of toDos (minimum) ....

- Architecture decision, should namespace menu on left affect the web hook display table
- Spinner on table load should be table skeleton
- Carbon guidelines / Accessibility etc
- Change validation to be on click of create, use red boxes (Webhook)
   - Validate display name chars (steal from main dashboard)
   - Validate URL entry (steal from main dashboard)
   - Validate docker registry field ?
- Change validation to be on click of create, use red boxes (Access Token)
- Fix password area (couple of characters behind eye)
- Tab navigation around panels
- Better add button or just Icon 
- Namespace should not be query parameter on POST
- Close button on notifications

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -0,0 +1,218 @@
...
+          },
+          {
+            key: 'pipeline',
+            header: 'Pipeline',
+          },
+          {
+            key: 'namespace',
+            header: 'Namespace',
+          }
+        ];
```
jessm12: Are these keys and headers necessary when the header and key are the same, could you use just a list of header names?

webhooks-extension/src/components/WebhookCreate/WebhookCreate.scss
```diff
@@ -0,0 +1,185 @@
+.webhook-create {
+
```
jessm12: Just a couple of comments about styling, where possible `rem` is probably preferable over `px`, when it's 0 this can be just on its own. Also, there are some carbon preset colours that could be useful if there are colours we like available

webhooks-extension/src/components/WebhookCreate/WebhookCreate.scss
```diff
@@ -0,0 +1,185 @@
...
+    align-items: center;
+    fill: red;
+    justify-content: center;
+  }
+  
+  .token {
+    align-items: center;
+    display: flex;
+  }
+
```
jessm12: Might be able to reduce some duplication of styles here, for example the only difference between these two is the visibility

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -0,0 +1,698 @@
...
+      newTokenValue: '',
+      createSecretDisabled: true,
+      // toggle access token 'password' or 'text'
+      pwType: 'password',
+      visibleCSS: 'token-visible',
+      invisibleCSS: 'token-invisible'
+    };
+  }
+
+  async fetchNamespaces() {
```
jessm12: Can these fetch/get methods be extracted into a different file to be used by the component? I think that could make the code easier to read and would reduce the amount of code in this one file, though I do see they are altering state so might not be an option currently

webhooks-extension/src/WebhookApp.js
```diff
@@ -0,0 +1,33 @@
...
+import { WebhookCreate } from './components/WebhookCreate';
+import { WebhookDisplayTable } from './components/WebhookDisplayTable';
+
+class WebhooksApp extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      // Used to show webhook created notification on table display
+      showNotification: false,
```
jessm12: `ShowNotification` is a general variable, can it be more specific so it doesn't require comment

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -0,0 +1,698 @@
...
+      apiServiceAccounts: '',
+      pipeline: '',
+      gitsecret: '',
+      serviceAccount: '',
+    });
+    if (!this.state.pipelineFail) {
+      this.fetchPipelines(itemText.selectedItem);
+    }
+    this.fetchSecrets(itemText.selectedItem);
+    this.fetchServiceAccounts(itemText.selectedItem);
```
jessm12: Should you be handling the rerender loop here as you are with the `fetchPipelines`?

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -0,0 +1,698 @@
...
+    this.setState({pipeline: itemText.selectedItem });
+  }
+
+  handleChangeSecret = (itemText, value) => {
+    this.setState({gitsecret: itemText.selectedItem });
+  }
+
+  handleChangeServiceAcct = (itemText, value) => {
+    this.setState({serviceAccount: itemText.selectedItem });
+  }
```
jessm12: You should be able to bring these three methods into one method by just substituting the state name, gitsecret etc, also value is not used here

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  handleSelectedRows = (rows) => {
+    console.log(rows);
+  }
+
+  formatCellContent(id, value) {
+    // Render the git repo as a clickable link
+    if (id.endsWith(":repository")) {
+      return <a href={value} target="_blank">{value}</a>
```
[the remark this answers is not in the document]
dibbles: I think though am not 100% certain that react's Link is for internal navigation only?? Not going to waste time investigating heavily at this point in time - would also need to work out how to open in new window.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
dibbles: The header and key are subtly different ... there is capitalisation in the Header and in one case spacing ... "Git Repository" ..... the 'key' value is matched in the setting up of initialRows .... so im currently inclined to keep it as it is.

webhooks-extension/src/components/WebhookCreate/WebhookCreate.scss
dibbles: some modifications made

webhooks-extension/src/components/WebhookCreate/WebhookCreate.scss
dibbles: @extend added to numerous elements

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
dibbles: I'd rather any refactor was done later ... we likely need to look at whether we could/should introduce any redux usage which might further alter things

webhooks-extension/src/WebhookApp.js
dibbles: changing to ShowNotificationOnTable

Issue #66, opened 2019-06-13
Webhooks tab indexing on creation form


dibbles:
The webhook creation UI seems to be a bit random when it comes to using the tab key to navigate around the page.... we should flow in a logical manner across the fields (this might have been looked at under the a11y issue https://github.com/tektoncd/experimental/issues/63)

carlos-logro:
@dibbles @AlanGreene 
so I managed to have the tabIndexes working in order & **partially** skip the disabled ones but for some reason the child elements of the Dropdown maintain the `tabIndex = 0`& right now Im stuck in figuring out how to bypass that. I even asked in the #carbon-react channel & did not have a solution. Ended up creating an issue in the carbon design system repo: https://github.com/carbon-design-system/carbon/issues/3232.


As for solution for this Im thinking of not displaying the dropdowns until the namespace is selected. What do you guys think???

carlos-logro:
came up with a better solution although it be a little cheating.

check the PR for more details.

PR #117, opened 2019-06-28
fix tab indexing & replace tooltips to hoverable ones

carlos-logro:
issue: #66 

#Changes

Fixed the tab indexing by initializing all the values to 0 & leaving the browser do the logical ordering and surprisinly works. But its worth mentioning that I had to cheat a little with plain javascript when trying to skip the disabled components because for some reason their child elements still had `tabIndex set to 0` and was interrupting the order. I also changed the tooltips to "hoverable" ones but only to those that are in the main page and not the modals because for some reason they were not being outputted correctly. Lastly, I had to update the test with some errors regarding the `window.toScroll` & create a new test file to bypass the `'initialFocus' did not return a node` error. 

# Submitter Checklist

review them:

### DIFF
--- a/webhooks-extension/config/extension-service.yaml
+++ b/webhooks-extension/config/extension-service.yaml
@@ -8,7 +8,7 @@ metadata:
   annotations:
     tekton-dashboard-display-name: Webhooks
     tekton-dashboard-endpoints: "webhooks.web"
-    tekton-dashboard-bundle-location: "web/extension.bae1277c.js"
+    tekton-dashboard-bundle-location: "web/extension.cb445729.js"
 spec:
   type: NodePort
   ports:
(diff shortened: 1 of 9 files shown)


dibbles:
@carlos-logro do you have a code formatter running? Most of the modifications are unrelated to the actually functional change and I wondered whether we should have the discussion about having a code formatter checked in, that everyone should apply to their dev environment - otherwise we will just keep seeing files changing back and forth as different people use different formatters.

Looks like you have unit test failures to fix up - but in general it looks fine.

carlos-logro:
@dibbles @AlanGreene  manage to remove the formatting. As for the failed test I don't know why does error messages are showing up because I didn't touch none of those files.

jessm12:
I've just given this another test and look over and all looks good, behaves as expected, just the hash conflicts to resolve and I'd say it's good to go 😄

carlos-logro:
@a-roberts @jessm12 ready for merge :)

a-roberts:
Yep, thanks Carlos (assuming tests continue to pass 😅)

/approve
/lgtm

Issue #61, opened 2019-06-13
Webhooks table should only show webhooks in the selected namespace


dibbles:
One could argue that the webhooks extension should only show items in the table that are in the selected namespace from the dropdown of the main dashboard (left hand menu).

PR #148, opened 2019-07-16
filter webhooks by selected namespace

carlos-logro:
issue: #61 

# Changes

This PR depends on a PR from the dashboard (https://github.com/tektoncd/dashboard/pull/354) where the selected namespace is passed as a prop to WebhookApp container which then passes the namespace to the WebhookTable. In the table, the namespace helps filter out the webhooks while also using to see if the namespace column needs to be shown in the table. Furthermore, if a webhook is not created under a certain namespace or all namespaces, a message will be shown to the user. Lastly, all test were updated & the test cases added were in the WebhookTable.test file.


carlos-logro:
/hold cancel /cc @dibbles @AlanGreene 

added package-lock file because the unit tests were not passing without it although i did add the package originally.

carlos-logro:
added package-lock file because the unit tests were not passing without it although i did add the package originally.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -171,37 +172,50 @@ export class WebhookDisplayTable extends Component {
           <Redirect to={this.props.match.url + "/create"} />
         )
       } else {
+        const { namespace } = this.props;
         // There are webhooks so display table
         const headers = [
           {
-            key: 'name',
-            header: 'Name',
+            key: "name",
```
AlanGreene: Why the change to double quotes here? Consistent quote style should be enforced by the linter.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -171,37 +172,50 @@ export class WebhookDisplayTable extends Component {
...
+        let initialRows = [];
         // Populate the data for the rows array from the data from the webhooks get request made on page load
-        this.state.webhooks.map(function (webhook, keyIndex) {
-          initialRows[keyIndex] = {
-            id: webhook['name']+"|"+webhook['namespace'],
-            name: webhook['name'],
-            repository: webhook['gitrepositoryurl'],
-            pipeline: webhook['pipeline'],
-            namespace: webhook['namespace'],
+        this.state.webhooks.map(function(webhook) {
```
AlanGreene: `map` should return an array, if you're just using it to iterate use `forEach` instead

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -171,37 +172,50 @@ export class WebhookDisplayTable extends Component {
...
+              id: webhook["name"] + "|" + webhook["namespace"],
+              name: webhook["name"],
+              repository: webhook["gitrepositoryurl"],
+              pipeline: webhook["pipeline"]
+            });
           }
-        })
+
+          if (namespace === ALL_NAMESPACES) {
+            initialRows[initialRows.length - 1].namespace =
```
AlanGreene: it would be clearer to set this before pushing the row onto the array

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
AlanGreene: could also deconstruct the `webhook` param here since you only ever need individual fields, avoiding all the `webhook[foo]`

```js
({ gitrepositoryurl, name, namespace, pipeline }) => {

}
```

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -280,6 +294,23 @@ export class WebhookDisplayTable extends Component {
               />
             </div>
 
+            {initialRows.length === 0 && namespace === ALL_NAMESPACES && (
```
AlanGreene: Refactor to avoid duplicating the containers, it's only the strings that differ in these 2 cases. This will help to ensure consistency after any future updates as there's only a single instance of the container that needs to be updated.

```js
{initialRows.length === 0 && (
  <p className=...>
    {namespace === ALL_NAMESPACES ? 'No webhooks created under any ...' : 'No webhooks created under namespace ...'}
  </p>
)}
```

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
```diff
@@ -24,22 +24,29 @@ describe('without webhooks', () => {
...
     jest.spyOn(API, 'getWebhooks').mockImplementation(() => Promise.resolve([noWebhooks]));
-    const { queryByText, queryByTestId } = renderWithRouter(<WebhookDisplayTable match={{}} />);
+    const { queryByText, queryByTestId } = renderWithRouter(<WebhookDisplayTable match={{}} namespace="*"/>);
     expect(queryByTestId('webhook-notification')).toBeNull();
     expect(queryByText(/Loading/i)).toBeTruthy();
   });
-}); 
+
+  it('should display string suggesting to create webhooks when no namespace has webhooks', () => {
+    const noWebhooks = [ ];
```
AlanGreene: there's no reference to this 'string suggesting to create webhooks' in the test, this looks identical to the test above

webhooks-extension/src/WebhookApp.js
```diff
@@ -25,15 +26,26 @@ class WebhooksApp extends Component {
...
 }
 
-export default withRouter(WebhooksApp)
+function mapStateToProps(state, props) {
+  return {
+    namespace: props.selectors.getSelectedNamespace(state)
+  };
+}
+
+const mapDispatchToProps = {};
```
AlanGreene: you can omit this from the call to `connect` since it's not being used

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
carlos-logro: oops must of not pushed that part

carlos-logro: removed after noticing that when there are no webhooks under any namespace the page redirects to the creating modal.

carlos-logro:
@AlanGreene updates done here.

Issue #150, opened 2019-07-17
Webhook tooltips trigger form to POST 


dibbles:
If you click on a tooltip on the webhook creation form you receive an error - as for some reason it attempts to submit the form.

PR #155, opened 2019-07-17
prevent tooltips from posting form

carlos-logro:
issue: #150 

# Changes

For some weird case the hoverable tooltips when clicked were POSTing the create webhook form. Tried numerous approaches but the one that worked the most was adding a onClickHandler that fires a the preventDefault function.

### DIFF
--- a/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
+++ b/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
@@ -12,6 +12,16 @@ import Infomation from "@carbon/icons-react/lib/information/16";
 
 import './WebhookCreate.scss';
 
+const CustomTooltip = props => (
+  <TooltipIcon {...props} >
+    <Infomation />
+  </TooltipIcon>
+);
+
+CustomTooltip.defaultProps = {
+  onClick: e => e.preventDefault()
+};
+
 class WebhookCreatePage extends Component {
 
   constructor(props) {
@@ -482,9 +492,7 @@ class WebhookCreatePage extends Component {
 
             <div className="row">
               <div className="help-icon" id="name-tooltip">
-                <TooltipIcon tooltipText="The display name for your webhook in this user interface. This name must consist of lower case alphanumeric characters, '-' or '.'">
-                  <Infomation />
-                </TooltipIcon>
+                <CustomTooltip tooltipText="The display name for your webhook in this user interface." />
               </div>
               <div className="item-label">
                 <div className="createLabel">Name</div>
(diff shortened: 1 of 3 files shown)


webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -484,7 +488,7 @@ class WebhookCreatePage extends Component {
 
             <div className="row">
               <div className="help-icon" id="name-tooltip">
-                <TooltipIcon tooltipText="The display name for your webhook in this user interface.">
+                <TooltipIcon onClick={this.preventTooltipPost} tooltipText="The display name for your webhook in this user interface.">
```
AlanGreene: `TooltipIcon` is intended to provide a tooltip explaining the purpose of an icon. If you want a standalone icon to trigger a tooltip for other purposes see [`Tooltip`](http://react.carbondesignsystem.com/?path=/story/tooltip--only-custom-icon)

However, in most cases forms should use a combination of labelText + helperText to present any relevant information to the user.

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -509,7 +513,7 @@ class WebhookCreatePage extends Component {
 
             <div className="row">
               <div className="help-icon" id="git-tooltip">
-                <TooltipIcon tooltipText="The URL of the git repository to create the webhook on.">
+                <TooltipIcon onClick={this.preventTooltipPost} tooltipText="The URL of the git repository to create the webhook on.">
```
AlanGreene: Instead of having to add this `onClick` to every Tooltip it would be better to define your own Tooltip component (wrapping the Carbon one) with a default onClick and use that. It will help to ensure consistency and make it easier to update all instances in future.

```js
const MyCustomTooltip = props => (
  <CarbonTooltip {...props} />
);

MyCustomTooltip.defaultProps = {
  onClick: e => e.preventDefault()
};

...

<MyCustomTooltip tooltipText="The URL of ..." />
```

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
carlos-logro: the reason behind using TooltipIcon is because its the one that offers hovering instead of clicking

carlos-logro:
@AlanGreene the reason behind using these tooltips is because:
1) These specific tooltips are the only ones that feature mouseover action & also show the text when the user uses de tab button to go through the form. 
2) the team agreed that adding labels & helperText will make the form too big & so tooltips would make the form look less cluttered.

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -499,9 +507,7 @@ class WebhookCreatePage extends Component {
 
             <div className="row">
               <div className="help-icon" id="git-tooltip">
-                <TooltipIcon tooltipText="The URL of the git repository to create the webhook on.">
-                  <Infomation />
-                </TooltipIcon>
+                <CustomTooltip onClick={this.preventTooltipPost} tooltipText="The URL of the git repository to create the webhook on."/>
```
AlanGreene: these `onClick`s still need to be removed, few more below

carlos-logro:
@AlanGreene this is ready for another review or merge. Thank you!

PR #166, opened 2019-07-24
Fix PR unittest status by returning the result of the tests not an echo

dibbles:
# Changes
 The unittest run was previously returning the result of the echo command rather than returning the result of the unittests.  I've added in a local variable to store this result and return it.

# Submitter Checklist

review them:

for more details._


dibbles:
@vdemeester Not a problem in your code ..... this was added by one of ours.  Hopefully my unittests will fail and we can start fixing the code/tests that are broken.

a-roberts:
/lgtm

😄

Issue #137, opened 2019-07-08
Webhooks front end code should use the actions and selectors provided by the dashboard


AlanGreene:
# Expected Behavior

The dashboard exposes a number of items to extensions to ensure consistency and compatibility, and to reduce code duplication. Among these are the redux actions and selectors required to request and process all of the core data used by the dashboard (namespaces, pipelines, etc.).

The extension code should use these, to ensure compatibility with any changes made in the dashboard and to benefit from the real-time updates provided.

# Actual Behavior
The webhooks extension implements its own API layer and makes the REST calls directly.

Recent example of a change required in the extension that could have been avoided: https://github.com/tektoncd/experimental/pull/133

# Reference
All of the dashboard's top-level selectors and a subset of the actions are provided as props to the extension:
https://github.com/tektoncd/dashboard/blob/ee0bbc2ee9366350d75f20230144bcb20adcd822/src/containers/Extension/Extension.js#L76

To use the selectors you must provide the current store state. One way to do this is to wrap your component in react-redux's [`connect`](https://react-redux.js.org/api/connect) function and use `mapStateToProps` as we do in the core dashboard.

react-redux is already made available to extensions in the same way as react, react-router-dom, Carbon, etc. https://github.com/tektoncd/dashboard/blob/99a34a490ea70dce792101f0196246480473bc39/src/containers/Extension/globals.js#L23

PR #169, opened 2019-07-24
replace api calls with dashboard's redux selectors & actions

carlos-logro:
issue: #137 

# Changes

### DIFF
--- a/webhooks-extension/config/extension-service.yaml
+++ b/webhooks-extension/config/extension-service.yaml
@@ -8,7 +8,7 @@ metadata:
   annotations:
     tekton-dashboard-display-name: Webhooks
     tekton-dashboard-endpoints: "webhooks.web"
-    tekton-dashboard-bundle-location: "web/extension.fc311844.js"
+    tekton-dashboard-bundle-location: "web/extension.c83f46fd.js"
 spec:
   type: NodePort
   ports:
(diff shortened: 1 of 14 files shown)


AlanGreene:
Please break this PR down into more easily reviewed chunks. Some suggestions:
- migrate a subset (even 1) of the APIs instead of doing them all in one go
- keep large whitespace / formatting changes separate from functional changes where possible
- avoid introducing unnecessary complexity in mock data (e.g. most of the resource metadata fields are irrelevant), it distracts from the actual tests / functionality of interest.

I'm also seeing a lot of inconsistency in use of single vs. double quotes, and other issues that should be picked up / fixed by the linter.

carlos-logro:
@AlanGreene 
Divided the PR so in this first one I included the update to the namespace API call only and each modified test. Also, at the moment the prettier & ESLint rules are not on in this repo, that's why they are not being picked up.

AlanGreene:
@carlos-logro conflict on the hash again. It might be worth investing (separately) in removing the need to specify that at all since virtually any front end change causes conflicts.

Otherwise looks good 👍

carlos-logro:
@AlanGreene 
re-hashed. 

& definitely, a workaround must be done with the hash to make it simpler merge PRs.

Issue #170, opened 2019-07-24
resolve warning message: Cannot update during an existing state transition (such as within `render`).


carlos-logro:
# Expected Behavior
state is being changed without warning messages. 

# Actual Behavior
state is being set in the render method which produces this warning message: 
`Warning: Cannot update during an existing state transition (such as within 'render'). Render methods should be a pure function of props and state.`

# Steps to Reproduce the Problem

1. Go to the browser & use extension normally.
2. Message will appear in the console.

PR #171, opened 2019-07-24
remove warning message when setting state in render method

carlos-logro:
issue: #170 

# Changes
 
the warning message was removed when the lifecycle method getDerivedStateFromProps was added with the lines that were changing state in the render method.

### DIFF
--- a/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
+++ b/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
@@ -136,6 +136,24 @@ class WebhookCreatePage extends Component {
     }
   }
 
+  componentDidUpdate() {
+    if (this.props.namespaces) {
+      if (this.state.createSecretDisabled) {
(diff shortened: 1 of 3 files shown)


webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -136,6 +136,25 @@ class WebhookCreatePage extends Component {
     }
   }
 
+  static getDerivedStateFromProps(nextProps, prevState) {
```
AlanGreene: Can we use `componentDidUpdate` instead? Generally `getDerivedStateFromProps` isn't required or recommended. More detail: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state

carlos-logro: ok ill see if it does the job!

carlos-logro:
@AlanGreene managed to replace it with the `componentDidUpdate`.

PR #176, opened 2019-07-29
replace pipelines api call with dashboard's redux selectors & actions

carlos-logro:
issue: #137 

# Changes


carlos-logro:
@AlanGreene this PR only includes everything regarding the pipeline api call. The third one & last one will be for the service account one.

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -86,10 +85,8 @@ class WebhookCreatePage extends Component {
   }
 
   async fetchPipelines(namespace) {
-    let pl;
     try {
-      pl = await getPipelines(namespace);
-      this.setState({apiPipelines: pl})
+      this.props.fetchPipelines(namespace);
     } catch (error) {
```
AlanGreene: the error should already be caught by the action and is exposed via the `getPipelinesErrorMessage` selector, no need for the try-catch here

carlos-logro: done. Added logic to output notification if error happens and a test case.

PR #182, opened 2019-07-31
replace service accounts api call with dashboard's redux selectors & actions

carlos-logro:

### DIFF
--- a/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
+++ b/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { withRouter } from 'react-router-dom';
 
 import { Button, TextInput, Dropdown, Form, Tooltip, DropdownSkeleton, Modal, InlineNotification, TooltipIcon } from 'carbon-components-react';
-import { getSecrets, getServiceAccounts, createWebhook, createSecret, deleteSecret } from '../../api/index';
+import { getSecrets, createWebhook, createSecret, deleteSecret } from '../../api/index';
 
 import AddAlt20 from '@carbon/icons-react/lib/add--alt/20';
 import SubtractAlt20 from '@carbon/icons-react/lib/subtract--alt/20';
@@ -78,7 +78,6 @@ class WebhookCreatePage extends Component {
       dockerRegistry: '',
       // fetched data from api calls
       apiSecrets: '',
-      apiServiceAccounts: '',
       // whether or not to show secret modals
       showDeleteDialog: false,
       showCreateDialog: false,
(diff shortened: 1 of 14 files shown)


webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -84,6 +83,14 @@ class WebhookCreatePage extends Component {
         showNotification: true
       });
     }
+    if(this.props.getServiceAccountsErrorMessage){
+      this.setState({
+        notificationMessage: this.props.getServiceAccountsErrorMessage,
```
AlanGreene: should this be 
```diff
-notificationMessage: this.props.getServiceAccountsErrorMessage,
+notificationMessage: this.props.getServiceAccountsErrorMessage(),
```

?

webhooks-extension/src/components/WebhookCreate/tests/WebhookCreate.test.js
```diff
@@ -190,6 +199,25 @@ describe('drop downs should be enabled when a namespace is selected', () => {
...
+        fetchPipelines={() => {}}
+        fetchServiceAccounts={() => {}}
+        isFetchingPipelines={false}
+        isFetchingServiceAccounts={false}
+        serviceAccounts={serviceAccounts}
+        getServiceAccountsErrorMessage={"Error fetching service accounts."}
+        setShowNotificationOnTable={() => {}}
+      />
+    );
+    await wait(() => expect(document.getElementById("pipeline").getElementsByClassName("bx--list-box__field").item(0).hasAttribute("disabled")).toBe(true))
```
AlanGreene: can we avoid these native DOM functions and instead use the API provided by react-testing-library based on the content visible to the user?

webhooks-extension/src/components/WebhookCreate/tests/WebhookCreate.test.js
```diff
@@ -190,6 +199,25 @@ describe('drop downs should be enabled when a namespace is selected', () => {
...
+      <WebhookCreate
+        match={{}}
+        namespaces={namespaces}
+        pipelines={pipelines}
+        fetchPipelines={() => {}}
+        fetchServiceAccounts={() => {}}
+        isFetchingPipelines={false}
+        isFetchingServiceAccounts={false}
+        serviceAccounts={serviceAccounts}
+        getServiceAccountsErrorMessage={"Error fetching service accounts."}
```
AlanGreene: no need for the braces here

`getServiceAccountsErrorMessage="Error fetching service accounts."`


naming here is also a bit odd, the `get` suggests this should be a function. See my other comment about where this is actually used.
Maybe just `serviceAccountsErrorMessage` would be better?

carlos-logro:
@AlanGreene this is the last PR in regards with the api calls done in the experimental repo. Its worth mentioning that it will only work if this PR in the dashboard repo is merged first: https://github.com/tektoncd/dashboard/pull/393.

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
carlos-logro: i see the naming is confusing. ill fix this.

webhooks-extension/src/components/WebhookCreate/tests/WebhookCreate.test.js
carlos-logro: i can go ahead & do it but most of the tests were already using DOM fns. It might make the PR really big

AlanGreene: ok let's just update these new tests, the others can be addressed later

AlanGreene: ok let's just update these new tests, the others can be addressed later

carlos-logro:
@AlanGreene for some reason the data-testid attribute was not showing up in the namespace & service account dropdown. had to use the roles to access the elements. Anyways I edited those two cases for now. I suggest creating an issue to tackle this test issue as well as another for the code format

carlos-logro:
@AlanGreene with the dependent PR from dashboard merged, this one is ready for merge as well, if everything meets expectations of course.

Issue #179, opened 2019-07-30
Notification not showing success when deleting all webhooks


dibbles:
# Expected Behavior

When deleting the last webhook a notification should be shown on the create webhook page to indicate the last webhook has been deleted and thus none now exist.

# Actual Behavior

Whilst the notification is shown if you only have one webhook (which you then delete), if you create two and then select both and press delete - the notification does not show.

# Steps to Reproduce the Problem

1. Create 2 webhooks
2. Select both in table
3. Click delete (no notification is shown on resultant create webhook page)

# Additional Info

PR #184, opened 2019-08-01
fix bug when deleting all webhooks in table

carlos-logro:
issue: #179 

# Changes
Previously I had it check if the last one was being deleted, now I changed the if statement to check if all the selected webhooks are all the webhooks in the table, then show the notification.

### DIFF
--- a/webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
+++ b/webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
@@ -132,7 +132,7 @@ export class WebhookDisplayTable extends Component {
 
     Promise.all(deletePromises).then( () => {
       this.fetchWebhooksForTable();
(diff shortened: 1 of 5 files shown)


webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -132,7 +132,7 @@ export class WebhookDisplayTable extends Component {
 
     Promise.all(deletePromises).then( () => {
       this.fetchWebhooksForTable();
-      if(this.state.webhooks.length - 1 === 0){
+      if(this.state.webhooks.length - rowsToUse.length === 0){
```
AlanGreene: is there a test to cover this?

carlos-logro:
@AlanGreene at the moment the test covering that line was given the case of only one webhook in the table. I added some test cases with 2 & 3 webhooks.

Issue #134, opened 2019-07-04
Move search bar border to outer edge


dibbles:
# Expected Behavior

When searching on the webhooks table, I would expect the blue border around the search entry to be on the outer edge

# Actual Behavior

Border currently appears on the inner text area, likely due to a change in the underlying carbon component

# Steps to Reproduce the Problem

1. Go to webhooks table
2. Click on the search icon

# Additional Info

<img width="942" alt="Screenshot 2019-07-04 at 13 41 19" src="https://user-images.githubusercontent.com/22838162/60667429-6e3cb780-9e61-11e9-80b2-56c952f9e46c.png">

PR #195, opened 2019-08-06
Move search bar outline to outer edge

carlos-logro:
issue: #134 

# Changes
Achieved by removing the padding in the search bar.


AlanGreene:
Was this not fixed in Carbon?

carlos-logro:
> Was this not fixed in Carbon? [carbon-design-system/carbon#3304](https://github.com/carbon-design-system/carbon/issues/3304)

Same exact thing I did.... how can i access that update?? i assume by updating the whole carbon design module?

AlanGreene:
Yes you'll need to update `carbon-components` in tektoncd/dashboard and should also update in this repo for consistency in tests or any standalone dev efforts

carlos-logro:
@AlanGreene managed to update the dependencies & the search bar outline got fixed but ran into an issue that I believe is with the Carbon Components. Basically after the updates I checked to see if all the tests were passing but apparently in the `WebhookCreate.test` file there are failing cases. These cases test the namespace & SA dropdown to verify if they are disabled or enabled and I believe the producer behind the failure is because for some reason the `disabled` attribute is not appearing. 

before update: 
*disabled attr appeared & all test successful* 
Screen Shot 2019-08-08 at 3 19 18 PM

after update:
*disabled attr missing & those specific test cases failing*
Screen Shot 2019-08-08 at 3 03 12 PM
Screen Shot 2019-08-08 at 3 03 31 PM

carlos-logro:
@AlanGreene do I have to do a PR with an update to the carbon components in the dashboard? Only had to change the test cases related to the disabled dropdowns. Apparently with the update they remove the `disabled` attr & instead added it as a class name.

AlanGreene:
yes you'll need to update the dependency in the dashboard as that's the one that's actually used in the build

carlos-logro:
@AlanGreene created the PR with the necessary update & referenced it above.

Issue #192, opened 2019-08-06
On refreshing create webhooks page, 404 error returns getting secrets


a-roberts:
# Expected Behavior

It shouldn't

# Actual Behavior

http://localhost:9097/#/extensions/webhooks-extension/create

# Steps to Reproduce the Problem

1. Go to the webhook create page
2. Refresh the page
3. Observe

![image](https://user-images.githubusercontent.com/12814972/62528173-4f508d00-b834-11e9-9c4b-9c8529434b2c.png)

It tries to use the url http://localhost:9097/v1/extensions/webhooks-extension/create/webhooks/credentials instead of http://localhost:9097/v1/extensions/webhooks-extension/webhooks/credentials 

# Additional Info
Code to fix is likely here
https://github.com/tektoncd/experimental/blob/master/webhooks-extension/src/api/index.js#L27

replace anything after, not just the extension part

PR #196, opened 2019-08-06
fix 404 error when refreshing create webhook page

carlos-logro:
issue: #192 

# Changes
The approach I did was basically to search & remove the keyword `create` that was is not supposed to be present in the api uri call.

### DIFF
--- a/webhooks-extension/src/api/index.js
+++ b/webhooks-extension/src/api/index.js
@@ -24,8 +24,8 @@ export function getSelectedRows(selectedRows) {
 
 export function getAPIRoot() {
   const { href, hash } = window.location;
-  let newHash = hash.replace('#/extensions','v1/extensions')
-  let baseURL = href.replace(hash, newHash);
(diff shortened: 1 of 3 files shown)


webhooks-extension/src/api/index.js
```diff
@@ -24,7 +24,10 @@ export function getSelectedRows(selectedRows) {
 
 export function getAPIRoot() {
   const { href, hash } = window.location;
-  let newHash = hash.replace('#/extensions','v1/extensions')
+  let newHash = hash.replace('#/extensions','v1/extensions');
+  if(hash.search('create') !== -1){
```
AlanGreene: with this approach you'll have to update this for other potential paths in future
why not just replace the whole hash with what you're expecting?

carlos-logro: soo basically hard code the newHash variable?

carlos-logro:
@AlanGreene this PR was updated, i hardcoded the go to hash

PR #215, opened 2019-09-03
add yaml highlighting to codeblock

mcristina422:
# Changes

This adds a yaml block to the codeblock example so that Github will color the text


a-roberts:
tekton-listener's graduated to become Tekton Triggers (which is official and pucker and ready to be tested/contributed to), I'm going to close this one out, thanks anyway @mcristina422 😄

Issue #201, opened 2019-08-14
Error reported when deleting webhooks with multiple PipelineRuns


jessm12:
# Expected Behavior
When using the experimental webhooks-extension for the Dashboard, deleting a webhook and its associated `PipelineRuns` completes successfully and reports as a success

# Actual Behavior
<img width="1158" alt="Screenshot 2019-08-14 at 11 01 51" src="https://user-images.githubusercontent.com/32932155/63015319-83f1c380-be88-11e9-8fd4-6fca9ff5011c.png">
@dibbles first observed this problem when deleting two webhooks at once which both had a large number of runs I was able to see the same in my environment with a similar scenario.

The deletion of both the webhooks themselves and the `PipelineRuns` appeared to be successful though did seem to take a while to complete. 

# Steps to Reproduce the Problem

1. Create multiple webhooks and trigger them multiple times
2. Try to delete the webhooks and their associated `PipelineRuns`
3. Observe the error as shown above 

# Additional Info
We aren't sure of the cause of this issue but suspect it could be some form of service timeout 🤔

a-roberts:
Picked this up, got two Git repos with webhooks triggering the same pipeline (using the 0.1.1 release) on OpenShift 3.11.

Clicked to delete both webhooks, confirmed I see the error being reported and yet all is well

```
An error occurred deleting webhook(s).

Check the webhook(s) existed and both the dashboard and extension pods are healthy.
```

No errors in the backend, when I implemented this I added a bunch of delete promises with a 500ms timeout (potentially for *everything* to be done and not just for each promise, but I'll check). Definitely to do with the frontend handling of it that I added, so on the case, and it's the same deletion code and table on master so I'm sure it'll be seen there.

a-roberts:
PR https://github.com/tektoncd/experimental/pull/218

a-roberts:
Merged so closing

PR #218, opened 2019-09-05
Increase delete webhook timeout to 5s, display warning if timeout, refresh page

a-roberts:
# Changes

For https://github.com/tektoncd/experimental/issues/201

Tested with two webhooks with 10 PipelineRuns each on a pretty slow OpenShift on Fyre cluster

I realise this isn't exactly massive scale so perhaps we need to come up with a better solution eventually (@dibbles has noticed retrieving pipelines being very slow too for a given namespace, so I think we need to improve our "slow system" handling as part of a larger item)

Either way, this gets us past the current problem where 500 ms is definitely too short a timeout period, and because of the refresh and warning, users will see that even if a timeout does occur (for testing I set it to 1 ms so I'd always see the notification), a notification occurs and the table is refreshed anyway incase it really was deleted.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
+++ b/webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
@@ -32,7 +32,11 @@ const {
 } = DataTable;
 
 const ALL_NAMESPACES = "*";
+var displayTimeoutWarningNotification = false;
+
 export class WebhookDisplayTable extends Component {
+
+
   state = {
     showTable: true,
     showDeleteDialog: false,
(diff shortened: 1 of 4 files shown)


webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -146,14 +149,23 @@ export class WebhookDisplayTable extends Component {
         });
       }
      }).catch( () => {
+      this.fetchWebhooksForTable();
+      let notificationStatusToSet = '';
+      let notificationMessagetoSet = '';
+      if (displayTimeoutWarningNotification) {
+        notificationStatusToSet = 'warning'
+        notificationMessagetoSet = 'Warning - timed out waiting to delete webhooks and potentially PipelineRuns: manually check that the resources you expect have been deleted successfully.';
```
jessm12: This message is quite verbose, I would slim it down a bit 😅

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -146,14 +149,23 @@ export class WebhookDisplayTable extends Component {
         });
       }
      }).catch( () => {
+      this.fetchWebhooksForTable();
+      let notificationStatusToSet = '';
+      let notificationMessagetoSet = '';
```
jessm12: casing

Issue #226, opened 2019-09-10
Task monitor fails to update the pull request


dibbles:
# Expected Behavior

Using github.com, the pull request is supposed to be updated by the task monitor.

# Actual Behavior

The tekton code fails to parse the provided URL and pull out the pull request number.  Additionally, from looking at the sink log.

# Steps to Reproduce the Problem

1. Webhook in github
2. Create pull request
3.

# Additional Info

dibbles:
https://github.com/tektoncd/experimental/pull/228

PR #228, opened 2019-09-10
Change data used from webhook payload in pipelineresources

dibbles:
Issue #226 
Issue #226 

Changes

The pull request pipeline resource appears to want the HTMLURL from the
GitHub webhooks payload, otherwise the string manipulation fails to find
the pull request number.  Changed the sink to use 
`webhookData.PullRequest.HTMLURL` rather than the current
`strings.Replace(webhookData.PullRequest.URL, "api/v3/repos/", "", 1)`

Additionally, the way the branch name is obtained in the sink needs to
differ between push and pull_request events as the payloads are different.
In the case of the pull_request event we no longer send the ref through
`extractBranchFromRef` and the method itself has been renamed to 
`extractBranchFromPushEventRef`

# Submitter Checklist

review them:

- [n/a ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._


a-roberts:
Yeah, we know it's definitely not good for GitHub.com and this works a treat still on GHE, so let's get it in 😹 

/lgtm

PR #230, opened 2019-09-10
Add experimental Octant plugin

ImJasonH:
# Changes

This adds a simple plugin for [Octant](https://github.com/vmware/octant), a new extensible Kubernetes dashboard.

Out of the box, Octant displays high-level information about built-in Kubernetes resources like Pods, and minimal information it can derive about custom resource types like ours.

This plugin adds deeper information about Tekton resources, such as a link to a TaskRun's underlying Pod, which also surfaces that Pod's logs.

Octant provides built-in support for Graphviz visualizations, which could be useful for visualizing the dependency graph of a Pipeline, and the execution status of a PipelineRun. (This is currently blocked on https://github.com/vmware/octant/issues/248)

# Submitter Checklist

review them:

- [:green_heart:] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [:green_heart:] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)


AlanGreene:
I'm seeing timeouts loading deps from git.apache.org so had to add this to my go.mod:
`replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999`

Also had to make sure I had `GO111MODULE=on` in my env, otherwise this was super simple to install and run. Looks great +💯

octant-plugin/main.go
```diff
@@ -0,0 +1,446 @@
...
+			}
+		}),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+	p.Serve()
+}
+
+func toUnstructured(i interface{}) (*unstructured.Unstructured, error) {
```
sbwsg: it doesn't look like this is used anywhere (from a quick ctrl+f `toUnstructured`). Still needed?

ImJasonH: Right you are! It was previously used to create new resources, but that's not yet supported. I've removed it.

I've also split out all the type-specific print methods so they're not part of a super-huge `switch` block.

Issue #50, opened 2019-06-05
Display build status in webhook 'branches' collection


mnuttall:

mii-w:
<img width="615" alt="Screen Shot 2019-09-20 at 11 33 04 AM" src="https://user-images.githubusercontent.com/51759906/65339797-d1093980-db9a-11e9-8227-67aff4622271.png">

carlos-logro:
first pass has been merged, unassigning but not closing due to new designs
/unassign

AlanGreene:
The webhooks extension is deprecated, closing stale issues.

/close

Issue #48, opened 2019-06-05
View list of branches for a webhook


mnuttall:
Further to #47 and as per the .pdf sketch linked in #37, clicking on a webhook name in the list of webhooks should take the user to a list of branches in the associated git repository. This will need additional support from the REST API to retrieve this list - we might as well get the list of branches from git, rather than by scraping them from all the pipelineruns we might have. 

(There's the possibility of some branches existing but having no builds, and of builds existing whose branches have been deleted. That could be a problem but we're deferring it for now.)

mii-w:
First take from referencing Alan's and Duane's screens
<img width="615" alt="Screen Shot 2019-09-20 at 11 33 04 AM" src="https://user-images.githubusercontent.com/51759906/65339656-8c7d9e00-db9a-11e9-9b75-e4b553d9b702.png">

mii-w:
![Screen Shot 2019-10-02 at 12 52 01 PM](https://user-images.githubusercontent.com/51759906/66064495-7b764a80-e513-11e9-94a3-de475f55e33f.png)
![Screen Shot 2019-10-02 at 12 52 09 PM](https://user-images.githubusercontent.com/51759906/66064502-7e713b00-e513-11e9-9ef6-d6078c62861f.png)

mii-w:
Clickable wireframe Prototype:
https://ibm.invisionapp.com/share/PCNZBZYNKQF

carlos-logro:
first pass has been merged, unassigning but not closing due to new designs 
/unassign

Issue #37, opened 2019-06-04
Webhooks for multiple branches in a single repository


mnuttall:
See slide 9 in this presentation by @dibbles. : 
[Extension.pdf](https://github.com/tektoncd/experimental/files/3251593/Extension.pdf)

Currently we build commits on any branch but end up with a jumble of pipeline runs sorted at best by date. We need to be able to filter pipeline runs by repository and then further, optionally, by branch. @MNUTTALL talked to @skaegi  about this at the end of May. We agreed that applying labels to PipelineRuns and then filtering by those labels is the correct design.

- #47 Add 'Repo' and 'Branch' filters to PipelineRuns
- #48 Clicking webhook name in webhooks collection takes user to a list of branches (all branches) in the repo
- #49 Clicking branch name takes user to a view of filtered pipeline runs 
- #50 Display latest status in 'branches' collection - build status and time columns
- #51 Display the settings collected when creating hook, on hook details tab

mnuttall:
[Extension.pdf](https://github.com/tektoncd/experimental/files/3251585/Extension.pdf)

PR #237, opened 2019-09-13
View Branches when Webhook Clicked

carlos-logro:
as per epic #37, issues #48 & #50  are implemented in this PR

# Changes
In the `WebhookDisplayTable` container I added a "passive" modal component that will display the branches within a simple sortable table. It uses the GitHub API to fetch the branches & saves the data in the component's state rather than the redux store. Lastly, I added the necessary test cases as well.

# Submitter Checklist

review them:

### DIFF
--- a/webhooks-extension/src/WebhookApp.js
+++ b/webhooks-extension/src/WebhookApp.js
@@ -30,6 +30,7 @@ class WebhooksApp extends Component {
   render() {
     const {
       fetchPipelines,
+      fetchPipelineRuns,
       fetchServiceAccounts,
       pipelinesErrorMessage,
       serviceAccountsErrorMessage,
@@ -56,6 +57,7 @@ class WebhooksApp extends Component {
               setshowLastWebhookDeletedNotification={
                 this.setshowLastWebhookDeletedNotification
               }
+              fetchPipelineRuns={fetchPipelineRuns}
             />
           )}
         />
@@ -107,7 +109,9 @@ const mapDispatchToProps = (dispatch, props) => ({
   fetchPipelines: namespace =>
     dispatch(props.actions.fetchPipelines({ namespace })),
   fetchServiceAccounts: namespace =>
-    dispatch(props.actions.fetchServiceAccounts({ namespace }))
+    dispatch(props.actions.fetchServiceAccounts({ namespace })),
+  fetchPipelineRuns: (namespace, filters) =>
+    dispatch(props.actions.fetchPipelineRuns({namespace, filters}))
 });
 
 export default connect(
--- a/webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
+++ b/webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
@@ -1,0 +1,236 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import React, { Component } from 'react';
+import { Modal } from 'carbon-components-react';
+
+import './WebhookBranches.scss';
+
+import {
+  DataTable,
+  DataTableSkeleton,
+  InlineNotification
+} from 'carbon-components-react';
+
+const {
+  TableContainer,
+  Table,
+  TableHead,
+  TableRow,
+  TableBody,
+  TableCell,
+  TableHeader
+} = DataTable;
+
+export class WebhookBranches extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      rows: [],
+      loading: true,
+      error: null
+    };
+  }
+
+  componentDidMount() {
+    let { url, namespace, pipeline } = this.props.webhook;
+    let [server, org, repo] = url
+      .toLowerCase()
+      .replace(/https?:\/\//, "")
+      .split("/");
+
+    this.props
+      .fetchPipelineRuns(
+        namespace,
+        [
+          `gitOrg=${org}`,
+          `gitServer=${server}`,
+          `gitRepo=${repo}`,
+          `tekton.dev/pipeline=${pipeline}`
+        ]
+      )
+      .then(pipelineRuns => {
+        let branches = [];
+        const rows = pipelineRuns
+          .sort(
+            (a, b) =>
+              new Date(
+                b.status.conditions[
+                  b.status.conditions.length - 1
+                ].lastTransitionTime
+              ) -
+              new Date(
+                a.status.conditions[
+                  a.status.conditions.length - 1
+                ].lastTransitionTime
+              )
+          )
+          .reduce((result, pipelineRun) => {
+            if (
+              branches.indexOf(pipelineRun.metadata.labels.gitBranch) === -1
+            ) {
+              branches.push(pipelineRun.metadata.labels.gitBranch);
+              const time = new Date(
+                pipelineRun.status.conditions[
+                  pipelineRun.status.conditions.length - 1
+                ].lastTransitionTime
+              );
+              result.push({
+                id: `${pipelineRun.metadata.labels.gitBranch}-branch`,
+                branch: pipelineRun.metadata.labels.gitBranch,
+                time: `${time.toLocaleDateString()} - ${time.toLocaleTimeString()}`,
+                status:
+                  pipelineRun.status.conditions[
+                    pipelineRun.status.conditions.length - 1
+                  ].reason
+              });
+            }
+            return result;
+          }, []);
+
+        this.setState({
+          rows,
+          loading: false
+        });
+      })
+      .catch(error => {
+        error.response.text().then(text => {
+          this.setState({
+            error: text,
+            rows: [],
+            loading: false
+          });
+        });
+      });
+  }
+
+  render() {
+    const { close } = this.props;
+    const { rows, loading, error } = this.state;
+
+    const headers = [
+      {
+        key: 'branch',
+        header: 'Branch'
+      },
+      {
+        key: 'time',
+        header: 'Last Build Time'
+      },
+      {
+        key: 'status',
+        header: 'Status'
+      }
+    ];
+
+    return (
+      <Modal
+        open
+        id="webhook-branches-modal"
+        modalHeading="Latest PipelineRuns By Branch:"
+        passiveModal
+        onRequestClose={close}
+      >
+        {error && (
+          <InlineNotification
+            kind="error"
+            subtitle={error}
+            title="Error:"
+            lowContrast
+          />
+        )}
+        <div className="WebhookDetails">
+          <p>
+            <span>Webhook Name: </span>
+            {this.props.webhook.name}
+          </p>
+          <p>
+            <span>Repository: </span>
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href={this.props.webhook.url}
+            >
+              {this.props.webhook.url}
+            </a>
+          </p>
+          <p>
+            <span>Pipeline: </span>
+            {this.props.webhook.pipeline}
+          </p>
+          <p>
+            <span>Namespace: </span>
+            {this.props.webhook.namespace}
+          </p>
+        </div>
+        <DataTable
+          useZebraStyles
+          rows={rows}
+          headers={headers}
+          render={({ rows, headers, getHeaderProps, getRowProps }) => (
+            <TableContainer>
+              {loading ? (
+                <DataTableSkeleton
+                  rowCount={1}
+                  columnCount={headers.length}
+                  data-testid="loading-table"
+                />
+              ) : (
+                <Table>
+                  <TableHead>
+                    <TableRow>
+                      {headers.map(header => (
+                        <TableHeader
+                          key={header.id}
+                          {...getHeaderProps({ header })}
+                          isSortable
+                          isSortHeader
+                        >
+                          {header.header}
+                        </TableHeader>
+                      ))}
+                    </TableRow>
+                  </TableHead>
+                  <TableBody>
+                    {rows.map(row => (
+                      <TableRow {...getRowProps({ row })} key={row.id}>
+                        {row.cells.map((cell, index) => (
+                          <TableCell
+                            className="cellText"
+                            key={cell.id}
+                            data-status={
+                              index === row.cells.length - 1 ? cell.value : null
+                            }
+                          >
+                            {cell.value}
+                          </TableCell>
+                        ))}
+                      </TableRow>
+                    ))}
+                  </TableBody>
+                </Table>
+              )}
+            </TableContainer>
+          )}
+        />
+        {rows.length === 0 && !loading && (
+          <div className="noBranches">
+            <p>
+              Unable to identify any PipelineRuns initiated by this webhook.
+            </p>
+          </div>
+        )}
+      </Modal>
+    );
+  }
+}
--- a/webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
+++ b/webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
@@ -1,0 +1,53 @@
+/*
+Copyright 2019 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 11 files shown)


webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
+import React, { Component } from "react";
```
AlanGreene: single quotes for consistency with the rest of the code
can you also open an issue to track enabling prettier on this project to avoid this type of code style issue, or we may already have one

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
...
+    this.state = {
+      rows: [],
+      loading: false,
+      error: null
+    };
+  }
+
+  componentDidMount() {
+    this.setState({
+      loading: true
```
AlanGreene: the pattern we use elsewhere is to default to `loading: true` in the initial state

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
...
+            status: "No Results Available"
+          })),
+          loading: false
+        });
+      })
+      .catch(error => {
+        error.response.text().then(text => {
+          this.setState({
+            error: text,
+            rows: [
```
AlanGreene: Do we need to include this dummy row in the error state? Can we just leave rows empty and display the error?

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
...
+    return (
+      <Modal
+        open={true}
+        id="webhook-branches-modal"
+        data-testid="webhook-branches-modal"
+        modalHeading="Branches"
+        passiveModal
+        onRequestClose={close}
+      >
+        {error !== null && (
```
AlanGreene: can probably just use `{error && (...` unless error could have another falsey value such as undefined, 0, false, etc.

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
...
+          <InlineNotification
+            data-testid="branches-modal-notification"
+            kind="error"
+            subtitle={error}
+            title="Error: "
```
AlanGreene: trailing space

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
AlanGreene: missing copyright header
same on some other new files

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,151 @@
...
+          render={({ rows, headers, getHeaderProps, getRowProps }) => (
+            <TableContainer>
+              {loading ? (
+                <DataTableSkeleton
+                  rowCount={1}
+                  columnCount={headers.length}
+                  data-testid="loading-table"
+                />
+              ) : (
+                <Table className="bx--data-table--zebra">
```
AlanGreene: should this be using the DataTable prop to enable zebra styling instead of adding the class here? If Carbon changes the class name this would break, whereas the prop is part of the public API.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -348,6 +362,7 @@ export class WebhookDisplayTable extends Component {
                 </fieldset>
               </Modal>
             </div>
+            {this.state.selectedWebhook !== null && <WebhookBranches webhook={this.state.selectedWebhook} close={this.closeBranches}/>}
```
AlanGreene: no need for the `!== null` unless it can have other falsey values
`{this.state.selectedWebhook && ...`

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
```diff
@@ -85,4 +85,16 @@ describe('with webhooks', () => {
...
   });
+
+  it('display branch modal when table row clicked', async () => {
+    jest.spyOn(API, "getWebhooks").mockImplementation(() => Promise.resolve(webhooks));
+    jest.spyOn(API, "fetchBranches").mockImplementation(() => Promise.resolve([]));
+    const { getByTestId } = renderWithRouter(<WebhookDisplayTable match={{}} selectedNamespace="*"/>);
+
+    await waitForElement(() => getByTestId(/webhook-table-row-first-test-webhook/i));
+
+    fireEvent.click(getByTestId(/webhook-table-row-first-test-webhook/i));
```
AlanGreene: we should only use data-testid if there's no user-facing content we can reference in the tests, e.g. button text, headers, etc.

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
carlos-logro: weird since this is a new file to the repo I applied prettier before pushing... thought all the quotes would be fixed automatically like it does in the dashboard repo. Is it possible that its not configured properly in this repo?

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
carlos-logro: ill see if it breaks without it, if not then ill remove it

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
carlos-logro: will try it!

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
carlos-logro: numerous files including old ones not included in this PR are also missing the copyright header for example `WebhookApp`. To keep this PR small I propose adding the header in another PR

AlanGreene: I don't see a .prettierrc in the webhooks-extension root, you could add one with this content (same as dashboard):

```
{
  "singleQuote": true,
  "trailingComma": "none"
}
```

AlanGreene: I don't include copyright headers in the diff size ;)

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
```diff
@@ -0,0 +1,33 @@
...
+    width: 100%;
+    text-align: center;
+    margin-top: 5px;
+    font-style: italic;
+    font-size: 1rem;
+  }
+
+  .bx--data-table tr th,
+  .bx--data-table tr th button {
+    background: #7190c1;
```
AlanGreene: is this consistent with other blues already in use?

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
```diff
@@ -85,4 +85,16 @@ describe('with webhooks', () => {
     const { getByText } = renderWithRouter(<WebhookDisplayTable match={{}} selectedNamespace="default"/>);
     await waitForElement(() => getByText("No webhooks created under namespace 'default', click 'Add Webhook' button to add a new one."));
   });
+
+  it('display branch modal when table row clicked', async () => {
+    jest.spyOn(API, "getWebhooks").mockImplementation(() => Promise.resolve(webhooks));
```
AlanGreene: single quotes in JS, we only use double quotes for string props in JSX

webhooks-extension/src/components/WebhookBranches/tests/WebhookBranches.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+it('display notification when error occurs', async () => {
+  jest
+    .spyOn(API, 'getPipelineRuns')
+    .mockImplementation(() => Promise.reject(fetchBranchFailMock));
+
+  const { getByTestId, getByText } = renderWithRouter(
+    <WebhookBranches close={() => {}} webhook={webhook} />
+  );
+  await waitForElement(() => getByTestId(/branches-modal-notification/i));
```
AlanGreene: if the text below is contained in the same element we don't need this line and can remove the testid from the component
always prefer using-facing content as hooks for testing where possible, avoids requiring too much implementation knowledge in the tests

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
```diff
@@ -85,4 +85,16 @@ describe('with webhooks', () => {
...
+    jest.spyOn(API, "getPipelineRuns").mockImplementation(() => Promise.resolve([]));
+    const { getByText, getByTestId } = renderWithRouter(<WebhookDisplayTable match={{}} selectedNamespace="*"/>);
+
+    await waitForElement(() => getByText(/first-test-webhook/i));
+
+    fireEvent.click(getByText(/first-test-webhook/i));
+    expect(getByTestId("webhook-branches-modal")).not.toBeNull();
```
AlanGreene: can we test for the header or other content instead of testid?

webhooks-extension/src/api/index.js
```diff
@@ -89,3 +89,16 @@ export function deleteWebhooks(id, namespace, repo, deleteRuns) {
   const uri = `${apiRoot}/webhooks/${id}?namespace=${namespace}&repository=${repo}${deleteRunsQuery}`;
   return deleteRequest(uri);
 }
+
+export function getPipelineRuns({ filters = [] } = {}) {
```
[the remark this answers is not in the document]
carlos-logro: I did try to implement this but noticed that the reducer accepted the filters (which are retrieved from the URL) & not the action. Trying to carry it out this way was a little too much IMO for the goal we wanted. However, if the action can be updated to accept filters, the only param I need ATM, then this API call can be deleted.

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,209 @@
...
+  TableRow,
+  TableBody,
+  TableCell,
+  TableHeader
+} = DataTable;
+
+function getBackgroundColor(value) {
+  let style;
+  switch (value) {
+    case 'Success':
```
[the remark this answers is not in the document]
carlos-logro: nice!!! ill go ahead & emulate the example above.

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+@import '~carbon-components/scss/globals/scss/vars';
+
+#webhook-branches-modal {
+  .bx--data-table tbody td.cellText {
+    padding-left: 22.5px;
```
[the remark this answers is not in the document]
carlos-logro: Although it is a weird value, it manages to align the cell text just enough with the header text. Same thing is done within the secrets table: https://github.com/tektoncd/dashboard/blob/59c28b3d7244e6b61b29afdbcc33a65d8427c7aa/src/components/SecretsTable/Secrets.scss#L36

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,209 @@
...
+    super(props);
+    this.state = {
+      rows: [],
+      loading: true,
+      error: null
+    };
+  }
+
+  componentDidMount() {
+    let [, , url] = this.props.webhook.toLowerCase().split('|');
```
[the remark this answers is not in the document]
carlos-logro: just noticed that the namespace is not needed (which was the reason behind using this value) so ill go ahead & make the appropriate changes.

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
carlos-logro: Yeah this is the exact same blue tone used for the webhooks table.

webhooks-extension/src/components/WebhookDisplayTable/tests/WebhookTable.test.js
carlos-logro: sure!

carlos-logro:
should be merged after this PR in dashboard is merged: https://github.com/tektoncd/dashboard/pull/591

carlos-logro:
Rebased!

webhooks-extension/src/WebhookApp.js
```diff
@@ -105,7 +107,9 @@ const mapDispatchToProps = (dispatch, props) => ({
   fetchPipelines: namespace =>
     dispatch(props.actions.fetchPipelines({ namespace })),
   fetchServiceAccounts: namespace =>
-    dispatch(props.actions.fetchServiceAccounts({ namespace }))
+    dispatch(props.actions.fetchServiceAccounts({ namespace })),
+  fetchPipelineRuns: (namespace, pipelineName, filters) =>
+    dispatch(props.actions.fetchPipelineRuns(namespace, pipelineName, filters))
```
AlanGreene: the `fetchPipelineRuns` action expects an object as param

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
```diff
@@ -0,0 +1,228 @@
...
+          <InlineNotification
+            kind="error"
+            subtitle={error}
+            title="Error:"
+            lowContrast
+          />
+        )}
+        <div>
+          <p>
+            <strong>Webhook Name:</strong> {this.props.webhook.name}
```
AlanGreene: shouldn't use `<strong>` for styling like this, it's a semantic element indicating the wrapped content is important
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong

webhooks-extension/src/components/WebhookBranches/WebhookBranches.js
AlanGreene: still needs to be addressed ☝️

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
```diff
@@ -0,0 +1,49 @@
...
+    &[data-status] {
+      background-color: $inverse-support-01;
+      color: black;
+    }
+
+    &[data-status="Success"],
+    &[data-status="Succeeded"] {
```
AlanGreene: what's the difference between 'Success' and 'Succeeded'? seems odd to have both

webhooks-extension/src/WebhookApp.js
carlos-logro: i thought this was the way to go but for some reason when I surround the variables with `{}`, in the `PipelineRun` action it initializes namespace with the object passed and filters variable ends up being undefined....

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
carlos-logro: the reason behind this is that for some cases the PipelineRun was returning with `Success` status while other times with `Succeeded`

carlos-logro:
@AlanGreene made the changes, ready for another look! let me knoww thanks

webhooks-extension/src/components/WebhookBranches/WebhookBranches.scss
AlanGreene: something that should be raised in tektoncd/pipeline perhaps?

carlos-logro: probably. I double checked this with the current `PipelineRuns` I have and they are currently displaying `Succeeded`. So if @dibbles can confirm he also is getting the same then i am happy to remove `Success`

carlos-logro: @dibbles @AlanGreene its settled then! updating, let me know if anything else needs to be addressed

carlos-logro:
> @dibbles @AlanGreene its settled then! going to update the PR, let me know if anything else needs to be addressed

done!

webhooks-extension/src/WebhookApp.js
```diff
@@ -105,7 +107,9 @@ const mapDispatchToProps = (dispatch, props) => ({
   fetchPipelines: namespace =>
     dispatch(props.actions.fetchPipelines({ namespace })),
   fetchServiceAccounts: namespace =>
-    dispatch(props.actions.fetchServiceAccounts({ namespace }))
+    dispatch(props.actions.fetchServiceAccounts({ namespace })),
+  fetchPipelineRuns: (namespace, filters) =>
+    dispatch(props.actions.fetchPipelineRuns(namespace, filters))
```
AlanGreene: see https://github.com/tektoncd/dashboard/blob/c1aa3e4238e2519731bf25f3ac019e9aa0b66e5b/src/actions/pipelineRuns.js#L27

`fetchPipelineRuns` expects an object `{ namespace, filters }`, I'm not sure how this is working in its current state. We probably need some additional tests if this is getting through.

Issue #241, opened 2019-09-19
Triggers: Write Task with options to create/delete github repo webhook


dibbles:
Write a task in yaml that allows creation or deletion of a github repository webhook - code can predominantly be taken from https://github.com/tektoncd/triggers/blob/master/docs/create-webhook.yaml and then extended.

The yaml can be stored under a config/triggers_prototype folder.

a-roberts:
PR https://github.com/tektoncd/experimental/pull/248

a-roberts:
Merged so closing

PR #248, opened 2019-09-20
Add Tekton Task and TaskRun to create and delete GitHub webhooks

a-roberts:
Tested this with github.com and github.ibm.com using PATs and secrets created through the Tekton Dashboard's Webhook Extension.

This is for https://github.com/tektoncd/experimental/issues/241

# Changes

Credits to @akihikokuroda for the create webhook work this is largely based on 😄 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/config/triggers_prototype/README.md
+++ b/webhooks-extension/config/triggers_prototype/README.md
@@ -33,3 +33,42 @@ To test the Ingress set up, use the following `curl` command:
 and you should see a response back from the running EventListener pod.
 
 In this case, `-k` is to allow a self-signed certificate to be used, `-L` is to follow redirects and `-d` specifies the data to be sent.
+
+# GitHub Webhooks
+
+You can create or delete GitHub webhooks using the provided Task and TaskRun.
+
+- Apply the Task definition:
+`kubectl apply -f config/triggers_prototype/github/webhook.yaml`
+
+- Create a GitHub secret, for example with the Tekton Webhooks Extension or through `kubectl`. You will refer to the secret's name in the next step.
+
+If you opt to use `kubectl`, create the following file (for example named `secret.yaml`):
+
+```
+apiVersion: v1
+kind: Secret
+metadata:
+  name: basic-user-pass
+  annotations:
+    tekton.dev/git-0: https://github.com
+type: kubernetes.io/basic-auth
+stringData:
+  accessToken: <your access token used to access the Git provider>
+  secretToken: <anything here - it's used to validate the webhook>
+```
+
+then do `kubectl create -f secret.yaml`.
+
+Modify the TaskRun definition, replacing the parameters accordingly.
+
+To create a webhook, set `Mode` to `create`. To delete a webhook, set this to `delete`.
+
+- Apply the TaskRun definition, thus running the Task:
+`kubectl create -f config/triggers_prototype/github/webhook-run.yaml`
+
+For deleting, you'll need to specify the WebhookName parameter and can omit the ExternalUrl.
+
+To use an access token, you can specify GithubUserNameKey as "".
+
+Note that in order to not receive 503 service unavailable errors from the webhook when trying to reach your Ingress, you should ensure your Ingress is correctly configured (for example, check that the service name and port matches that of the EventListener you are using it with).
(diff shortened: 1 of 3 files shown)


a-roberts:
> Looks good to me! One question: How can the user find the webhook name (id) except from the url of each "Manage Webhook" page?

I think that's the only way at the moment - so if I've understood your point correctly, as you know the way we have it now (current UI code) is different in that we use a  named GitHubSource and  a ConfigMap and that's where we retrieve the webhook "names" from in our table. 

This is unless we were to also offer a "get webhooks for repository URL page" perhaps and surface that eventually through the UI, but I'm sure we'll be sticking with a ConfigMap so we'll let users create "named" hooks instead of using IDs. 

The create/delete code here in the yaml would remain the same though as the name has to be `web` on the POST and you delete by the ID.

a-roberts:
@dibbles could I get a look at this as well please? Aki's given us the thumbs up too so wanting to be sure this is the direction we want to run with 😸

webhooks-extension/config/triggers_prototype/github/README.md
```diff
@@ -0,0 +1,18 @@
+# GitHub Webhooks
+
+You can create or delete GitHub webhooks using the provided Task and TaskRun.
+
+- Apply the Task definition:
+`kubectl apply -f config/triggers_prototype/github/webhook.yaml`
+
+- Create a GitHub secret, for example with the Tekton Dashboard or importing it using the Tekton Webhooks Extension. You will refer to its name in the next step.
```
dibbles: I dont think we should advocate importing secrets from a git repo.  Also I would saying "using the Tekton Dashboard" rather than "with"

webhooks-extension/config/triggers_prototype/github/webhook.yaml
```diff
@@ -0,0 +1,93 @@
...
+  name: webhook-task
+  namespace: tekton-pipelines
+spec:
+  inputs:
+    params:
+    - name: Mode
+      description: Whether or not to create or delete the webhook
+      default: "create"
+    - name: ExternalUrl
+      description: The external access URL (used for creating the webhook)
```
dibbles: make this clearer i.e for example the ingress external URL in front of the EventListener

webhooks-extension/config/triggers_prototype/github/webhook.yaml
```diff
@@ -0,0 +1,93 @@
...
+      default: "replacemeip.nip.io"
+    - name: GithubOwner
```
dibbles: I wonder if this is GitHubOwnerOrOrgName rather than the Owner - could just change the description though??  Note - has this been tested against an org?

Also weird use of camel casing GitHub in descriptions but not on parameter names

webhooks-extension/config/triggers_prototype/github/webhook.yaml
```diff
@@ -0,0 +1,93 @@
...
+      default: "secretToken"
+    - name: GithubUrl
+      description: The URL of the GitHub server (used for creating and deleting the webhook)
+      default: "github.com"
+    - name: WebhookName
+      description: The webhook name to delete (as available at (your repository/hooks/<number here>)
+      default: "1234567"
+  steps:  
+  - name: create-or-delete-webhook
+    image: pstauffer/curl:latest
```
dibbles: have image licenses been checked?

dibbles:
Closing as superceded by https://github.com/tektoncd/experimental/pull/276

dibbles:
> Closing as superceded by #276

whoops - closed the wrong one!

webhooks-extension/config/triggers_prototype/github/webhook.yaml
a-roberts: Replaced param names, it's owner yeah

webhooks-extension/config/triggers_prototype/github/webhook.yaml
a-roberts: Looking at using curlimages/curl:latest as it looks to be official, see https://github.com/curl/curl-docker and note it's under the verified curl org

a-roberts:
@dibbles I've rebased on master and used an official curl image, replaced some param names so they're GitHub not Github, and tested I get a working webhook with my event listener using the latest version of Triggers (readme updated/merged in with the Ingress docs too)

Issue #242, opened 2019-09-19
Triggers: Write Task with options to create/delete Ingress


dibbles:
Write a task in yaml that allows creation or deletion of an Ingress endpoint for an event listener take code from https://github.com/tektoncd/triggers/blob/master/docs/create-webhook.yaml and then extended to allow for deletion.

The yaml can be stored under a config/triggers_prototype folder.

a-roberts:
PR https://github.com/tektoncd/experimental/pull/249

a-roberts:
Note that to get this working on Docker Desktop OK, I needed to install nginx as the Ingress controller:

```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud-generic.yaml
```
 Then curling the Ingress routes to my backend listener svc correctly.

a-roberts:
Merged so closing

PR #249, opened 2019-09-20
Add Tekton Task and TaskRun to create and delete Ingress

a-roberts:
This is for https://github.com/tektoncd/experimental/issues/242

# Changes

Again largely based on the work by @akihikokuroda [here](https://github.com/tektoncd/triggers/blob/master/docs/create-webhook.yaml)

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/config/triggers_prototype/README.md
+++ b/webhooks-extension/config/triggers_prototype/README.md
@@ -1,0 +1,33 @@
+# triggers-prototype
+
+This area is for files related to using Triggers, currently through the Tekton Webhooks Extension.
+
+## Ingress
+
+You can create or delete Ingress using the provided Task and TaskRun.
+
+- Apply the Task definition:
+`kubectl apply -f config/triggers_prototype/ingress.yaml`
+
+Modify the TaskRun definition, replacing the parameters accordingly.
+
+To create Ingress, set `Mode` to `create`. To delete Ingress, set this to `delete`.
+
+- Apply the TaskRun definition, thus running the Task:
+`kubectl create -f config/triggers_prototype/ingress-run.yaml`
+
+## Using Docker Desktop
+
+Note that in order for the EventListener service to be reachable over Ingress, you should install your own LoadBalancer - for example with:
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
+
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud-generic.yaml
+```
+
+To test the Ingress set up you can then do the curl:
+
+`curl -k -L -d 'foo' listener.<your external IP address used.nip.io -v`
+
+and you should see a response back from the running EventListener pod.
--- a/webhooks-extension/config/triggers_prototype/ingress.yaml
+++ b/webhooks-extension/config/triggers_prototype/ingress.yaml
@@ -1,0 +1,109 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: ingress-task
+  namespace: tekton-pipelines
+spec:
+  inputs:
+    params:
+    - name: Mode
+      description: Whether to create or delete Ingress
+    - name: CreateCertificate
+      description: Whether to create a certificate
+      default: "true"
+    - name: CertificateKeyPassphrase
+      description: The phrase that protects private key
+      default: "phrase1"
+    - name: CertificateSecretName
+      description: The secret name for the Ingress certificate
+      default: "secret1"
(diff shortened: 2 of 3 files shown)


webhooks-extension/config/triggers_prototype/ingress.yaml
```diff
@@ -0,0 +1,104 @@
...
+            - $(inputs.params.ExternalUrl)
+            secretName: $(inputs.params.CertificateSecretName)
+          rules:
+          - host: $(inputs.params.ExternalUrl)
+            http:
+              paths:
+              - backend:
+                  serviceName: $(inputs.params.EventListenerName)
+                  servicePort: 8082
+      EOF
```
akihikokuroda: Is this indentation correct?

a-roberts: Without it we get

```
Handling Ingress
sh: syntax error: unexpected end of file (expecting "fi")
```

So it must be, I didn't think it was either

akihikokuroda: OK.  I thought it must be same level as "cat".

webhooks-extension/config/triggers_prototype/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Note that in order for the EventListener service to be reachable over Ingress, you should install your own LoadBalancer - for example with:
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
+
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud-generic.yaml
+```
+
+To test the Ingress set up you can then do the curl:
```
dibbles: the curl ..... "a curl" ...... or better yet "use the curl command"

webhooks-extension/config/triggers_prototype/ingress.yaml
```diff
@@ -0,0 +1,109 @@
...
+      description: The secret name for the Ingress certificate
+      default: "secret1"
+    - name: ExternalUrl 
+      description: The external access URL
+      default: "https://$(inputs.params.EventListenerName).replaceme.nip.io"
+    - name: EventListenerName 
+      description: The EventListener name
+      default: "listener"
+    - name: Namespace
+      description: "The namespace to use for where the secret will be created/deleted"
```
dibbles: "The namespace where the certificate secret will be created/deleted"

webhooks-extension/config/triggers_prototype/ingress.yaml
```diff
@@ -0,0 +1,109 @@
...
+      default: "tekton-pipelines"
+  steps:
+  - name: generate-certificate
+    image: frapsoft/openssl
```
dibbles: Probably need to check licenses on the images being used

webhooks-extension/config/triggers_prototype/ingress.yaml
```diff
@@ -0,0 +1,109 @@
...
+          - hosts:
+            - $(inputs.params.ExternalUrl)
+            secretName: $(inputs.params.CertificateSecretName)
+          rules:
+          - host: $(inputs.params.ExternalUrl)
+            http:
+              paths:
+              - backend:
+                  serviceName: $(inputs.params.EventListenerName)
+                  servicePort: 8082
```
dibbles: What about having the port as an input?  I think it is supposed to be changing to port 80 to support knative at some point.

webhooks-extension/config/triggers_prototype/ingress.yaml
a-roberts: Pulled and ran interactively with
`docker run -it --entrypoint /bin/ash frapsoft/openssl`

No license file found in here
openssl binary itself, which we know is in here (obv) is version 1.0.2 from September 2016 (I hate how old this image is, but not enough that I want us to maintain our own openssl image)

```
For any release made before OpenSSL 3.0.0 (namely the 1.1.1, 1.1.0, 1.0.2, and all prior releases including those not currently supported), the dual OpenSSL and SSLeay license applies. Note that this is also true for any updates to those releases -- the "letter suffix" -- no matter when they are made. It also applies to the git branches for all those releases, and to any public forks that have not rebased to master (or 3.0.0).
```

Also, I remember doing export regulation training, openssl is definitely something you can use for encryption. Sooo we could recommend running a script we provide, assuming they've got openssl, then mount in as a secret, and have that certificate used by this task.

a-roberts: Alternative approach which I lovingly refer to as BYOC at https://github.com/tektoncd/experimental/pull/276

dibbles:
Closing as superceded by #276

Issue #271, opened 2019-10-01
No validation for access token names when creating through the webhooks creation page


Megan-Wright:
Issue in webhooks extension

Expected Behaviour:
I should not be able to create an access token called 'Key', as it contains a capital letter, and an appropriate error message should be displayed.

Actual Behaviour
I am unable to create this access token, however there is no helpful error message and rather this is displayed:

![Screenshot 2019-10-01 at 13.51.07.png](https://images.zenhubusercontent.com/5d25b016e0bbdc62005f7da6/bde8abc4-e6b2-4245-963a-cdada7951b6d)

Steps to recreate this issue:
When creating a webhook, try to add an access token with the name 'Key'.

PR #274, opened 2019-10-02
Validating Secret in Webhook Create Page

carlos-logro:
issue: #271 

# Changes
The input element for the secret name was missing the `invalid` property. Added that, a test case & also changed the approach when disabling & enabling the `create` button in the modal.

### DIFF
--- a/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
+++ b/webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
@@ -24,7 +24,7 @@ function validateInputs(value, id) {
       return false;
     }
 
-    if (/[^-.a-z0-9]/.test(trimmed)) {
+    if (!/[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/.test(trimmed)) {
       return false;
     }
   }
@@ -133,24 +133,6 @@ class WebhookCreatePage extends Component {
       document.getElementById(
         "serviceAccounts"
       ).firstElementChild.tabIndex = -1;
-    }
-  }
-
-  componentDidUpdate() {
-    if (this.props.namespaces) {
-      if (this.state.createSecretDisabled) {
-        if (this.state.newSecretName && this.state.newTokenValue) {
-          this.setState({
-            createSecretDisabled: false
-          })
-        }
-      } else {
-        if (!this.state.newSecretName || !this.state.newTokenValue) {
-          this.setState({
-            createSecretDisabled: true
-          })
-        }
-      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


carlos-logro:
Rebased!

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -286,6 +268,15 @@ class WebhookCreatePage extends Component {
     return false
   }
 
+  isCreateSecretButtonDisabled = () => {
+    return (
+      !this.state.newSecretName ||
+      !this.state.newTokenValue ||
+      this.state.invalidFields.indexOf("newSecretName") > -1 ||
```
AlanGreene: single quotes

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -725,6 +716,11 @@ class WebhookCreatePage extends Component {
                       onChange={this.handleModalText}
                       hideLabel
                       labelText="Secret Name"
+                      invalid={invalidFields.indexOf("newSecretName") > -1}
+                      invalidText={this.state.newSecretName === ""
```
AlanGreene: single quotes

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -23,7 +23,7 @@ function validateInputs(value, id) {
       return false;
     }
 
-    if (/[^-.a-z1-9]/.test(trimmed)) {
+    if (/[^-a-z0-9]/.test(trimmed)) {
```
AlanGreene: is there a reason we don't use the same regex as kubernetes?

carlos-logro: so this regex will return true if the `trimmed` value contains any invalid character while the one in Kubernetes will return true if the variable contains only valid characters. Its just the opposite but i guess we use the first simply because it is more compact and simple 😅 (same one is used in the secrets page).

carlos-logro:
@AlanGreene made the changes, ready for another look! let me know if we should revert the regex or keep the current approach. Thanks

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
AlanGreene: the regex provided by kubernetes does more than just ensure it contains valid characters, it enforces all of the rules including the fact that it can only start/end with an alphanumeric character (not `-` or `.`)

image

carlos-logro:
@AlanGreene ok, changing regex to Kubernetes one.... should I also change the one when creating a secret??

carlos-logro:
@AlanGreene rebased and regex updated.

carlos-logro:
re-hashed.

mnuttall:
Hi @carlos-logro I'm sorry just one more rebase please ?

mnuttall:
/lgtm 
thanks @carlos-logro

Issue #268, opened 2019-09-27
Webhooks extension page doesnt update when delete fails - deleted webhooks left


CarolynMabbott:
# Expected Behavior
Webhooks extension 
I delete all webhooks and they all get deleted and the page updates even if one delete fails

# Actual Behavior
I tried to delete all webhooks and two failed but the rest deleted 
The page didnt auto update with the deleted webhooks 
The page still showed the deleted webhooks being there 

# Steps to Reproduce the Problem

1. Create multiple webhooks (One of which must fail to delete, try making a webhook with the name "." without speech marks)
2. Delete all webhooks in one go
3. Should error and display only webhooks that haven't been deleted

a-roberts:
Have a hacky fix but it's only working in the case with one failing to delete webhook atm and not several, working on the test cases too.

a-roberts:
Gonna have to punt on this, have a few more pressing bugs to look at, my latest finding is that, with logs in the Go code, we have:

```
Before delete
{"level":"debug","msg":"getAllWebhooks is returning [{Name:mycluster Namespace:tekton-pipelines ServiceAccount:tekton-webhooks-extension GitRepositoryURL: myrepo AccessTokenRef:adamtoken Pipeline:simple-pipeline-insecure DockerRegistry:adamroberts HelmSecret: ReleaseName: PullTask: OnSuccessComment: OnFailureComment: GithubSource:tekton-8d55f} {Name:. Namespace:tekton-pipelines ServiceAccount:tekton-webhooks-extension GitRepositoryURL:myrepo.git AccessTokenRef:adamtoken Pipeline:simple-helm-pipeline DockerRegistry:adamroberts HelmSecret: ReleaseName: PullTask: OnSuccessComment: OnFailureComment: GithubSource:tekton-s9w89} {Name:.. Namespace:tekton-pipelines ServiceAccount:default GitRepositoryURL:myrepo.git AccessTokenRef:adamtoken Pipeline:pipeline0 DockerRegistry:efeff HelmSecret: ReleaseName: PullTask: OnSuccessComment: OnFailureComment: GithubSource:tekton-s9w89}]"}

Click to delete:
[...]

PR #277, opened 2019-10-04
Remove Successfully Deleted Webhooks when Attempting to Delete Multiples

carlos-logro:
issue: #268 

# Changes
Tried a couple of approaches, but mapping each promise by itself & then calling `Promise.all` seemed to worked the best.

Bonus:
I also made the table unselect the rows `when` the delete modal appears. Which means that if the user decides to cancel the delete call, the modal disappears and the page will look like it was refreshed. Let me know how you feel about this.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/src/WebhookApp.test.js
+++ b/webhooks-extension/src/WebhookApp.test.js
@@ -63,96 +63,6 @@ function fakeDeleteWebhooksSuccess() {
   };
 }
 
-const fakeRowsSelection = [
-  {
-    "id":"mywebhook|default",
-    "isSelected":true,
-    "isExpanded":false,
-    "cells":[
-      {
-        "id":"mywebhook|default:name","value":"mywebhook","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"name"
-        }
-      },
-      {
-        "id":"mywebhook|default:repository","value":"https://github.com/foo/bar","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"repository"
-        }
-      },
-      {
-        "id":"mywebhook|default:pipeline","value":"simple-helm-pipeline-insecure","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"pipeline"
-        }
-      },
-      {"id":"mywebhook|default:namespace","value":"default","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-      {
-        "header":"namespace"
-      }
-    }
-  ]},
-  {
-    "id":"mywebhook2|default",
-    "isSelected":true,
-    "isExpanded":false,
-    "cells":[
-      {
-        "id":"mywebhook2|default:name","value":"mywebhook2","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"name"
-        }
-      },
-      {
-        "id":"mywebhook2|default:repository","value":"https://github.com/foo/bar","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"repository"
-        }
-      },
-      {
-        "id":"mywebhook2|default:pipeline","value":"simple-helm-pipeline-insecure","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-        {
-          "header":"pipeline"
-        }
-      },
-      {"id":"mywebhook2|default:namespace","value":"default","isEditable":false,"isEditing":false,"isValid":true,"errors":null,"info":
-      {
-        "header":"namespace"
-      }
-    }
-  ]},
-  {
-    "id":"mywebhook3|default",
-    "isSelected":true,
-    "isExpanded":false,
-    "cells":[
-      {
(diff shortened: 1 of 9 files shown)


carlos-logro:
Rebased!

carlos-logro:
@AlanGreene I agree with what you said and because of that I decided to pass it to the `showDeleteDialogHandlerVisible` function where it will ultimately add it as its state. Then once the deletion class is done, `onCancel()` is called in the `finally` handle.


Now, the test is failing because apperantly it `Cannot read property 'onCancel' of undefined`.... tried to add a `jest.spyOn` but it didn't work. Any suggestions of how to bypass this problem?

carlos-logro:
@AlanGreene managed to fix the problem... apparently the test was not actually selecting the rows instead doing an api call. Removed that and updated all necessary test cases.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -134,48 +136,66 @@ export class WebhookDisplayTable extends Component {
...
+      ))
+      .then(resultsOfPromises => {
+        if (resultsOfPromises.indexOf(false) === -1 ) {
+          if (this.state.webhooks.length - rowsToUse.length === 0) {
+            this.props.setshowLastWebhookDeletedNotification(true);
+          } else {
+            this.setState({
+              showNotificationOnTable: true,
+              showDeleteDialog: false,
+              notificationStatus: "success",
```
AlanGreene: single quotes

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -134,48 +136,66 @@ export class WebhookDisplayTable extends Component {
...
-      }
-      this.setState({
-        showNotificationOnTable: true,
-        showDeleteDialog: false,
-        notificationStatus: notificationStatusToSet,
-        notificationStatusMsgShort: notificationMessageToSet,
-        notificationMessage: notificationFullMessage,
+      })
+      .finally(() => {
+        this.state.batchActions.onCancel();
```
AlanGreene: wouldn't this still unselect any remaining rows in case of error?

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
```diff
@@ -134,48 +136,66 @@ export class WebhookDisplayTable extends Component {
...
-      if(this.state.webhooks.length - rowsToUse.length === 0){
-        this.props.setshowLastWebhookDeletedNotification(true);
-      }
-      else {
+    const deletionError = () => (false);
+    const deletionSuccess = () => (true);
+
+    Promise.all(
+      deletePromises.map(promise =>
+        promise.then(deletionSuccess).catch(deletionError)
```
AlanGreene: What's the purpose of this? If the array contains false we throw an error and end up in the catch below anyway.

webhooks-extension/src/WebhookApp.test.js
```diff
@@ -225,9 +135,8 @@ const selectors = {
 };
 
 describe("change in components after last webhook(s) deleted & shows notification", () => {
-  const test = async (webhooks, fakeRowsSelected) => {
```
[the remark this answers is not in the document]
carlos-logro: `WebhookCreate/tests/` has tests regarding deletion. This test just checks that when all webhooks in table are deleted, it renders the create webhook component.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
carlos-logro: True, moving it inside `then()`.

webhooks-extension/src/components/WebhookDisplayTable/WebhookDisplayTable.js
carlos-logro: So when attempting to delete multiple webhooks and failing, the successfully deleted ones were not being removed. Therefore, I decided to check each individual promise first to see how that deletion went. Those respective promise results were deduced inside `.all()` by the calling of `then()` if successful and returning `true` or catch() if there was an error and returning `false`.  After mapping through each promise in the `.all()` call, we then check if the array returned contains a `false` which denotes that an error occurred in one of the mapped promises. If this happens, the control flow goes to the else statement and throws an Error which will ultimately make it go to the catch callback of the `.all()` method and here is were we change the state to show the error notification with the necessary info. Now, if the array is completely true then move along and check if those were all the webhooks in the table and if so, render the `CreateWebhook` component otherwise just show notification stating that the selected webhooks were deleted. 


Now i did try it without `.then(deletionSuccess).catch(deletionError)` and just having `deletePromises.map(promise => promise)` and although it works, i did notice not being as smooth and in some occasions it didn't even remove the successfully deleted webhooks.

carlos-logro:
@AlanGreene updated and rebased this PR. Let me know your opinion on the promise execution approach.

carlos-logro:
re-hashed.

mnuttall:
Hi @carlos-logro I'm sorry could you rebase this again please?

Issue #257, opened 2019-09-23
specify docker registry format when creating a webhook in tekton console


tam512:
# Expected Behavior
- When creating a new webhook in tekton console, the `Docker Registry` for `Target Pipeline Settings` should be able to accept both formats of `https://` or `without https`.  Right now it only works when entering the registry without  the `https://` such as `index.docker.io/<docker hub id>`
- If entering docker registry as `https://index.docker.io/<docker hub id>` the tekton pipeline runs will fail with error
```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "https://index.docker.io/<docker hub id>/njsloopback-sample:73a7e56": creating push check transport for https: failed: Get https://https/v2/: dial tcp: lookup https on 172.20.0.1:53: no such host
```
- If https format is not supported, then there should be a sample in the registry text box field or a clear format description when hover the mouse point over the `i` next to the Docker Registry 

# Actual Behavior
When creating a new webhook, the description for `Docker Registry` is `The docker register to push images to` as following
![Screen Shot 2019-09-23 at 2 10 01 PM](https://user-images.githubusercontent.com/22480038/65454996-e0db8480-de0b-11e9-921b-2fcd8d28523c.png)

# Steps to Reproduce the Problem

1. in tekton console, create a new webhook, the `Docker Regsistry` is in the last row of the panel

# Additional Info

mnuttall:
docker registry entries could in theory be in at least seven formats:
- http://index.docker.io/foo
- index.docker.io/foo
- foo
- https://index.docker.io/foo
- http://my.repo/foo
- https://my.repo/foo
- my.repo/foo

We need to understand which we can accept, check they work, and reject the rest.

Issue #251, opened 2019-09-20
Better description for the webhook fields - some are inaccurate


mtamboli:
# Expected Behavior

This problem is for creating webhook using Tekton Dashboard. It would help to have more helpful description for the fields.  Also, some of the description are incorrect/out of order. Screenshot pasted below.

# Steps to Reproduce the Problem
![image](https://user-images.githubusercontent.com/29490950/65340349-03fffd00-db9c-11e9-8aea-268e1ee69161.png)

1. Hover over all the fields to look for the description
2.
3.

# Additional Info

dibbles:
https://github.com/tektoncd/experimental/pull/260

mtamboli:
@dibbles Thanks for looking into this. There are couple of things which will help.

1. Having a link from the Webhooks page to the doc for explanation on each field.
2. It would help to know a bit explanation on `Access Token` field. We were expecting to be able to pick up secret defined in a separate tab. Good to explain that this is the token generated in the same git repository as repository URL.
3. Docker Registry:  Finding the correct syntax for this field was a bit of challenge. So it will be helpful to know examples of internal registry as well as docker hub. Webhook did not like if you provide https://index.docker.io/<namepspace>. It had to be index.docker.io/<docker_id>
For example:
Internal registry:  docker-registry.default.svc:5000/<namespace>
Docker Hub: index.docker.io/<docker_id>

We were also not sure if the image name need to be provided or not. So having some example will certainly help.

mtamboli:
@a-roberts I use tekton dashboard which is installed as part of Kabanero. I will need to get access to updated version  of dashboard to verify it. If I am not able to verify, I will close it and reopen later if there are any new issues.

PR #287, opened 2019-10-09
Update Webhooks Tooltips & Getting Started Guide

carlos-logro:
Issues: #251 #257 

# Changes
Update tooltips to be more descriptive/helpful as well as add link to docs. In the GettingStarted.md accepted docker registry formats were added as well as a link to a step by step guide on how to generate a access token. Lastly, trim `https://` or `http://` from the backend. 

# Submitter Checklist

review them:

for more details._


webhooks-extension/pkg/endpoints/webhook.go
```diff
@@ -66,6 +66,9 @@ func (r Resource) createWebhook(request *restful.Request, response *restful.Resp
 	}
 
 	dockerRegDefault := r.Defaults.DockerRegistry
+	// remove prefixes if any
+	webhook.DockerRegistry = strings.TrimPrefix(webhook.DockerRegistry, "https://")
```
a-roberts: This is a bit funky, we're stripping things here but in the getting started doc we have the prefix - so what's the deal with what's accepted and what isn't?

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -539,7 +542,7 @@ class WebhookCreatePage extends Component {
 
             <div className="row">
               <div className="help-icon" id="git-tooltip">
-                <CustomTooltip tooltipText="The URL of the git repository which you want to trigger a pipeline."/>
+                <CustomTooltip tooltipText="The URL of the GitHub repository to create a webhook on for the specified Pipeline."/>
```
a-roberts: +1 so folks in here don't think it's any old Git e.g. GitLab

webhooks-extension/pkg/endpoints/webhook.go
carlos-logro: so in theory we don't accept registries with `http` or `https` but, by stripping them in the back end, we can let the end user add whichever prefixes they want. Should i clarify this in the doc?

PR #290, opened 2019-10-14
Add GitHub interceptor and TaskRun

a-roberts:
Largely based on https://github.com/tektoncd/triggers/pull/144/files

Eventually we shouldn't refer to the validation server code at my repository

I could add the validation server code to go with this commit or we can have it a different repository. The validation code (currently on GitHub) can be found at https://github.com/a-roberts/tekton-validate-github-event.git at the **inceptor** branch

@dibbles for discussion/integration please 😄


a-roberts:
Note that after discussing this with @vtereso, this may suffer from a limitation we have to use the same access token for all of the EL triggered templates (or the interceptor would fail). So perhaps we should be passing in the secret name as an interceptor header and not being mounted in...TBC

a-roberts:
Added Wext-Secret-Name and logs indicate all is well, ready for another look please

```
adams-mbp:interceptor aroberts$ k logs -f myinterceptorservice-857b998d58-db6sm
2019/10/14 15:56:58 Clone URL coming in as json: https://mygithub.bar.com/AROBERTS/devops-world-example.git
2019/10/14 15:56:58 Handling GitHub Event with deliveryID: 3fdf4ff8-ee9b-11e9-8c5b-4860b84ebaf5
2019/10/14 15:56:58 Validation pass (repository URL and secret payload checked)
2019/10/14 15:56:58 Validation passed so writing response
```

webhooks-extension/config/interceptor-deployment.yaml
```diff
@@ -0,0 +1,20 @@
...
+    metadata:
+      labels:
+        app: tekton-webhooks-extension-validator
+    spec:
+      containers:
+        - name: validate
+          image: adamroberts/github-validate-task:latest
+          env:
+            - name: INSTALLED_NAMESPACE
+              fieldPath: metadata.namespace
```
dibbles: - name: INSTALLED_NAMESPACE
  valueFrom:
     fieldRef:
       fieldPath: metadata.namespace

webhooks-extension/config/interceptor-deployment.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tekton-webhooks-extension-validator
```
dibbles: we should add namespace: tekton-pipelines

webhooks-extension/config/interceptor-service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: tekton-webhooks-extension-validator
```
dibbles: we should add namespace: tekton-pipelines

a-roberts:
Pushed - now with enhanced logging and feedback addressed

a-roberts:
One suggestion is to add the trigger name in the validator to be printed, I can add this after (dev environment occupied by https://github.com/tektoncd/dashboard/issues/638 atm)

dibbles:
/lgtm
/approve

need to modify location of image etc.... but will change when moving to full triggers implementation

Issue #279, opened 2019-10-07
Improve Ingress task to use CSR instead of requiring upfront cert creation


a-roberts:
# Expected Behavior
Instead of creating one, e.g. through the provided helper script, we could generate one and have it approved in the cluster based on info at https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#create-a-certificate-signing-request
 and https://kubernetes.io/docs/concepts/cluster-administration/certificates/, which mentions
```
Certificates API
You can use the certificates.k8s.io API to provision x509 certificates to use for authentication as documented here.
```
# Actual Behavior
Users must create one upfront, perhaps using the secret, then reference the secret name

# Steps to Rep

akihikokuroda:
Yes, this is similar to what we did for the other project.

akihikokuroda:
This one is using the certmanager.  The api might be changed but it should be similar.
```
  replace_jenkins_ingress.sh: |
    #!/bin/bash
    cn={{ template "jenkins.ingress-host" . }}
    cd
    mkdir $cn.ssl
    cd $cn.ssl
    csr=$(pwd)/$cn.csr
    key=$(pwd)/$cn.key
    crt=$(pwd)/$cn.crt
    openssl genrsa -out $key 2048
    openssl req -new -sha256 -key $key -out $csr -subj "/C=US/ST=NC/O=IBM Inc./CN=$cn" \
        -reqexts SAN \
        -config <(cat /etc/ssl/openssl.cnf <(printf "[SAN]\nsubjectAltName=DNS:${cn},DNS:jenkins.127.0.0.1.nip.io"))
    cd ..
    (cat <<EOF
    apiVersion: certificates.k8s.io/v1beta1
    kind: CertificateSigningRequest
    metadata:
      name: ${cn}
    spec:
      groups:
      - system:authenticated
      request: $(cat $csr | base64 | tr -d '\n')
      usages:
      - digital signature
      - key encipherment
      - server auth
    EOF
    ) | kubectl create -f - 2>output.txt
    cat output.txt
    var=$(grep "error" output.txt)
    if [[ !  -z  "$var"  ]]
    then
     openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout tls.key -out tls.crt -subj "/CN={{ template "jenkins.ingress-host" . }}"
     kubectl create secret tls {{ .Release.Name }}-tls-secret  --key tls.key --cert tls.crt
    else 
[...]

PR #295, opened 2019-10-17
add root signed certificate to ingress

akihikokuroda:
# Changes

This PR is for https://github.com/tektoncd/experimental/issues/279

The ingress task has 2 more steps
- generate a certificate and the CSR
- get CSR signed and create the secret with it

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/config/triggers_prototype/ingress.yaml
+++ b/webhooks-extension/config/triggers_prototype/ingress.yaml
@@ -32,7 +32,7 @@ spec:
       type: string
   steps:
   - name: generate-certificate
-    image: jordi/openssl
+    image: cfssl/cfssl
     volumeMounts:
     - name: work
       mountPath: /var/tmp/work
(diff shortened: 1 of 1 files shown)


akihikokuroda:
Necessary RBAC must be documented.

create certificatesigningrequests.certificates.k8s.io
create secrets

webhooks-extension/config/triggers_prototype/ingress.yaml
```diff
@@ -28,6 +31,106 @@ spec:
       default: "8082"
       type: string
   steps:
+  - name: generate-certificate
+    image: cfssl/cfssl
```
a-roberts: Changes look good aside from export clearance stuff for shipping this (used in a Tekton Task I know) for products, I don't think this changes how Duane's code works (but we can then tidy up the readme and say you either use this task, or it's done through code, *or* you use the original "bring your own cert method)

I see this as a much better image to use than the non-official openssl images though 👍

akihikokuroda: @a-roberts What do I need for this?  Export clearance?

a-roberts: Well I was first concerned around using OpenSSL in a Tekton task as it does have encryption capabilities I reckon (and one can get in biiiig trouble for not following the procedures), and because our goodies make their way into products, it's not clear on how we should proceed from a legal perspective really. 

Everything else is great, it's just this bit we should clear up (perhaps leveraging IBM legal teams and management and this is of use to anybody else that wants to use this task too - because you would end up with that image on your machine and there are procedures to follow)

akihikokuroda:
The ingress creation has been moved from the task to the go code.  So I close this and create another PR.

PR #302, opened 2019-10-22
Fix task monitor by adding copy step

vtereso:
# Changes
`PipelineResources` were internally modified, which broke the task monitor on Tekton 0.7 (as well as 0.8). This adds a copy step so that the pull-request resource properly uploads when used within the task monitor
# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/config/task-monitor-result.yaml
+++ b/webhooks-extension/config/task-monitor-result.yaml
@@ -150,3 +150,10 @@ spec:
       handle.write(status)
       handle.close()
       EOF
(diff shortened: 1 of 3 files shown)


a-roberts:
@mnuttall we could do a 0.2.1 with the above change in and anything else we need for supporting OpenShift 4.2/bugs, I think this is critical enough a bug to have resolved

a-roberts:
Should make the same commit for config/triggers_prototype/task-monitor-result.yaml as well

PR #303, opened 2019-10-23
Use ALL_NAMESPACES for webhooks-extension state

### LINKED ISSUE #266: Webhooks extension does not show pipeline or service account when namespace on the left is different to the one selected
When creating a webhook you cannot select a pipeline or service account if the namespace you have selected in the create webhook is different to the namespace on the left

ncskier:
# Changes

Fixes #266
The `WebhookApp` was taking its state directly from the Redux store using
the Dashboard's selected namespace (`getSelectedNamespace()`). Instead, we
want to use `ALL_NAMESPACES` so that the `WebhookApp`'s state does not
change when the Dashboard's selected namespace changes.

Also, this PR adds an empty state to the Pipeline and ServiceAccount dropdowns.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/src/WebhookApp.js
+++ b/webhooks-extension/src/WebhookApp.js
@@ -5,6 +5,8 @@ import { connect } from 'react-redux';
 import { WebhookCreate } from './components/WebhookCreate';
 import { WebhookDisplayTable } from './components/WebhookDisplayTable';
 
+const ALL_NAMESPACES = '*';
+
 class WebhooksApp extends Component {
 
   constructor(props) {
@@ -91,13 +93,13 @@ function mapStateToProps(state, props) {
   return {
     namespace: props.selectors.getSelectedNamespace(state),
     namespaces: props.selectors.getNamespaces(state),
-    pipelines: props.selectors.getPipelines(state),
+    pipelines: props.selectors.getPipelines(state, { namespace: ALL_NAMESPACES }),
     isFetchingNamespaces: props.selectors.isFetchingNamespaces(state),
     isFetchingPipelines: props.selectors.isFetchingPipelines(state),
     pipelinesErrorMessage: props.selectors.getPipelinesErrorMessage(state),
     serviceAccountsErrorMessage: props.selectors.getServiceAccountsErrorMessage(state),
     isFetchingServiceAccounts: props.selectors.isFetchingServiceAccounts(state),
-    serviceAccounts: props.selectors.getServiceAccounts(state)
+    serviceAccounts: props.selectors.getServiceAccounts(state, { namespace: ALL_NAMESPACES })
   };
 }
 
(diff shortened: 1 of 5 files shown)


webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
```diff
@@ -352,7 +352,7 @@ class WebhookCreatePage extends Component {
     return <Dropdown
```
AlanGreene: We have a ServiceAccountsDropdown component that already provides an empty state, might be worth switching to that if it's suitable. Alternatively you can use the TooltipDropdown directly which has an `emptyText` prop for this purpose.

webhooks-extension/Gopkg.lock
```diff
@@ -90,6 +90,22 @@
   pruneopts = "UT"
   revision = "6f77996f0c42f7b84e5a2b252227263f93432e9b"
 
+[[projects]]
```
AlanGreene: please keep dependency updates in separate PRs unless related to the code change, this makes it easier to identify potential issues introduced by the new versions

webhooks-extension/src/WebhookApp.js
```diff
@@ -5,6 +5,8 @@ import { connect } from 'react-redux';
 import { WebhookCreate } from './components/WebhookCreate';
 import { WebhookDisplayTable } from './components/WebhookDisplayTable';
 
+var ALL_NAMESPACES = '*';
```
AlanGreene: use `const` instead of `var`

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
ncskier: I actually created issue https://github.com/tektoncd/experimental/issues/304 after working on this one for the same reason as your comment here 😅 

@AlanGreene it was my understanding that the components from the Dashboard weren't currently shared with the Extension, so please let me know if this is incorrect, and if so, how I can use those components.

AlanGreene: We publish an npm package with the UI components (https://www.npmjs.com/package/@tektoncd/dashboard-components) - we still need to add documentation and automate the release process but the components should be usable.

However, I forgot that the ServiceAccountsDropdown isn't included as it's a container (connected to the store) 😢 It would be relatively simple to move the TooltipDropdown to the components package though.

ncskier: Ok, I'll hold off on updating the components in this PR, and leave it to issue #304 if that's alright with you @AlanGreene.

ncskier:
Thanks for the feedback @AlanGreene. I think I've addressed all of your comments, but let me know if there's anything else I can do 🙂

webhooks-extension/src/components/WebhookCreate/WebhookCreate.js
ncskier: Is it possible to share the Containers since they're connected to the store, or only the components?

AlanGreene: It would be possible although a little awkward, and likely of limited value. The containers make assumptions that may not be valid for 3rd party consumers outside the dashboard application and actually using these published containers would probably require similar amount of code as writing your own using the existing UI components.

PR #306, opened 2019-10-24
Add verified OpenShift 4.2 install and usage instructions

a-roberts:
For https://github.com/tektoncd/dashboard/issues/648


a-roberts:
@mnuttall here's my first attempt, please take a look

PR #311, opened 2019-10-30
plgen: tekton DSL based on Dockerfile

gireeshpunathil:
# Changes

`plgen` is a tekton pipeline generator from simple verbs that align to `Dockerfile` syntax

The objective of plgen is to provide a high level DSL for generating tekton pipeline from simple verbs that align to Dockerfile syntax, hide the complex kubernetes semantics, and thereby improve the overall user experience with tekton pipelines.

 - the Dockerfile verbs `ARG, FROM, RUN, LABEL, ENV, MOUNT`, and `USER` are used in the generator without losing much of their original meaning
 - as we are not using any Dockerfile parser, we are free to deviate when situation demands it
 - my idea of covering more scenarios is to wait for some end user feedback, set the semantics straight, and then specialize on those, however I am open to suggestions

thanks to those who provided vital feedback - @ImJasonH @brandonkal @mnuttall @vdemeester

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/plgen/gen.go
+++ b/plgen/gen.go
@@ -1,0 +1,511 @@
+// Copyright © 2019 IBM Corporation and others.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v2"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// Master structure that represents the entire pipeline
+type PlGen struct {
+	pr          PipelineResource
+	pspecs      Spec
+	psteps      []Steps
+	plt         PipelineTask
+	pl          Pipeline
+	role        Role
+	rolebinding RoleBinding
+	plr         PipelineRun
+}
+
+// Used by PipelineResource,
+type Metadata struct {
+	Name string `yaml:"name"`
+}
+
+// Used by PipelineResource, through PipelineResource->Items->PipelineResourceSpec
+type PipelineResourceParams struct {
+	Name  string `yaml:"name"`
+	Value string `yaml:"value"`
+}
+
+// Used by PipelineResource, through PipelineResource->Items
+type PipelineResourceSpec struct {
+	Type   string                   `yaml:"type"`
+	Params []PipelineResourceParams `yaml:"params"`
+}
+
+// Used by PipelineResource, top level to PipelineResource
+type Items struct {
+	APIVersion string               `yaml:"apiVersion"`
+	Kind       string               `yaml:"kind"`
+	Metadata   Metadata             `yaml:"metadata"`
+	Spec       PipelineResourceSpec `yaml:"spec"`
+}
+
+// Used by PlGen, top level.
+type PipelineResource struct {
+	APIVersion string  `yaml:"apiVersion"`
+	Items      []Items `yaml:"items"`
+	Kind       string  `yaml:"kind"`
+}
+
+// Used by PlGen, top level
+type Role struct {
+	APIVersion string   `yaml:"apiVersion"`
+	Kind       string   `yaml:"kind"`
+	Metadata   Metadata `yaml:"metadata"`
+}
+
+// Used by RoleBinding, through Rolebinding
+type Subjects struct {
+	Kind      string `yaml:"kind"`
+	Name      string `yaml:"name"`
+	Namespace string `yaml:"namespace"`
+}
+
+// Used by RoleBinding, through Rolebinding
+type RoleRef struct {
+	Kind     string `yaml:"kind"`
+	Name     string `yaml:"name"`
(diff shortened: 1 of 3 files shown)


gireeshpunathil:
for those who want to view the plgen in action without compiling and running it:

A targz pipeline can be encoded as below:

```Dockerfile
LABEL targz
USER core
FROM ubuntu
ARG input=http://github.com/foo/archive.tar.gz
ENV foo=bar
RUN tar xzvf $input
RUN cat source/file.txt
```


which gets (currently) translated to:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: core
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: admin
subjects:
- kind: ServiceAccount
  name: core
  namespace: default
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: v1
items:
- apiVersion: tekton.dev/v1alpha1
  kind: PipelineResource
  metadata:
    name: input
  spec:
    type: git
    params:
    - name: resource0
      value: http://github.com/foo/archive.tar.gz
kind: List
---
apiVersion: tekton.dev/v1alpha1
kind: Pipeline
metadata:
  name: l-txt-pipeline
spec:
  resources:
  - name: input
    type: git
  tasks:
  - name: l-txt
    taskRef:
      name: l-txt-task
      kind: ""
    resources:
      inputs:
      - name: input
        resource: input
---
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: l-txt-task
spec:
  inputs:
    resources:
    - name: input
      type: git
  outputs: {}
  steps:
  - name: targz
    image: ubuntu
    env:
    - name: foo
      value: bar
    command:
    - /bin/bash
    args:
    - -c
    - tar xzvf http://github.com/foo/archive.tar.gz
    - -c
    - cat source/file.txt
---
apiVersion: tekton.dev/v1alpha1
kind: PipelineRun
metadata:
  name: l-txt-pipeline-run
spec:
  serviceAccount: core
  timeout: 1h0m0s
  pipelineRef:
    name: l-txt-pipeline
  trigger:
    type: manual
  resources:
  - name: input
    resourceref:
      name: input
---
```

gireeshpunathil:
bump. what it takes to help move this forward? thanks.

gireeshpunathil:
nearing 3 months with no review - not a great way to motivate new contributors! let me know what is the blocker here:

 - too much code to review? I can present it in one of the forums you suggest
 - lacking a description / flow? I have attempted to sketch one, but pls suggest if needs more
 - lacking review time? I am happy to wait, but pls state so!

vdemeester:
Hey @gireeshpunathil, sorry for the loooong wait :sweat: 

>     * too much code to review? I can present it in one of the forums you suggest
>     * lacking a description / flow? I have attempted to sketch one, but pls suggest if needs more
>     * lacking review time? I am happy to wait, but pls state so!

It's definitely a lack of review time (and focusing on other subject, like the soon-to-be beta api of tektoncd/pipeline, …).

On the "high-level" approach, I wonder how much using `Dockerfile` as a DSL for `Task`/`Pipeline` make sense :

- `Dockerfile` is really limited in what it can do : there is no graph involved, and there is some command in it (like `USER`) that are weirdly translatable in a `Task`.
- `Dockerfile` are meant to generate an `oci-image` in the end (and most of the time are written this way) and this assumption is very limited to the translation into `Task` (your experiment doesn't seem to generate images).


There is a few weird thing in the generated yamls, that I don't think work

-  the generated `PipelineResource` would validate, the param `resource0` is not supported by a `PipelineResource` of type `git`
```
apiVersion: v1
items:
- apiVersion: tekton.dev/v1alpha1
  kind: PipelineResource
  metadata:
    name: input
  spec:
    type: git
    params:
    - name: resource0
      value: http://github.com/foo/archive.tar.gz
```

- `/bin/bash -c … -c …` only execute the first (as far as I know)
```
    command:
    - /bin/bash
    args:
    - -c
    - tar xzvf http://github.com/foo/archive.tar.gz
    - -c
    - cat source/file.txt
```

gireeshpunathil:
@vdemeester  - thanks for the review!

> Dockerfile is really limited in what it can do : there is no graph involved, and there is some command in it (like USER) that are weirdly translatable in a Task.

 - my premise is that pipeline tasks are imperative in nature, that fits the Dockerfile philosphy
 - I kind of agree, the existing Dockerfile verbs are limited, but `plgen` parser is extensible to fit pipeline needs
 - it allows to define and standardize new verbs that are most natural for the pipeline tasks.
 - pipeline tasks themselves are not too complex, and can be seen as well scoped in a minimal set of vocabulary


> Dockerfile are meant to generate an oci-image in the end (and most of the time are written this way) and this assumption is very limited to the translation into Task (your experiment doesn't seem to generate images).

 - of course, we are not using the syntax to create images, instead for defining pipelines
 - Dockerfile is an analogy, meant for fast ramp up and easy representation
 - the analogy should become evident to any consumer on the first use
 - software programming domain is filled with such analogies: `print`, `thread`, `kill`, `gc`, `throw`, `native` ... no user will associate these keywords with their original meaning, beyond the first use IMO.


> There is a few weird thing in the generated yamls, that I don't think work

 - agreed, I can get back to my branch, familiarize with the code (it has been some time) and fix those; but given that there are design level questions that needs to be ratified; I will delay the bug fixing until then.


In general, I appeal to contemplate this approach as independent from `yaml` semantics as possible - In my experience, the problem of comparing with an established baseline is that:

 - the gaps of comparing entity from the baseline gets highlighted more
 - any novelty in the comparing entity gets masked off, as those don't appear for comparison, because those are missing in the baseline.

PR #321, opened 2019-11-05
Remove tekton-listener from experimental (has become Tekton Triggers)

a-roberts:
😄  cleaning up, noticed https://github.com/tektoncd/experimental/pull/215 which is going into what I assume is now an obsolete area since the graduation

@iancoffey @bobcatfish @vdemeester


a-roberts:
> LGTM, thanks for the memories tekton-listener! 👍 😄

Thanks for taking a look! Sorry, think I've invalidated that lgtm as it turns out our webhooks extension area was in a bit of state. Worked some rebase magic and things should start to look better now 😄 if not I'll say something again here

bobcatfish:
Thanks for the cleanup @a-roberts , for all the hard work @iancoffey and for the memories tekton-listener :D

image

PR #323, opened 2019-11-05
[Testing] Nothing changed - look for the hashes on prow

a-roberts:
Used for checking the current hashes and to see if we fail on them being incorrect


a-roberts:
Got it, PR to update the hash https://github.com/tektoncd/experimental/pull/324 so closing this

PR #326, opened 2019-11-07
Add troubleshooting for OCP 4.2 route access with a member roll

a-roberts:
Found as part of https://github.com/kabanero-io/kabanero-foundation/issues/126

Useful to comment here for users that aren't installing us through Kabanero


a-roberts:
We should actually mention this as an install requirement, not in the troubleshooting (it'll never occur otherwise), so hold this

/hold

mnuttall:
Needs to be repurposed to the v0.2.1 branch please.

a-roberts:
Committed to https://github.com/tektoncd/experimental/blob/v0.2.1/webhooks-extension/docs/GettingStarted.md directly, closing

PR #348, opened 2019-11-13
Use interceptor folder for nightly builds

a-roberts:
Should result in the interceptor image being built and published, thus available in our published builds.


mnuttall:
https://api.thecatapi.com/api/categories/list promised me sinks. I got a butterfly. 3/7.

PR #351, opened 2019-11-13
Use go code to create/delete webhook

### LINKED ISSUE #327: Use a go library/3rd party open source code to create webhook
Currently the webhook is created using a Task - this is time consuming and difficult to monitor.  Change to using a go library to create the webhook in github and take cleanup actions if a failure occurs.

For example, if the webhook has been added to the eventlistener and then creation of the webhook fails, we should also remove the entry from the event listener.

Remove the Task yaml from the config directory and all the yaml files in which it is embedded.  Test code can also be removed from shared-test-funcs that creates/returns dummy reaults for taskruns.

ncskier:
# Changes
Fixes #327
Use go code to create/delete webhooks instead of using a TaskRun.
To create/delete the webhook, we subscribe/unsubscribe using the GitHub
PubSubHubbub API. More info about the API here:
https://developer.github.com/v3/repos/hooks/#pubsubhubbub

# Submitter Checklist

review them:

for more details._


dibbles:
yes - remove the task and also delete using the git client

webhooks-extension/pkg/endpoints/github.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// getGitHubHubbubAPI returns the API URL for the GitHub PubSubHubbub API
+func getGitHubHubbubAPI(u *url.URL) string {
+	// Public GitHub PubSubHubbub API URL is "https://api.github.com/hub"
+	hubbubAPI := "https://api.github.com/hub"
+
+	// Enterprise GitHub API URL is "https://github.company.com/api/v3/hub"
+	if company := getGitHubEnterpriseCompany(u); company != "" {
+		hubbubAPI = fmt.Sprintf("https://github.%s.com/api/v3/hub", company)
```
dibbles: Implies if you use github enterprise your server api must be github.company.com/api/v3  ...... could a company not use any dns hostname they want?  Is this a guideline given by GitHub for GHE installs?  Do we need to consider alternatives?

ncskier: Thanks Duane, good catch! I've updated the code as well as the tests to catch these cases (https://github.com/tektoncd/experimental/pull/351/files#diff-e4756d2adeff9f4613bce3d5106b8b37R108). Please let me know if there's anything else I missed!

ncskier:
Thanks for the review Duane! I've removed all instances of that create-webhook Task, and I've updated the git URL processing to catch the GHE case you mentioned.

Also, here's the issue I created for moving from the PubSubHubbub API to the Hooks API: https://github.com/tektoncd/experimental/issues/362.

Let me know if there are any more changes I can make.

Issue #250, opened 2019-09-20
Pull Request webhook should only fire on creation(?)


dibbles:
# Expected Behavior

Pipelines should not be fired when pull request closed, labelled, assigned etc... Only on creation or underlying code updated 

# Actual Behavior

Pipelines triggered on any pull request action

# Steps to Reproduce the Problem

1. configure webhook
2. create pull request (pipelines triggered)
3. perform some actions on pull request eg. close (pipelines triggered)

# Additional Info

mnuttall:
The GitHub 'Pull Request' category includes, 
> Pull request opened, closed, reopened, edited, assigned, unassigned, review requested, review request removed, labeled, unlabeled, or synchronized.
We only want a PipelineRun on a few of these events. New PipelineRuns are required when a PR is opened, or new code is pushed to its associated branch, or a closed PR is reopened. We don't want to address this issue until https://github.com/tektoncd/experimental/issues/247 is delivered and we're using Triggers to initiate webhook'd pipelines. At that point we have two options:
- Filter unwanted PR events out in Validation, or
- Filter unwanted PR events out in Pipelines. 

Doing this in validation is probably ok - although it could give us a GitHub vs GitLab problem down the road.

vtereso:
With the move to Triggers (under the current architecture), we need to have a validator/interceptor for events. With multiple interceptors, we could have authentication for one inteceptor and repo/event matching in other. A single interceptor could also do this all in one, but would be less modular (in case this feature isn't available when this is picked up). Although we could have one large service to authenticate for both GitHub and GitLab (or any other provider), separating them might be easier to manage. In any case, once we tackle this I don't think we should have any blockers.

akihikokuroda:
Here is the doc for the pullrequest action. 
 https://developer.github.com/v3/activity/events/types/#pullrequestevent

akihikokuroda:
Check the `action` of the pullrequest and
process: `opened`,  `reopened`, ignore: `assigned`, `unassigned,` `review_requested`, `review_request_removed`, `labeled`, `unlabeled`, `edited`, `closed`, `ready_for_review`, `locked`, `unlocked`

> new code is pushed to its associated branch

 should be in the pushreqeust

AlanGreene:
Should we also include `synchronize` (when PR commits are updated)?

akihikokuroda:
Probably,  I don't know what this mean yet, though.

[...]

PR #354, opened 2019-11-14
check actions to process in the interceptor

akihikokuroda:
# Changes
This PR is for https://github.com/tektoncd/experimental/issues/250

The pull request is processed only when the action is one of opened, reopened or synchronize.

The event listener sends the actions to be processed in the header to the interceptor.  The interceptor checks if the action in the event matches one of the actions sent by the event listener if the actions are not empty.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/cmd/interceptor/interceptor.go
+++ b/webhooks-extension/cmd/interceptor/interceptor.go
@@ -31,6 +31,7 @@ const (
 )
 
 type Result struct {
+	Action string `json:"action"`
 	Repository struct {
 		CloneURL string `json:"clone_url"`
 	} `json:"repository"`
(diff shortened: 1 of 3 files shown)


webhooks-extension/cmd/interceptor/interceptor.go
```diff
@@ -103,8 +104,19 @@ func main() {
...
+					wantedActions := request.Header["Wext-Incoming-Actions"]
+					if len(wantedActions) == 0 {  
+						validationPassed = true
+						log.Printf("[%s] Validation PASS (repository URL, secret payload, event type checked)", foundTriggerName)
+					} else {
+						actions := strings.Split(wantedActions[0], ",")
+						for _, action := range actions {
+							if action == result.Action {
+								validationPassed = true
+								log.Printf("[%s] Validation PASS (repository URL, secret payload, event type, action checked)", foundTriggerName)
```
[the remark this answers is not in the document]
akihikokuroda: @ncskier Thanks for review.  I updated it.

webhooks-extension/cmd/interceptor/interceptor.go
```diff
@@ -113,8 +114,19 @@ func main() {
...
+						validationPassed = true
+						log.Printf("[%s] Validation PASS (repository URL, secret payload, event type checked)", foundTriggerName)
+					} else {
+						actions := strings.Split(wantedActions[0], ",")
+						for _, action := range actions {
+							if action == result.Action {
+								validationPassed = true
+								log.Printf("[%s] Validation PASS (repository URL, secret payload, event type, action:%s checked)", foundTriggerName, action)
+							}
+						}
```
ncskier: I think that this change surfaces a bug in the existing interceptor code. When `validationPassed` is `false`, it is never written as an error. On line 155 (https://github.com/tektoncd/experimental/pull/354/files#diff-4e40b3eebcbb16692d761b108f341566R155) we should have an `else` statement to catch the case when `validationPassed` is false. For example:
```go
if validationPassed {
	log.Printf("[%s] Validation PASS so writing response", foundTriggerName)
	_, err := writer.Write(payload)
	if err != nil {
		log.Printf("[%s] Failed to write response for Github event ID: %s. Error: %s", foundTriggerName, id, err.Error())
		http.Error(writer, fmt.Sprint(err), http.StatusInternalServerError)
		return
	}
} else {
	http.Error(writer, "Validation failed", http.StatusExpectationFailed)
}
```

akihikokuroda: @ncskier Good catch!!  I fix it.  Thanks!

PR #382, opened 2019-12-03
Add a suggested image tag to the webhook event payload

dibbles:
# Changes

Issue 377: This change adds a new property into the webhook body
payload by the interceptor.  The new property is
"webhooks-tekton-image-tag" and can be used in the trigger
bindings file(s).

If the push of a git tag causes the webhook to trigger, this property
will be set to the value/label of the tag.  In other cases, the
property is set to the short commit id.

# Submitter Checklist

review them:

for more details._


dibbles:
I need to add some docs....

Issue #395, opened 2019-12-11
Deleting the pipelinerun causes pull request status to be successful


dibbles:
# Expected Behavior

When a pull request is created, I would expect the status of the pull request to be "failed" or "error", if I delete the pipelinerun whilst it is running.

# Actual Behavior

The status is success and the table of pipelineruns simply shows the table headers and no pipelineruns.

# Steps to Reproduce the Problem

1. create webhook
2. create pull request to repo
3. when pipelinerun and monitor task are running, delete the pipelinerun

# Additional Info

a-roberts:
Reproduced oddities here, deleted the running PipelineRun for my pull request as mentioned in the recreate.

I noticed a comment was made saying the PipelineRun had failed, then it went back to pending, and then I commented with just the table headers, so here's what I end up with:

![image](https://user-images.githubusercontent.com/12814972/70702809-cc01e500-1cc6-11ea-89b8-552690233c74.png)

![image](https://user-images.githubusercontent.com/12814972/70702848-e1770f00-1cc6-11ea-8534-4a5966203222.png)

a-roberts:
One way I think we can fix this is with a simple else for the if:

`if len(api_response["items"]) > 0:`

We can set a flag and then only write the comment (the table) in this scenario. 

Unfortunately to complete my testing of this we really want https://github.com/tektoncd/experimental/issues/404 fixed, which I found while implementing this. For this code change we've agreed on using `Error` for the status and I propose `No PipelineRuns found!` for the description.

a-roberts:
So far I have this

![image](https://user-images.githubusercontent.com/12814972/70816767-1b7c0a00-1dc8-11ea-9e70-85f1d39c21a3.png)

which involves a customisable missing comment (I've got a German translation example too that compliments the others)

Testing the scenario now where the monitor task is looking at multiple PipelineRuns and only the one is deleted, so we should have a table with 

Missing
Success

I believe this involves setting up multiple pipelines on the same repo so they will have the same trigger ID.

a-roberts:
So far I have this

![image](https://user-images.githubusercontent.com/12814972/70816767-1b7c0a00-1dc8-11ea-9e70-85f1d39c21a3.png)

which involves a customisable missing comment (I've got a German translation example too that compliments the others)

Testing the scenario now where the monitor task is looking at multiple PipelineRuns and only the one is deleted, so we should have a table with 

Missing
Success

[...]

PR #408, opened 2019-12-13
Handle missing PipelineRuns for monitor task

a-roberts:
For https://github.com/tektoncd/experimental/issues/395

Also adds override for the missing comment

Testing is:
- reopen two PullRequests thus triggering two PipelineRuns
- delete one PipelineRun, check table reflects state accordingly

Also:
- reopen two PullRequests thus triggering two PipelineRuns
- delete both PipelineRuns, check table reflects state accordingly

### DIFF
--- a/webhooks-extension/base/400-monitor-task.yaml
+++ b/webhooks-extension/base/400-monitor-task.yaml
@@ -140,10 +140,8 @@ spec:
                  "Status | Pipeline | PipelineRun | Namespace\n"
                  ":----- | :------- | :--------------- | :--------\n"
                  ) + "\n".join(results)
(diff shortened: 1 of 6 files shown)


a-roberts:
Started work again on this today with a simple diff function and failing the PR if one or more PIpelineRuns do go missing, it was awkward to test so I've made some notes as comments in the file too

image

dibbles:
Looks fine ..... I would possibly ask for a different description for the message. `The text of the missing comment` could be better expressed as the text to use in the situation where a pipelineRun cannot be found .... or some such.  

Up to you though .... I know you just followed the previous description formats.  Let meknow when you want a "formal" review added (fix failure)

PR #418, opened 2020-01-07
tekdoc :: parse input file to yaml

chandanikumari:
# Changes

# Submitter Checklist

review them:

for more details._


wlynch:
/lgtm

@ImJasonH I don't see any obvious reason why the build tests are failing. Any ideas?

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
-	_ "github.com/tektoncd/pipeline/pkg/pod"
+var (
+	filename = flag.String("f", "", "Name of the file to parse")
 )
+func check(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+type taskStruct struct {
```
wlynch: Instead of using a new struct, could we just use the already defined task struct? We should be able to reference the fields directly like we do with this type!

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
+		check(tmpl.Execute(os.Stdout, InputParams))
+	}
+
+	if task.Spec.Inputs.Resources != nil{
+		print(" ***Resources :-***\n")
+		for _, i := range task.Spec.Inputs.Resources {
+			InputParams := taskStruct{Name: i.Name, Type1: i.Type}
+			tmpl := template.Must(template.New("Details").Parse("- Name : {{.Name}} \n- Type : {{.Type1}}\n"))
+			check(tmpl.Execute(os.Stdout, InputParams))
+			}
```
wlynch: Make sure to use gofmt! https://blog.golang.org/go-fmt-your-code

There should be a plugin for whatever editor you use. (if you have trouble finding one just let us know and we can help find one)

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
+	}
+
+	if task.Spec.Outputs != nil {
+		println("### Output:-")
+		for _, i := range task.Spec.Outputs.Resources {
+			InputParams := taskStruct{Name: i.Name, Type1: i.Type}
+			tmpl := template.Must(template.New("Details").Parse("- Name : {{.Name}} \n- Type : {{.Type1}}\n" ))
```
wlynch: This looks exactly the same as how we handle input resources, which is generally a sign that we should be consolidating code into its own func that both can use.

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
+
+	var task v1alpha1.Task
+	if _, _, err := scheme.Codecs.UniversalDeserializer().Decode(dat, nil, &task); err != nil {
+		log.Fatal(err)
+	}
+
 	// TODO: use Go templates to print v1alpha1.Task as Markdown
-	// TODO: tests!
+	fmt.Println("#",task.Name)
+	fmt.Printf("## Install the Task \n kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/%s/%s.yaml\n ### Inputs \n ***Parameters :-***\n",task.Name,task.Name)
```
wlynch: This could also be templatized if you want! :D

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
+
+	for _, i := range task.Spec.Inputs.Params {
+		InputParams := taskStruct{Name: i.Name, Description: i.Description, Default: i.Default, Type: i.Type}
+		tmpl := template.Must(template.New("Details").Parse("- Name : {{.Name}}, {{.Description}}, ({{.Default}})\n" ))
```
wlynch: You could potentially merge all of these into a single template if you wanted to!

I think it's fine to leave as is for now, but something to consider for the next PR! ^_^

tekdoc/main.go
```diff
@@ -17,15 +17,72 @@ limitations under the License.
...
+		check(tmpl.Execute(os.Stdout, InputParams))
+	}
+
+	if task.Spec.Inputs.Resources != nil{
+		print(" ***Resources :-***\n")
```
wlynch: I have no idea how/why this works. 🙃 

It's probably safer to stick with fmt.Print (and fmt.Println below)

wlynch: TIL: https://golang.org/ref/spec#Bootstrapping

chandanikumari: okay sure!

tekdoc/main.go
chandanikumari: I'll work on it and get back

tekdoc/main.go
chandanikumari: Added the plugin, thanks!

tekdoc/main.go
chandanikumari: Noted!

tekdoc/main.go
chandanikumari: This is the next step on which I've started working on :D

tekdoc/main.go
chandanikumari: > Instead of using a new struct, could we just use the already defined task struct? We should be able to reference the fields directly like we do with this type!

@wlynch Could you please help me finding the existing struct which I can reuse in my code?

chandanikumari:
@wlynch please review

tekdoc/main.go
```diff
@@ -17,15 +17,67 @@ limitations under the License.
...
-	// TODO: tests!
+
+	x := a{
+		Taskname:  task.Name,
+		Input:     task.Spec.Inputs.Params,
+		Resources: task.Spec.Inputs.Resources,
+		Output:    task.Spec.Outputs.Resources,
+	}
+
+	t := `# {{.Taskname}}
```
wlynch: This looks great! :D

tekdoc/main.go
```diff
@@ -17,15 +17,67 @@ limitations under the License.
...
+{{range .Resources}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}
+{{end}}
+### Output:-
+{{range .Output}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}
+{{end}}`
+
+	tmpl := template.Must(template.New("test").Parse(t))
+	err = tmpl.Execute(os.Stdout, x)
+	if err != nil {
+		panic(err)
```
wlynch: Let's use log.Fatal to exit gracefully.

tekdoc/main.go
```diff
@@ -17,15 +17,67 @@ limitations under the License.
...
+{{end}}
+### Resources:-
+{{range .Resources}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}
+{{end}}
+### Output:-
+{{range .Output}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}
+{{end}}`
+
+	tmpl := template.Must(template.New("test").Parse(t))
+	err = tmpl.Execute(os.Stdout, x)
```
wlynch: This is fine, but a good habit to get into is to prefer using:
```go
if err := ...; err != nil {
 ...
}
```

whereever possible. This helps avoid problems of not checking errors / overwriting values.

tekdoc/main.go
```diff
@@ -17,15 +17,67 @@ limitations under the License.
...
+	var task v1alpha1.Task
+
+	if _, _, err := scheme.Codecs.UniversalDeserializer().Decode(dat, nil, &task); err != nil {
+		log.Fatal(err)
+	}
+
 	// TODO: use Go templates to print v1alpha1.Task as Markdown
-	// TODO: tests!
+
+	x := a{
```
wlynch: Let's add a TODO comment to use the task struct directly.

tekdoc/main.go
```diff
@@ -17,15 +17,67 @@ limitations under the License.
...
+}
+
 func main() {
-	fmt.Println("Hello this is STILL TekDoc")
 
 	// TODO: take a file path, parse it as YAML -> v1alpha1.Task
+	flag.Parse()
+	dat, err := ioutil.ReadFile(*filename)
+	if err != nil {
+		log.Fatal(err)
```
wlynch: Let's add some context in the error logs throughout this file. e.g.

```go
log.Fatalf("error reading file: %v", err)
```

```go
log.Fatalf("error decoding Task: %v", err)
```

etc.

tekdoc/main.go
chandanikumari: Noted!

chandanikumari:
@wlynch Please review

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
-	// TODO: tests!
+	x := a{
+		Taskname: task.Name,
+	}
+	tmpl := template.Must(template.New("test").Parse(`# {{.Taskname}}
+## Install the Task
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/{{.Taskname}}/{{.Taskname}}.yaml
+### Input:-
+`))
+	if err = tmpl.Execute(os.Stdout, x); err != nil {
```
wlynch: if err := tmpl.Execute(os.Stdout, task); err != nil {

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
+		log.Fatalf("error executing the template: %v", err)
+	}
+	if task.Spec.Inputs.Params != nil {
+		x = a{
+			Input: task.Spec.Inputs.Params,
+		}
+		t := `{{range .Input}}- {{.Name}}, {{.Description}}
+{{end}}`
+		tmpl = template.Must(template.New("test").Parse(t))
+		if err = tmpl.Execute(os.Stdout, x); err != nil {
```
wlynch: if err := tmpl.Execute(os.Stdout, task.Spec.Inputs); err != nil {

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
+	}
+	tmpl := template.Must(template.New("test").Parse(`# {{.Taskname}}
+## Install the Task
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/{{.Taskname}}/{{.Taskname}}.yaml
+### Input:-
+`))
+	if err = tmpl.Execute(os.Stdout, x); err != nil {
+		log.Fatalf("error executing the template: %v", err)
+	}
+	if task.Spec.Inputs.Params != nil {
```
wlynch: This still runs the risk of panicking if task.Spec or task.Spec.Inputs are nil.

I think we want to have a check and then stop execution if task.Spec is nil (since then there would be nothing to do), and then check if task.Spec.Inputs is nil here.

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
+		}
+		t := `{{range .Input}}- {{.Name}}, {{.Description}}
+{{end}}`
+		tmpl = template.Must(template.New("test").Parse(t))
+		if err = tmpl.Execute(os.Stdout, x); err != nil {
+			log.Fatalf("error executing the template: %v", err)
+		}
+	}
+
+	if task.Spec.Inputs.Resources != nil{
```
wlynch: Same nil panic concern applies here as well.

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
+		x = a{
+			Resources: task.Spec.Inputs.Resources,
+		}
+		t := `### Resources:-
+{{range .Resources}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}	
+{{end}}`
+		tmpl = template.Must(template.New("test").Parse(t))
+		if err = tmpl.Execute(os.Stdout, x); err != nil {
```
wlynch: if err := tmpl.Execute(os.Stdout, task.Spec.Inputs); err != nil {

tekdoc/main.go
```diff
@@ -17,15 +17,91 @@ limitations under the License.
...
+
+	if task.Spec.Outputs != nil{
+		x = a{
+			Output: task.Spec.Outputs.Resources,
+		}
+		t := `### Output:-
+{{range .Output}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}	
+{{end}}`
+		tmpl = template.Must(template.New("test").Parse(t))
+		err = tmpl.Execute(os.Stdout, x)
```
wlynch: if err := tmpl.Execute(os.Stdout, task.Spec.Outputs); err != nil {

chandanikumari:
@wlynch please review

tekdoc/main.go
```diff
@@ -17,15 +17,82 @@ limitations under the License.
...
+		log.Fatal(err)
+	}
+
 	// TODO: use Go templates to print v1alpha1.Task as Markdown
-	// TODO: tests!
+	if task.Spec.Inputs != nil {
+
+		tmpl := template.Must(template.New("test").Parse(`# {{.Name}}
+## Install the Task
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/{{.Name}}/{{.Name}}.yaml
```
wlynch: I do want to get this in, so this is fine for now, but we should consider breaking this out of the `Inputs` conditional in the future. i.e. It is possible (although I expect not common) for a task to have no inputs, but we would still want this to be printed out.

tekdoc/main.go
```diff
@@ -17,15 +17,82 @@ limitations under the License.
...
+{{range .}}- {{.ResourceDeclaration.Name}}, {{.ResourceDeclaration.Type}}	
+{{end}}`
+			tmpl := template.Must(template.New("test").Parse(t))
+			err := tmpl.Execute(os.Stdout, task.Spec.Outputs.Resources)
+			if err != nil {
+				log.Fatalf("error executing the template: %v", err)
+			}
+		}
+	} else {
+		fmt.Println("There is no task input")
```
wlynch: Also a minor nit (so not going to block submission on this): We should change this to log.Info or print to stderr instead of printing to stdout (which is the default for fmt.Println).

wlynch:
Oh! We should also squash these commits prior to submitting. (not sure if prow does this automatically). See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing

PR #427, opened 2020-01-10
make https for default protocol for webhook in OpenShift

akihikokuroda:
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/webhooks-extension/overlays/openshift-all/deployment-patch.yaml
+++ b/webhooks-extension/overlays/openshift-all/deployment-patch.yaml
@@ -11,11 +11,11 @@ spec:
       containers:
         - name: webhooks-extension
           env:
-          # If this endpoint's protocol is https, ssl verification will be enabled on the github webhook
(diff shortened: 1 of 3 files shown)


a-roberts:
I've verified this on OCP 4.2 with the following steps.

- Install from Aki's branch - only setting the callback URL correctly (so keeping the https default prefix for the callback url)
- Got a green webhook 😄  with

image

- Pushed to my repository, observed my buildah Pipeline starting OK

- Then as an experiment I enabled SSL verification, pushed to the repository, webhook goes red as I suspected

`We couldn’t deliver this payload: Peer certificate cannot be authenticated with given CA certificates`

So this PR is working as intended, we already call out the need to set SSL verification otherwise here https://github.com/tektoncd/experimental/blob/master/webhooks-extension/docs/WebhookSecurity.md

akihikokuroda:
@a-roberts Thanks for review.  I updated the comment in the yaml file.

PR #461, opened 2020-02-12
Oci tool: makes use of oci-artifacts to store and retrieve Tekton resources

sthaha:
# Changes

This PR relates to the [#1839: Versioning of Tasks](https://github.com/tektoncd/pipeline/issues/1839) discussion. The PR adds an `oci` tool that is currently able to store and retrieve Tekton resource (Pipeline, Task, ClusterTask...) from a registry. 

Credits: Thanks for Pierre Tasci @pierretasci for the initial implementation. 
# Submitter Checklist

review them:

- [🙅‍♂ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._


oci/pkg/oci/image.go
```diff
@@ -0,0 +1,165 @@
...
+func ValidateImageName(name string) (*ImageReference, error) {
+	if name == "" {
+		return nil, fmt.Errorf("%s is not a valid image name", name)
+	}
+
+	return getNameParts(name)
+}
+
+// getNameParts will attempt to return a { "hostname/image-name", "tag or sha" } list of the provided image name. There maybe less than the full 2 elements if one of the parts isn't included in the original name.
+func getNameParts(name string) (*ImageReference, error) {
```
ImJasonH: The go-containerregistry library has a pretty solid set of methods for this kind of name parsing, I'd love to avoid having to maintain it in two places.

https://godoc.org/github.com/google/go-containerregistry/pkg/name#ParseReference

oci/pkg/oci/bundle.go
```diff
@@ -0,0 +1,156 @@
...
+// metadata about the resource.
+type ParsedTektonResource struct {
+	Name     string
+	Kind     *schema.GroupVersionKind
+	Contents string
+	Object   runtime.Object
+}
+
+// ReadPaths will recursively search each file path for Tekton resources and return the parsed specs or an error.
+func ReadPaths(filePaths []string) ([]ParsedTektonResource, error) {
```
ImJasonH: It looks like a lot of these methods are only called from within this package, so they don't have to be exported.

oci/pkg/oci/bundle.go
```diff
@@ -0,0 +1,156 @@
...
+	contents, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Try to tease out the type of the file from the extension and load them into a slice (if there are multiple
+	// entities in a single file).
+	var entities []string
+	switch ext := path.Ext(filePath); true {
+	case ext == ".yaml" || ext == ".yml":
```
ImJasonH: I'm pretty sure there's a standard K8s package to universally decode K8s config files. Can we use that instead?

oci/pkg/oci/image.go
```diff
@@ -0,0 +1,165 @@
...
+	orascontent "github.com/deislabs/oras/pkg/content"
+	orascontext "github.com/deislabs/oras/pkg/context"
+	"github.com/deislabs/oras/pkg/oras"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func init() {
+	orascontext.GetLogger(context.Background()).Logger.SetLevel(logrus.ErrorLevel)
```
ImJasonH: If we expect/hope this code will end up in either tkn or Tekton cobtrollers we should try to standardize on the logging used in those codebases, I don't think they use logrus today.

oci/pkg/oci/image.go
```diff
@@ -0,0 +1,165 @@
...
+	fmt.Printf("Pushed %s@%s to remote registry\n", pushName, desc.Digest)
+	return nil
+}
+
+// PullImage will fetch the image and return the Tekton resource specified by the kind and name.
+func PullImage(ref ImageReference, kind string, name string) ([]byte, error) {
+	resolver := docker.NewResolver(docker.ResolverOptions{})
+	memoryStore := orascontent.NewMemoryStore()
+
+	_, _, err := oras.Pull(
```
ImJasonH: go-containerregistry also has pretty straightforward methods for pulling and pushing, which are already used in Tekton. Can we try to use those to make this easier to integrate down the line?

ImJasonH:
Cc @ jonjohnsonjr

oci/pkg/oci/bundle.go
ImJasonH: This is done in tekdoc: https://github.com/tektoncd/experimental/blob/master/tekdoc/main.go#L45

pierretasci: The UniversalDeserializer is being used here: https://github.com/tektoncd/experimental/pull/461/files/28fc8d52254c72a62a82d422db91ff4ecaf11b0a#diff-be720fcc306c3d06da4d0c6eac9ed71eR126 to decode the file contents and ensure they are parseable as Tekton crds.

The problem is it the deserializer, AFAIK, will only parse a single resource from a set of contents even if the file is a `---` separated yaml config or a json array. We want to allow parsing multiple resources from a single file. If we take a look at how kubectl accomplishes this in the `cli-runtime` it is a bit more roundabout than what is trying to be accomplished here: https://github.com/kubernetes/kubectl/blob/4e4ec7cb5beac1fe75fd4be05d77c65030c30088/pkg/cmd/apply/apply.go#L323

To that end, for the first iteration of a tool that is meant just to be a POC, it might be worthwhile to keep the logic simple and more bespoke while we figure out how Tekton will support bundling these images. Perhaps we decide only Catalog CRDs should be bundled and we don't need parsing input at all"

oci/pkg/oci/image.go
pierretasci: I believe this is actually just here to suppress the built-in logger of the `oras` package which spits out warnings for using a custom mime that isn't publicly registered anywhere. I don't think this PR specifies a logger and is, unfortunately, only importing logrus to set the log level.

oci/pkg/oci/bundle.go
pierretasci: I believe ReadPaths and ParsedTektonResource are the only exported members and they are referenced outside of the `oci` package inside of the `action` package. Perhaps there is a case to be made for re-organizing but this tends to follow a cobra cli layout as far as I can tell.

vdemeester:
> I think there is a case to be made here for exploring using go-containerregistry instead of oras if that is what Tekton already uses. Perhaps that can be done as a follow on exploration since this code as-is accomplishes the functionality intended.

I would definitely vote for that :angel: This is the experimental repository, it's meant to be experiments and to be simpler to get stuff in :angel:

ImJasonH:
/lgtm
/approve

+1 for getting it merged first as a POC, sorry if my comments made it seem like we needed to get those in ASAP. I can send some PRs to see what it looks like using go-containerregistry.

ImJasonH:
Re: supporting yamls with `---`s I think it’d be fine to just prohibit those for the time being, and require inputs to be one per file. But again, we can discuss in a follow up PR, after this is merged.

PR #468, opened 2020-02-18
Use go-containerregistry to push and pull

ImJasonH:
Using this library should make it easier to integrate with Tekton controller components when the time comes.

# Submitter Checklist

review them:

- [N] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [Y] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._


ImJasonH:
> Also, I tried to check this out and run it locally on a simple pipeline. It hung for a long time and didn't give much indication what it was doing. Might be worth looking into a way to print out what is going on.

Hmm, I'm not sure what would cause this to hang, since it's basically just a bunch of small HTTP requests and tar operations.

https://github.com/google/go-containerregistry/issues/670 tracks upload/download progress repoprting in go-containerregistry, but I'm not sure there's much movement yet.

PR #477, opened 2020-02-25
OCI: minor code cleanup

sthaha:
- No functional change; only a minor cleanup

review them:

- [🙅] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/oci/pkg/oci/bundle.go
+++ b/oci/pkg/oci/bundle.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path"
 	"reflect"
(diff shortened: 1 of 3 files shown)


oci/pkg/oci/bundle.go
```diff
@@ -83,8 +84,8 @@ func readPath(filePath string) ([]ParsedTektonResource, error) {
 		return nil, err
 	}
 
-	// Try to tease out the type of the file from the extension and load them into a slice (if there are multiple
-	// entities in a single file).
+	// Try to tease out the type of the file from the extension and load them
+	// into a slice (if there are multiple entities in a single file).
```
pierretasci: [nit] I don't believe 80 chars is enforced elsewhere is it?

sthaha: agree, it isn't enforced and I agree we shouldn't enforce personal prefs as well .. It was reading the code and the edit was just part of how I was understanding it. ... I guess the existing comment already consumed two lines :)

PR #515, opened 2020-04-15
Add icons in charts description

eddycharly:
This PR adds icons in helm charts descriptions.


eddycharly:
do we have a url for dedicated logos per component ?
would be nice to have the component logo instead.

vdemeester:
> do we have a url for dedicated logos per component ?
> would be nice to have the component logo instead.

We don't have yet :stuck_out_tongue_closed_eyes: well we do have a specific logo for tektoncd/triggers but not for the others.

PR #539, opened 2020-05-13
Connect to a local MySQL instance to store results

ImJasonH:
This includes commented out config to point to a remotely-hosted Google
Cloud SQL instance instead of running the MySQL instance inside the
cluster.

Also, fix bug in watcher that resulted in reconcile never being called.

# Submitter Checklist

review them:

- [y] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [y] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

for more details._


ImJasonH:
/test tekton-experimental-unit-tests
🙄

results/cmd/api/main.go
```diff
@@ -19,28 +19,63 @@ package main
...
+		log.Fatal("Must provide both DB_USER and DB_PASSWORD")
+	}
+
+	if *dbaddr == "" {
+		log.Fatal("Must provide --db_addr")
+	}
+
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = ":50051"
```
wlynch, line 56: I suspect this was chosen because it's some sort of known default? Are there docs to point to?

results/config/api.yaml
```diff
@@ -28,9 +28,50 @@ spec:
...
+            secretKeyRef:
+              name: mysql-db-creds
+              key: mysql-db-user
+        - name: DB_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: mysql-db-creds
+              key: mysql-db-password
 
+      # Uncomment to configure CloudSQL proxy.
```
wlynch, line 46: Is this something we intend on using in the future, or should we just delete this?

results/cmd/api/main.go
ImJasonH: Honestly, I don't know if it's a default port. It seems to be commonly used in gRPC examples, which is where I picked it up ([e.g.](https://grpc.io/docs/guides/auth/#go)). It only really has to match whatever's in `config/api.yaml`.

results/config/api.yaml
ImJasonH: I think it's nice to demonstrate that other non-local MySQL databases can be used. I might just move it wholesale into a separate install doc, in a future change.

results/cmd/api/main.go
wlynch: Let's add a comment to link to that file.

ImJasonH: Done.

ImJasonH:
Whooops, never merged this. Rebased on master to resolve conflicts.

results/README.md
```diff
@@ -46,15 +68,18 @@ $ go get -u github.com/golang/protobuf/protoc-gen-go
...
-$ apt-get install sqlite3
+kubectl delete -f config/
 ```
 
-5. Create results.db and taskrun table under the cmd/api folder
+And don't forget to delete the secret:
 
 ```
-$ sqlite3
-$ sqilte> .read results.sql
```
wlynch, line 59: Let's keep both local and remote instructions.

ImJasonH: With this change there won't be any `sqlite` at all, and there's only really the mode where you connect to a remote mysql database (either running in-cluster or in Cloud SQL)

PR #558, opened 2020-06-15
Create CLI tool based on Tekton Generators

YolandaDu1997:
# Changes

I am working on [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590).

The proposal is [tektoncd/comunity/#125](https://github.com/tektoncd/community/issues/125).

This project aims at generating Tekton spec from simplified configs, which will help users bootstrap pipelines in a configurable way. 

Some milestones here:
1. Create tool that given an input spec with steps, generates the resulting Tekton resources for the particular type.
	- Create binary that invokes another binary on the path based on the type.
	- Read input steps and generate resulting pipeline config that mounts GitHub workspace and configures steps accordingly.
2. Add support for writing output to disk.
3. Add support for tasks and pipelines
4. Add support for applying config to cluster.
5. Add support for deleting configs from cluster.
6. Add support for other types (e.g. GitLab?)

# Submitter Checklist

review them:

for more details._


generators/demo.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: workspace-prettier-pvc
```
wlynch, line 4: A good practice is to put resource used for tests in a `testdata` folder.

generators/parser.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func Parse(r io.Reader) (PVC, error) {
+	fmt.Println("Yaml Parser Demo:")
+
+	reader := yaml.NewYAMLToJSONDecoder(r)
+	res := PVC{}
+	err := reader.Decode(&res)
+	if err != nil {
+		fmt.Errorf("%v", err)
```
wlynch, line 51: You can remove this. This only formats an error, it doesn't actually print anything out: https://pkg.go.dev/fmt?tab=doc#Errorf

generators/parser.go
```diff
@@ -0,0 +1,58 @@
...
+
+type PVC struct {
+	KindGroup       `json:",inline"`
+	ApiVersionGroup `json:",inline"`
+	MetadataGroup   `json:",inline"`
+	SpecGroup       `json:",inline"`
+}
+
+func Parse(r io.Reader) (PVC, error) {
+	fmt.Println("Yaml Parser Demo:")
```
wlynch, line 45: Remove.

generators/parser.go
```diff
@@ -0,0 +1,58 @@
+package generators
```
wlynch, line 1: This should have a package comment. 

If you haven't already, set up lint tools for your editor, or run [golint](https://github.com/golang/lint) / [golangci-lint](https://golangci-lint.run/) manually.

generators/parser.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type PVC struct {
+	KindGroup       `json:",inline"`
+	ApiVersionGroup `json:",inline"`
+	MetadataGroup   `json:",inline"`
+	SpecGroup       `json:",inline"`
+}
+
+func Parse(r io.Reader) (PVC, error) {
```
wlynch, line 44: Exported functions should also have comments.

generators/parser.go
```diff
@@ -0,0 +1,58 @@
+package generators
+
+import (
+	"fmt"
+	"io"
+
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+type KindGroup struct {
```
wlynch, line 10: Feel free to unexport these structs for the timebeing - we'll likely replace these in the next iteration.

generators/parser.go
```diff
@@ -0,0 +1,58 @@
...
+
+	reader := yaml.NewYAMLToJSONDecoder(r)
+	res := PVC{}
+	err := reader.Decode(&res)
+	if err != nil {
+		fmt.Errorf("%v", err)
+		return res, err
+	}
+
+	fmt.Printf("%s\n", res)
```
wlynch, line 55: Remove (or move this to main.go)

generators/parser_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"testing"
+)
+
+func TestParse(t *testing.T) {
+	file, err := os.Open("demo.yaml")
```
wlynch, line 11: `os.File` is itself a `io.Reader`. You can simplify the following by doing:

```go
file, err := os.Open("demo.yaml")
if err != nil {
  t.Error(err)
}
ret, err := Parse(file)
...
```

generators/parser_test.go
```diff
@@ -0,0 +1,34 @@
...
+		t.Errorf("%v", err)
+	}
+
+	var buf bytes.Buffer
+	buf.Write(content)
+	ret, err := Parse(&buf)
+	if err != nil {
+		t.Errorf("Error from 'Parse': %v", err)
+	}
+	fmt.Println(ret.SpecGroup.Spec.AccessModes[0])
```
wlynch, line 33: This by itself isn't a good test, since it won't actually fail if the content isn't what we expect. I don't know off the top of my head what the behavior of this parse library is, but many json/yaml libraries will ignore extra / unknown fields so long as the content is well-formed. If this is the case, I could pass an unrelated yaml file to this method and the test would pass, even though the resulting struct may be empty.

Let's verify the actual struct content. https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc is a great library for comparing complex structs.

Also feel free to use a simpler struct if you want - we're likely going to be modifying the message entirely next.

generators/pkg/parser/parser_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestParse(t *testing.T) {
+	file, err := os.Open("../../testdata/products.yaml")
```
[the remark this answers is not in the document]
YolandaDu1997: Sure. Does this mean that I will commit after the approval of [tektoncd/community#125](https://github.com/tektoncd/community/issues/125)?

wlynch:
Something else I thought off- we should add a README.md with a short description of the project and how to use.

YolandaDu1997:
> Something else I thought off- we should add a README.md with a short description of the project and how to use.

Sure. The README.md file will just describe the current yaml parser or the whole project?

wlynch:
> > Something else I thought off- we should add a README.md with a short description of the project and how to use.
> 
> Sure. The README.md file will just describe the current yaml parser or the whole project?

Both! We can have a brief description of the broader goals what the project aims to do, but when talking about features we should probably focus on current capabilities.

generators/pkg/parser/parser_test.go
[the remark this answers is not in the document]
wlynch: You can still add it into this PR. This PR just likely won't be approved until #125 is resolved.

wlynch:
We should also update the commit / PR message to be more descriptive about what we're doing and why - primarily so others who aren't familiar with the project can understand what is going on at a high level.

https://github.com/tektoncd/community/blob/master/standards.md#commit-messages has some additional guidance here. https://youtu.be/46f6lPmmy9s?t=824 is also a good talk to watch (the whole thing is worth watching, but I linked to the relevant bit directly).

generators/pkg/parser/parser_test.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	got, err := Parse(file)
+	if err != nil {
+		t.Fatalf("Error from 'Parse': %v", err)
```
dibyom, line 18: very small nit: Error should not be capitalized. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

YolandaDu1997:
> We should also update the commit / PR message to be more descriptive about what we're doing and why - primarily so others who aren't familiar with the project can understand what is going on at a high level.
> 
> https://github.com/tektoncd/community/blob/master/standards.md#commit-messages has some additional guidance here. https://youtu.be/46f6lPmmy9s?t=824 is also a good talk to watch (the whole thing is worth watching, but I linked to the relevant bit directly).

Done!

generators/pkg/parser/parser_test.go
YolandaDu1997: Got it! Already fix it!

wlynch:
I think we should be good to go to squash the commits for final approval.

YolandaDu1997:
> I think we should be good to go to squash the commits for final approval.

Done!

bobcatfish:
I haven't reviewed but approving so the OWNERS file goes in

/approve

PR #559, opened 2020-06-16
Implement CREATE functon of taskrun CRUD Version 1

yuege01:
# Implement CreateTaskRun function of Tekton Results API

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/experimental/blob/master/CONTRIBUTING.md)
for more details._

PR #560, opened 2020-06-17
Get task from database v1

yuege01:
# Implementation of GetTaskRun() of Tekton Results API
GetTaskRun() function gets an existing taskrun in the database as requested by users and sends back result to them.
1. Currently, we use kubernetes UUID to query and get a result in the database, we will use server generated results_id soon.
2. GetTaskRun() only gets one result from database, if multiple results are found, we only return the first one.

# Submitter Checklist

review them:

for more details._


wlynch:
/hold

Putting a hold on this since this includes changes from #559, and I'd like to review these diffs separately.

yuege01:
> /hold
> 
> Putting a hold on this since this includes changes from #559, and I'd like to review these diffs separately.

Yes, since get and delete have to use some of fields created in create-task, so I inherited them from create-task.

yuege01:
@wlynch @dibyom

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
 	}
 	return taskrunFromClient, nil
 }
+
+// Get TaskRun from database and send it to watcher
+func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
+	rows, err := database.Query("SELECT taskrunlog FROM taskrun WHERE uid = " + req.GetUid())
```
wlynch, line 88:
```suggestion
	rows, err := database.Query("SELECT taskrunlog FROM taskrun WHERE uid = ?", req.GetUid())
```

As is, this is susceptible to SQL injection (e.g. if req.uid = "foo; DROP TABLE taskruns")

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	}
+	var taskrunblob []byte
+	taskrun := &pb.TaskRun{}
+	for rows.Next() {
+		rows.Scan(&taskrunblob)
+		if err = proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
+		}
+		metadata := taskrun.GetMetadata()
+		fmt.Println(metadata.GetUid() + metadata.GetName() + metadata.GetNamespace())
```
wlynch, line 100: Prefer using the `log` package over `fmt.Print`

wlynch: Or alternatively, do we think this is a useful log, or something we can omit?

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+// Get TaskRun from database and send it to watcher
+func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
+	rows, err := database.Query("SELECT taskrunlog FROM taskrun WHERE uid = " + req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to query on database: %v", err)
+	}
+	var taskrunblob []byte
+	taskrun := &pb.TaskRun{}
+	for rows.Next() {
```
wlynch, line 94: We should log a warning if this returns multiple rows for some reason.

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+		rows.Scan(&taskrunblob)
+		if err = proto.Unmarshal(taskrunblob, taskrun); err != nil {
```
wlynch, line 96:
```suggestion
		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
```

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
 }
+
+// Get TaskRun from database and send it to watcher
+func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
+	rows, err := database.Query("SELECT taskrunlog FROM taskrun WHERE uid = " + req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to query on database: %v", err)
+	}
+	var taskrunblob []byte
```
wlynch, line 92: Move this into the loop so that there's less of a concern of accidentally reusing the same buffer.

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	taskrun := &pb.TaskRun{}
+	for rows.Next() {
+		rows.Scan(&taskrunblob)
+		if err = proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
```
wlynch, line 97: We should probably return an error here as well

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	defer cancel()
+	objectmeta := &pb.ObjectMeta{Uid: "31415926", Name: "mytaskrun", Namespace: "default"}
+	r, err := srv.CreateTaskRun(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1beta1", Metadata: objectmeta}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	t.Logf("Created taskrun: %s", r.String())
+
+	// Contact the server and get inserted data into database.
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second)
```
wlynch, line 86: You should be able to reuse the same context as before.

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	defer cancel()
+
+	r, err = srv.GetTaskRun(ctx, &pb.GetTaskRunRequest{Uid: "123459"})
+	if err != nil {
+		t.Fatalf("could not get taskrun: %v", err)
+	}
+	t.Logf("Get taskrun: %s", r.String())
```
wlynch, line 93: This will only print to a console if a test fails, but if the test fails we'll never reach here. We should be able to remove this.

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	t.Logf("Created taskrun: %s", r.String())
+
+	// Contact the server and get inserted data into database.
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	r, err = srv.GetTaskRun(ctx, &pb.GetTaskRunRequest{Uid: "123459"})
```
wlynch, line 89: Where is the value "123459" coming from? Is this deterministic?

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+
+func TestGetTaskRun(t *testing.T) {
+	// Create a temporay database
+	testDB, err := setupTestDB("testdb", t)
+	if err != nil {
+		t.Fatalf("failed to create temp file for db: %v", err)
+	}
+
+	// connect to fake server and do testing
+	srv := &server{db: testDB}
```
wlynch, line 75: We may want to consider moving this into the setup func if this is going to be done for every test.

results/cmd/api/main_test.go
wlynch: Shouldn't we be using r.GetUid() instead?

results/cmd/api/main.go
```diff
@@ -81,3 +81,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
 	}
 	return taskrunFromClient, nil
 }
+
+// Get TaskRun from database and send it to watcher
+func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
+	rows, err := database.Query("SELECT taskrunlog FROM taskrun WHERE uid = " + req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to query on database: %v", err)
```
wlynch, line 90:
```suggestion
		log.Fatalf("failed to query on database: %w", err)
```

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
 	}
 	return db, nil
 }
+
+func TestGetTaskRun(t *testing.T) {
+	// Create a temporay database
+	testDB, err := setupTestDB("testdb", t)
+	if err != nil {
+		t.Fatalf("failed to create temp file for db: %v", err)
```
wlynch, line 71:
```suggestion
		t.Fatalf("failed to setup db: %v", err)
```

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,33 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	}
+
+	// connect to fake server and do testing
+	srv := &server{db: testDB}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	objectmeta := &pb.ObjectMeta{Uid: "31415926", Name: "mytaskrun", Namespace: "default"}
```
wlynch, line 78: This isn't being reused, so you can just define this in the TaskRun below.

I'd also recommend breaking up the struct to 1 field per line for ease of reading.

results/cmd/api/main_test.go
wlynch: Ah, I think this is the `uid INTEGER` field. We shouldn't key on this, because users have no way of knowing what this value is.

I'm also forgetting why we have both `uid` and `logid` can you add a comments to the schema definition to document why we have each field?

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go
yuege01: You mean we shouldn't use uid to do query in the test? for now, to keep we get a unique taskrun, I just temporarily used it.
I've added why we have logid and uid  in results.sql

results/cmd/api/main_test.go

results/cmd/api/main.go
yuege01: log.Fatalf has no directive as %w, only fmt.Errorf has.

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go
```diff
@@ -61,5 +63,35 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+				Name:      "mytaskrun",
+				Namespace: "default",
+			}}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	t.Logf("Created taskrun: %s", r.String())
+
+	// Test if we can find inserted taskrun
+	r, err = srv.GetTaskRun(ctx, &pb.GetTaskRunRequest{Uid: r.GetMetadata().GetUid()})
```
wlynch, line 93: We should also validate that we got the response we expected. https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff will be useful here!

results/cmd/api/results.sql
```diff
@@ -1,2 +1,4 @@
+-- logid serve as auto-generated key by sqltite to uniquely identify a taskrun log
+-- uid is the key from taskrun itself we now use to do query on taskrun logs, future we may use cluster-info as key
```
wlynch, line 2: Which field from the taskrun does this correspond to? Should this be a string instead?

results/cmd/api/main_test.go
yuege01: Equal can not detect TaskRun type, so I used taskrun.String() to convert each one to a string then do the comparison, does it make sense to you?

results/cmd/api/results.sql
yuege01: Yes.

results/cmd/api/main_test.go

results/cmd/api/results.sql
yuege01: done

results/cmd/api/main_test.go
```diff
@@ -54,12 +57,42 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
 	}
 	if _, err := statement.Exec(); err != nil {
 		t.Fatalf("failed to execute the taskrun table creation statement statement: %v", err)
 	}
-	return db, nil
+	return srv, nil
+}
+
+func TestGetTaskRun(t *testing.T) {
+	// Create a temporay database
```
wlynch, line 71:
```suggestion
	// Create a temporary database
```

results/cmd/api/main_test.go
wlynch: Reopening (this wasn't done). The `cmp` library should allow you to compare complex structs. In particular, the `Diff` method is probably what you want so that you have a concise output.

results/cmd/api/main_test.go

dibyom:
I think we can cancel the hold now?

results/cmd/api/main.go
```diff
@@ -63,21 +63,47 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
-		return nil, fmt.Errorf("failed to excute insertion a new taskrun: %v", err)
+		return nil, fmt.Errorf("failed to excute insertion a new taskrun: %w", err)
 
 	}
 	return taskrunFromClient, nil
 }
+
+// Get TaskRun from database and send it to watcher
+func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
```
dibyom, line 87: the database variable is only used once. why not change the next line to s.db.Query() and get rid of the database variable?

results/cmd/api/main_test.go
```diff
@@ -54,12 +57,42 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+				Namespace: "default",
+			}}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	t.Logf("Created taskrun: %s", r.String())
+
+	// Test if we can find inserted taskrun
+	res, err := srv.GetTaskRun(ctx, &pb.GetTaskRunRequest{Uid: r.GetMetadata().GetUid()})
+	if err != nil || !cmp.Equal(r.String(), res.String()) {
```
dibyom, line 95: I'd move the cmp and err comparison to separate lines so that it is easier to debug what went wrong if the test fails. Also, you should use `cmp.Diff` to compare complex objects like structs. That way, if the test fails, the output will contain the diff between what we wanted and we we got. See the example in https://godoc.org/github.com/google/go-cmp/cmp#Diff

results/cmd/api/main.go

results/cmd/api/results.sql
```diff
@@ -1,2 +1,9 @@
+<<<<<<< HEAD
```
wlynch, line 1: Looks like we need to resolve a diff here.

yuege01: Sure.

PR #561, opened 2020-06-18
Delete taskrun version 1

yuege01:
# Implement DeleteTaskRun()
Implementing a DeleteTaskRun() function to delete an existing taskrun according to kubernete UUID specified in user request
1. Right now we use k8s UUID to delete, will switch to API server generated ID soon.

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

results/cmd/api/main.go
```diff
@@ -81,3 +106,24 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	}
+	results, err := statement.Exec(req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to execute delete statement: %v", err)
+	}
+	affect, err := results.RowsAffected()
+	if err != nil {
+		log.Fatalf("failed to retrieve results: %v", err)
+	}
+	fmt.Println(affect)
```
dibyom, line 124: was this intentional?

results/cmd/api/main.go
```diff
@@ -81,3 +106,24 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	if affect == 0 {
+		return &(empty.Empty{}), errors.New("Row Not Found: did not find a row with specified condition in database")
```
dibyom, line 126: minor nit: row not found: instead of Row Not Found

results/cmd/api/main.go
```diff
@@ -39,7 +41,16 @@ func main() {
 	}
 	defer db.Close()
 	srv := &server{db: db}
+	database := srv.db
 
+	statement, err := database.Prepare("CREATE TABLE IF NOT EXISTS taskrun (logid binary(16) PRIMARY KEY, taskrunlog BLOB, uid INTEGER, name TEXT, namespace TEXT)")
```
dibyom, line 46: this was moved to the .sql file right? so, we do not need this here?

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	objectmeta := &pb.ObjectMeta{Uid: "123459", Name: "mytaskrun", Namespace: "default"}
+	r, err := srv.CreateTaskRun(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "1", Metadata: objectmeta}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	res, err := srv.DeleteTaskRun(ctx, &pb.DeleteTaskRunRequest{Uid: "123459"})
+	if err != nil {
+		t.Fatalf("could not delete taskrun: %v", err)
+	}
+	println(r.String())
```
dibyom, line 87: We already have a t.Logf below, so lets remove the println?

yuege01: Yeah, this is one of legacy problems, I am looking for if there are more. Done

results/cmd/api/main.go
yuege01: Done

results/cmd/api/main.go
yuege01: I removed this print. Done

results/cmd/api/main.go
yuege01: yeah, it is not shown in conflict diff so I missed it, I will also delete them in other functions. Done

results/cmd/api/main.go
```diff
@@ -81,3 +96,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	results, err := statement.Exec(req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to execute delete statement: %v", err)
+	}
+	affect, err := results.RowsAffected()
+	if err != nil {
+		log.Fatalf("failed to retrieve results: %v", err)
+	}
+	if affect == 0 {
+		return &(empty.Empty{}), errors.New("row not found: did not find a row with specified condition in database")
```
wlynch, line 115:
```suggestion
		return nil, status.Errorf(status.NotFound, "TaskRun not found")
```

Use https://pkg.go.dev/google.golang.org/grpc/status?tab=doc instead.

Also in general, I like to avoid database specific terms in user facing error messages, because the implementation of the API server is a detail users don't know/care about. e.g. we could reimplement this using a filesystem instead of a database, but the results API would still be the same!

results/cmd/api/main.go
```diff
@@ -81,3 +96,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	}
+	return &(empty.Empty{}), nil
```
wlynch, line 117:
```suggestion
	return nil, nil
```

results/cmd/api/main.go
```diff
@@ -81,3 +96,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	database := s.db
+	statement, err := database.Prepare("DELETE FROM taskrun WHERE uid = ?")
+	if err != nil {
+		log.Fatalf("failed to create delete statement: %v", err)
+	}
+	results, err := statement.Exec(req.GetUid())
+	if err != nil {
+		log.Fatalf("failed to execute delete statement: %v", err)
+	}
+	affect, err := results.RowsAffected()
```
wlynch, line 110: Same as Get, I'm not sure if the `uid` field is what we want to use here. LogID, or even (namespace, task run name) may make more sense.

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	testDB, err := setupTestDB("testdb", t)
+	if err != nil {
+		t.Fatalf("failed to create temp file for db: %v", err)
+	}
+
+	// connect to fake server and do testing
+	srv := &server{db: testDB}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	objectmeta := &pb.ObjectMeta{Uid: "123459", Name: "mytaskrun", Namespace: "default"}
```
wlynch, line 78: I don't think this is used multiple times, so you can just inline this below.

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	r, err := srv.CreateTaskRun(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "1", Metadata: objectmeta}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	res, err := srv.DeleteTaskRun(ctx, &pb.DeleteTaskRunRequest{Uid: "123459"})
```
wlynch, line 83: Is this UID generated by the server? If so we should derive the parameter values from the returned task above so that the test will be less brittle to changes to the UID generation.

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	if err != nil {
+		t.Fatalf("could not delete taskrun: %v", err)
+	}
+	t.Logf("deleted taskrun: %s", res.String())
```
wlynch, line 87: Also mentioned this in the Get PR - I'm not sure if this provides significant value, and can be reomoved. Let me know if you think otherwise though!

results/cmd/api/main_test.go
```diff
@@ -63,3 +63,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+}
+
```
wlynch, line 89: How can we test that the call to Delete actually worked?

results/proto/api.proto
```diff
@@ -18,6 +20,10 @@ message CreateTaskRunRequest {
 	// TODO: Accept some unique cluster identifier?
 }
 
+message DeleteTaskRunRequest {
+	string uid = 1;
```
wlynch, line 24: From https://google.aip.dev/135

```
A name field must be included. It should be called name.
- The field should be annotated as required.
- The field should identify the resource type that it references.
```

results/cmd/api/main.go
```diff
@@ -57,6 +59,19 @@ type server struct {
...
+
+func (e *notFoundError) Error() string {
+	return e.text
+}
+
+// New returns an error that formats as the given text.
+func New(text string) error {
+	return &notFoundError{text}
+}
+
```
wlynch, line 74: This doesn't appear to be used. Remove.

results/cmd/api/main.go
```diff
@@ -81,3 +96,23 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
 	}
 	return taskrunFromClient, nil
 }
+
+func (s *server) DeleteTaskRun(ctx context.Context, req *pb.DeleteTaskRunRequest) (*empty.Empty, error) {
+	database := s.db
+	statement, err := database.Prepare("DELETE FROM taskrun WHERE uid = ?")
+	if err != nil {
+		log.Fatalf("failed to create delete statement: %v", err)
```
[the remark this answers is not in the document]
yuege01: done

results/cmd/api/main_test.go

results/cmd/api/main_test.go
yuege01: done, I used r.GetMetadata().GetUid() instead

results/cmd/api/main_test.go
yuege01: deleted.

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: I am thinking about it as well. We can discuss about it next week

results/cmd/api/main_test.go
yuege01: Added a new test

results/cmd/api/main.go
yuege01: Deleted

results/proto/api.proto
yuege01: Sure, I was using uid as name, next time we can discuss which field we use to Get, Delete or Update.

results/cmd/api/main_test.go
```diff
@@ -61,5 +63,40 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+		ApiVersion: "1",
+		Metadata: &pb.ObjectMeta{
+			Uid:       "123459",
+			Name:      "mytaskrun",
+			Namespace: "default"}}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+
+	// Delete inserted taskrun
```
wlynch, line 89: We should also test deleting a deleted taskrun.

yuege01: done

results/cmd/api/main_test.go
```diff
@@ -99,3 +100,43 @@ func TestGetTaskRun(t *testing.T) {
...
+	rows, err := srv.db.Query("SELECT taskrunlog FROM taskrun WHERE uid = ?", r.GetMetadata().GetUid())
+	if err != nil {
+		t.Fatalf("failed to query on database: %v", err)
+	}
+	if rows.Next() {
+		t.Fatalf("failed to delete taskrun: %v", r.String())
+	}
+
+	// Check if a deleted taskrun can be delete again
+	if _, err := srv.DeleteTaskRun(ctx, &pb.DeleteTaskRunRequest{Uid: r.GetMetadata().GetUid()}); err == nil {
```
wlynch, line 138:
```suggestion
	if _, err := srv.DeleteTaskRun(ctx, &pb.DeleteTaskRunRequest{Uid: r.GetMetadata().GetUid()}); status.Code(err) != code.NotFound {
```

results/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/google/go-cmp v0.4.0
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
 	github.com/tektoncd/pipeline v0.12.0
+	github.com/yuege01/experimental v0.7.0 // indirect
```
wlynch, line 10: This shouldn't be needed, since everything should be relative to the current module. We should remove this.

yuege01: done

results/cmd/api/main_test.go

PR #562, opened 2020-06-19
Update task version 1

yuege01:
# Implementation of UpdateTaskRun()
This pull request is to implement a UpdateTaskRun() function to receive update request from users and update taskruns 
in the database.
1. Currently, we use kubernetes UUID as key to update, we will use server generated results_id soon.
2. Due to lack of FieldMaskUtil API, right now we are updating whole taskrun instead of selected fields, may add this future in the future once we have the API

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as cyou
review them:

for more details._

### DIFF
--- a/results/cmd/api/main.go
+++ b/results/cmd/api/main.go
@@ -59,8 +59,7 @@ type server struct {
 
 // CreateTaskRun receives TaskRun from Watcher and save it to local Sqlite Server.
 func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest) (*pb.TaskRun, error) {
-	database := s.db
-	statement, err := database.Prepare("INSERT INTO taskrun (taskrunlog, uid, name, namespace) VALUES (?, ?, ?, ?)")
+	statement, err := s.db.Prepare("INSERT INTO taskrun (taskrunlog, uid, name, namespace) VALUES (?, ?, ?, ?)")
 	if err != nil {
 		log.Printf("failed to insert a new taskrun: %v\n", err)
 		return nil, fmt.Errorf("failed to insert a new taskrun: %w", err)
@@ -105,4 +104,26 @@ func (s *server) GetTaskRun(ctx context.Context, req *pb.GetTaskRunRequest) (*pb
 		}
 	}
 	return taskrun, nil
+}
+
+// UpdateTaskRun receives TaskRun and FieldMask from client and uses them to update records in local Sqlite Server.
+func (s *server) UpdateTaskRun(ctx context.Context, req *pb.UpdateTaskRunRequest) (*pb.TaskRun, error) {
+	// Update the entire row in database based on uid of taskrun.
+	statement, err := s.db.Prepare("UPDATE taskrun SET name = ?, namespace = ?, taskrunlog = ? WHERE uid = ?")
+	if err != nil {
+		log.Printf("failed to update a existing taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to update a exsiting taskrun: %w", err)
+	}
+	taskrunFromClient := req.GetTaskRun()
+	blobData, err := proto.Marshal(taskrunFromClient)
+	if err != nil {
+		log.Println("taskrun marshaling error: ", err)
+		return nil, fmt.Errorf("taskrun marshaling error: %w", err)
+	}
+	taskrunMeta := taskrunFromClient.GetMetadata()
+	if _, err := statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, taskrunMeta.GetUid()); err != nil {
+		log.Printf("failed to execute update of a new taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to execute update of a new taskrun: %w", err)
+	}
+	return taskrunFromClient, nil
 }
--- a/results/cmd/api/main_test.go
+++ b/results/cmd/api/main_test.go
@@ -3,12 +3,13 @@ package main
 import (
 	"context"
 	"database/sql"
-	"github.com/google/go-cmp/cmp"
 	"io/ioutil"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/golang/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
 	pb "github.com/tektoncd/experimental/results/proto/proto"
 )
 
@@ -97,5 +98,45 @@ func TestGetTaskRun(t *testing.T) {
 	}
 	if diff := cmp.Diff(r.String(), res.String()); diff != "" {
 		t.Fatalf("could not get the same taskrun: %v", diff)
+	}
+}
+func TestUpdateTaskRun(t *testing.T) {
+	// Create a temporary database
+	srv, err := setupTestDB("testdb", t)
+	if err != nil {
+		t.Fatalf("failed to create temp file for db: %v", err)
+	}
+
+	// Connect to fake server and create a taskrun
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	objectmeta := &pb.ObjectMeta{Uid: "123459", Name: "mytaskrun", Namespace: "default"}
+	r, err := srv.CreateTaskRun(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "1", Metadata: objectmeta}})
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+
+	// Update the created taskrun
+	objectmeta = &pb.ObjectMeta{Uid: "123459", Name: "ziweifan", Namespace: "tomorrow"}
+	r, err = srv.UpdateTaskRun(ctx, &pb.UpdateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1alpha1", Metadata: objectmeta}})
+	if err != nil {
+		t.Fatalf("could not update taskrun: %v", err)
+	}
+
+	// Validate by checking if we can get the updated taskrun
+	rows, err := srv.db.Query("SELECT taskrunlog FROM taskrun WHERE uid = ?", r.GetMetadata().GetUid())
+	if err != nil {
+		t.Fatalf("failed to query on database: %v", err)
+	}
+	for rows.Next() {
+		var taskrunblob []byte
+		taskrun := &pb.TaskRun{}
+		rows.Scan(&taskrunblob)
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			t.Fatal("unmarshaling error: ", err)
+		}
+		if diff := cmp.Diff(taskrun.String(), r.String()); diff != "" {
+			t.Fatalf("Update Function not properly implemented: %v", diff)
+		}
 	}
 }
--- a/results/cmd/watcher/main.go
+++ b/results/cmd/watcher/main.go
@@ -26,10 +26,12 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	taskruninformer "github.com/tektoncd/pipeline/pkg/client/injection/informers/pipeline/v1beta1/taskrun"
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
+
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/cache"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
@@ -54,11 +56,9 @@ func main() {
 	}
 	defer conn.Close()
 	client := pb.NewResultsClient(conn)
-
 	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), "watcher", func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
 		logger := logging.FromContext(ctx)
 		taskRunInformer := taskruninformer.Get(ctx)
-
 		c := &reconciler{
 			logger:        logger,
 			taskRunLister: taskRunInformer.Lister(),
@@ -70,7 +70,6 @@ func main() {
 			AddFunc:    impl.EnqueueControllerOf,
 			UpdateFunc: controller.PassNew(impl.EnqueueControllerOf),
 		})
-
 		return impl
 	})
 }
@@ -83,7 +82,6 @@ type reconciler struct {
 
 func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 	r.logger.Infof("reconciling resource key: %s", key)
-
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		r.logger.Errorf("invalid resource key: %s", key)
@@ -101,7 +99,7 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 		return err
 	}
 
-	r.logger.Infof("Sending update for %s/%s (uid %s)", namespace, name, tr.UID)
+	r.logger.Infof("Sending update for %s/%s (uid %s)", namespace, tr.Name, tr.UID)
 
 	// Send the new status of the TaskRun to the API server.
 	p, err := convert.ToProto(tr)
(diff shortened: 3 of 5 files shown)


yuege01:
@wlynch @dibyom

results/cmd/api/main.go
```diff
@@ -81,3 +81,27 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+		return nil, fmt.Errorf("failed to update a exsiting taskrun: %v", err)
+	}
+	taskrunFromClient := req.GetTaskRun()
+	blobData, err := proto.Marshal(taskrunFromClient)
+	if err != nil {
+		log.Println("taskrun marshaling error: ", err)
+		return nil, fmt.Errorf("taskrun marshaling error: %v", err)
+	}
+	taskrunMeta := taskrunFromClient.GetMetadata()
+	if _, err = statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, taskrunMeta.GetUid()); err != nil {
```
wlynch, line 102:
```suggestion
	if _, err := statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, taskrunMeta.GetUid()); err != nil {
```

results/cmd/api/main_test.go
```diff
@@ -63,3 +64,27 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
 	}
 	return db, nil
 }
+func TestUpdateTaskRun(t *testing.T) {
+	// Create a temporay database
```
wlynch, line 68:
```suggestion
	// Create a temporary database
```

results/demo/main.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	
+	
+			pods, err := clientset.CoreV1().Pods("").List(v1.ListOptions{})
+			if err != nil {
+				panic(err.Error())
+			}
+			for _, pod := range pods.Items {
+				fmt.Printf("%s %s\n", pod.GetName(), pod.GetCreationTimestamp())
+			}
```
wlynch, line 33: Run this through gofmt.

results/demo/main.go
```diff
@@ -0,0 +1,34 @@
...
+func main() {
+	kubeconfig := flag.String("kubeconfig", "/root/.kube/config", "absolute path to the kubeconfig file")
+	flag.Parse()
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		panic(err.Error())
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
```
wlynch, line 23: I don't think we need to panic here. We can just `log.Fatal` the error

results/demo/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func main() {
+	kubeconfig := flag.String("kubeconfig", "/root/.kube/config", "absolute path to the kubeconfig file")
```
wlynch, line 15: typically flags are defined in the package `var` section

wlynch, line 15: We shouldn't recommend the use of root. We should default to the user's home directory.

wlynch: e.g. `filepath.Join(os.Getenv("HOME"), ".kube", "config")`

results/demo/main.go
wlynch, line 33: This looks like this is just listing Pods on the cluster. Is this necessary? We may want to consider removing this file until we have something more meaningful to put here.

results/proto/api.proto
```diff
@@ -20,6 +21,7 @@ message CreateTaskRunRequest {
 
 message UpdateTaskRunRequest {
 	TaskRun task_run = 1;
+	google.protobuf.FieldMask update_mask = 2;
```
wlynch, line 24: Proto files are often used as the API definition itself. It doesn't look like we're doing anything with this field, so we should remove this for the time being.

results/cmd/watcher/main.go
```diff
@@ -54,25 +58,23 @@ func main() {
...
 		logger := logging.FromContext(ctx)
 		taskRunInformer := taskruninformer.Get(ctx)
-
 		c := &reconciler{
 			logger:        logger,
 			taskRunLister: taskRunInformer.Lister(),
 			client:        client,
 		}
-		impl := controller.NewImpl(c, c.logger, pipeline.PipelineRunControllerName)
+		impl := controller.NewImpl(c, c.logger, pipeline.TaskRunControllerName)
```
wlynch, line 69: These changes appear to be unrelated to adding support for `Update`. Let's break this out to it's own PR.

results/cmd/api/main.go
```diff
@@ -81,3 +81,27 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+
+// UpdateTaskRun receives TaskRun and FieldMask from client and uses them to update records in local Sqlite Server.
+func (s *server) UpdateTaskRun(ctx context.Context, req *pb.UpdateTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
+
+	// Update the entire row in database based on uid of taskrun.
+	statement, err := database.Prepare("UPDATE taskrun SET name = ?, namespace = ?, taskrunlog = ? WHERE uid = ?")
+	if err != nil {
+		log.Printf("failed to update a existing taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to update a exsiting taskrun: %v", err)
```
wlynch, line 93:
```suggestion
		return nil, fmt.Errorf("failed to update a exsiting taskrun: %w", err)
```

`%w` is generally preferred over `%v` when returning errors since it preserves underlying error information. See https://blog.golang.org/go1.13-errors for more details.

results/cmd/api/main.go
```diff
@@ -81,3 +81,27 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	}
+	taskrunFromClient := req.GetTaskRun()
+	blobData, err := proto.Marshal(taskrunFromClient)
+	if err != nil {
+		log.Println("taskrun marshaling error: ", err)
+		return nil, fmt.Errorf("taskrun marshaling error: %v", err)
```
wlynch, line 99:
```suggestion
		return nil, fmt.Errorf("taskrun marshaling error: %w", err)
```

results/cmd/api/main_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	pb "github.com/tektoncd/experimental/results/proto/proto"
+	fieldmask "google.golang.org/genproto/protobuf/field_mask"
```
wlynch, line 12: This library is unused, let's remove it. (Running `goimports` should fix this automatically!)

results/cmd/api/main.go
```diff
@@ -81,3 +81,27 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
 	}
 	return taskrunFromClient, nil
 }
+
+// UpdateTaskRun receives TaskRun and FieldMask from client and uses them to update records in local Sqlite Server.
+func (s *server) UpdateTaskRun(ctx context.Context, req *pb.UpdateTaskRunRequest) (*pb.TaskRun, error) {
+	database := s.db
```
wlynch, line 87: nit: It looks like this is only used once, so I'm not sure if this is strictly necessary (e.g. we could just use `s.db.Prepare` below instead). What do you think?

results/cmd/api/main.go
```diff
@@ -81,3 +81,27 @@ func (s *server) CreateTaskRun(ctx context.Context, req *pb.CreateTaskRunRequest
...
+	taskrunFromClient := req.GetTaskRun()
+	blobData, err := proto.Marshal(taskrunFromClient)
+	if err != nil {
+		log.Println("taskrun marshaling error: ", err)
+		return nil, fmt.Errorf("taskrun marshaling error: %v", err)
+	}
+	taskrunMeta := taskrunFromClient.GetMetadata()
+	if _, err = statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, taskrunMeta.GetUid()); err != nil {
+		log.Printf("failed to execute update of a new taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to execute update of a new taskrun: %v", err)
```
wlynch, line 104:
```suggestion
		return nil, fmt.Errorf("failed to execute update of a new taskrun: %w", err)
```

results/demo/main.go
yuege01: oh, later I found this one is useless I just used it for debugging watcher. Will delete it.

results/demo/main.go
yuege01: Will delete it.

results/demo/main.go
yuege01: will delete it.

results/cmd/api/main_test.go
yuege01: It is actually used in TestUpdatedTaskRun to create a fieldmask

results/cmd/api/main.go
yuege01: Done

results/cmd/api/main_test.go
yuege01: Done

results/demo/main.go
yuege01: Removed the file since it is unused now.

results/demo/main.go
yuege01: done

results/demo/main.go
yuege01: done

yuege01: deleted it. but good advice, thanks!

results/demo/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: Sure

results/cmd/api/main.go
yuege01: Thanks for the advice

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done, I have deleted this line

results/cmd/api/main.go

results/proto/api.proto
yuege01: Do you mean it is okay for now not to follow api definition in Google AIP guideline?

results/cmd/api/main_test.go
wlynch: Looking at the implementation of Update, it doesn't look like this is used. =\

This actually points out something important though! We should probably validate the result of the update somehow, either by using Get or reading the DB.

results/cmd/watcher/main.go
```diff
@@ -83,6 +85,7 @@ type reconciler struct {
 
 func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 	r.logger.Infof("reconciling resource key: %s", key)
+	print(ctx)
```
wlynch, line 88: Avoid using the builtin print. You should be able to use r.logger instead.

results/proto/api.proto
wlynch: Discussed offline- the ultimate goal is to support this, but for now it doesn't make much sense to have a field that isn't actually used.

results/cmd/api/main_test.go
yuege01: oh, you mean in the UpdateTaskRun function itself. I will delete it along with its definition in proto file.

results/demo/main.go
yuege01: done

results/cmd/api/main_test.go
yuege01: Done

results/cmd/watcher/main.go
yuege01: Just deleted it.

results/proto/api.proto
yuege01: Done

results/cmd/api/main_test.go
yuege01: I also added a further test to see if values really have changed.

results/cmd/api/main_test.go
```diff
@@ -61,5 +65,46 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+	// Validate by checking if we can get the updated taskrun
+	rows, err := srv.db.Query("SELECT taskrunlog FROM taskrun WHERE uid = ?", r.GetMetadata().GetUid())
+	if err != nil {
+		t.Fatalf("failed to query on database: %v", err)
+	}
+	for rows.Next() {
+		var taskrunblob []byte
+		taskrun := &pb.TaskRun{}
+		rows.Scan(&taskrunblob)
+		if err = proto.Unmarshal(taskrunblob, taskrun); err != nil {
```
wlynch, line 102:
```suggestion
		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
```

results/cmd/watcher/main.go
```diff
@@ -70,9 +72,9 @@ func main() {
 			AddFunc:    impl.EnqueueControllerOf,
 			UpdateFunc: controller.PassNew(impl.EnqueueControllerOf),
 		})
-
 		return impl
 	})
+	println()
```
wlynch, line 77: Remove

results/cmd/watcher/main.go
```diff
@@ -24,19 +24,23 @@ import (
...
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+
 	taskruninformer "github.com/tektoncd/pipeline/pkg/client/injection/informers/pipeline/v1beta1/taskrun"
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
+
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
wlynch, line 35: What is this used for? (as well as knative.dev/pkg/system/testing)

results/cmd/watcher/main.go
```diff
@@ -24,19 +24,23 @@ import (
 	"github.com/tektoncd/experimental/results/pkg/convert"
 	pb "github.com/tektoncd/experimental/results/proto/proto"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+
```
wlynch, line 27: nit: Don't separate these from the rest of the tekton packages.

results/cmd/api/main_test.go
```diff
@@ -61,5 +65,46 @@ func setupTestDB(dbName string, t *testing.T) (*sql.DB, error) {
...
+			t.Fatal("unmarshaling error: ", err)
+		}
+		metadata := taskrun.GetMetadata()
+		if metadata.GetName() != r.GetMetadata().GetName() || metadata.GetNamespace() != r.GetMetadata().GetNamespace() {
+			t.Fatal("Update Function not properly implemented.")
+		}
```
wlynch, line 108: Use https://pkg.go.dev/github.com/google/go-cmp/cmp?tab=doc#Diff instead to compare the entire struct.

results/cmd/watcher/main.go
yuege01: Without this import, when I ran the watcher.go, in gave me an error

results/cmd/api/main_test.go
yuege01: Done

results/cmd/watcher/main.go
yuege01: Done

results/cmd/watcher/main.go
yuege01: Done

results/cmd/api/main_test.go
yuege01: Done

results/cmd/watcher/main.go
yuege01: Done

wlynch:
Mostly LGTM! Just one thing I think we can do to simplify. Once the Get change is in, let's make the changes and squash, as well as update the commit/PR description.

results/go.sum
```diff
@@ -484,7 +492,20 @@ github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UV
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
 github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H4=
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
+<<<<<<< HEAD
```
wlynch, line 495: Fix diffs

results/cmd/watcher/main.go
```diff
@@ -26,17 +26,22 @@ import (
...
 	"k8s.io/client-go/tools/cache"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/injection/sharedmain"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
+
+	// this import ensures watcher connect to cluster
+	_ "knative.dev/pkg/system/testing"
```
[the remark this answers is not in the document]
yuege01: done

results/go.sum
yuege01: done

PR #564, opened 2020-06-22
Add CLI and Write support to Tekton Generators

YolandaDu1997:
# Changes

I am working on [Tekton Generators](https://github.com/tektoncd/pipeline/issues/2590).

This project aims at generating Tekton spec from simplified configs, which will help users bootstrap pipelines in a configurable way. 

I want to make some changes to this project:

1. To help users interact with generators, add the CLI command `show` to it.
2. Add support for writing output to disk. 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/generators/cmd/cli/cmd/root.go
+++ b/generators/cmd/cli/cmd/root.go
@@ -1,0 +1,22 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "tkn-gen",
+	Short: "This is the CLI for tekton generators",
+	Long: `tkn-gen will allow you to generate Tekton spec from the given simple spec. 
+			It also contains support for managing generated configuration.`,
+}
+
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
--- a/generators/cmd/cli/cmd/show.go
+++ b/generators/cmd/cli/cmd/show.go
@@ -1,0 +1,30 @@
+package cmd
+
+import (
+	"generators/pkg/writer"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var filename string
+var showCmd = &cobra.Command{
+	Use:   "show",
+	Short: "Show generated configuration.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(filename) == 0 {
+			cmd.Help()
+			return nil
+		}
+
+		if err := writer.WriteToDisk(filename, os.Stdout); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+func init() {
+	showCmd.Flags().StringVarP(&filename, "filename", "f", "", "spec file")
+	rootCmd.AddCommand(showCmd)
+}
--- a/generators/cmd/cli/cmd/write.go
+++ b/generators/cmd/cli/cmd/write.go
@@ -1,0 +1,35 @@
+package cmd
+
+import (
+	"errors"
+	"generators/pkg/writer"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var specFilename string
+var outputFilename string
+var writeCmd = &cobra.Command{
+	Use:   "write",
+	Short: "Write generated configuration to disk.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(specFilename) == 0 {
+			return errors.New("No input file specified")
+		}
+
+		file, err := os.Create(outputFilename)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+
+		return writer.WriteToDisk(specFilename, file)
+	},
+}
+
+func init() {
+	writeCmd.Flags().StringVarP(&specFilename, "filename", "f", "", "input spec file")
+	writeCmd.Flags().StringVarP(&outputFilename, "output", "o", "gen-task.yaml", "generated config file")
+	rootCmd.AddCommand(writeCmd)
+}
--- a/generators/pkg/generator/generate_test.go
+++ b/generators/pkg/generator/generate_test.go
@@ -31,12 +31,8 @@ func TestGenerateTask(t *testing.T) {
 				"--destination=gcr.io/wlynch-test/kaniko-test",
 				"--verbosity=debug"},
 		}}}}
-	got, err := GenerateTask(spec)
-	if err != nil {
-		t.Fatalf("error from 'GenerateTask': %v", err)
-	}
-
-	if diff := cmp.Diff(got, want); diff != "" {
+	got := GenerateTask(spec)
+	if diff := cmp.Diff(want, got); diff != "" {
 		t.Errorf("Tasks mismatch (-want +got):\n %s", diff)
 	}
 }
--- a/generators/pkg/parser/parser.go
+++ b/generators/pkg/parser/parser.go
@@ -3,24 +3,26 @@
 package parser
 
 import (
+	"fmt"
+	"generators/pkg/generator"
 	"io"
+	"io/ioutil"
 
-	"k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
 )
 
-type products struct {
-	Items []int    `json:"items,omitempty"`
-	Names []string `json:"names,omitempty"`
-}
-
 // Parse parses a yaml file from io.Reader and stores the
-// result in the products struct
-func Parse(r io.Reader) (products, error) {
-	reader := yaml.NewYAMLToJSONDecoder(r)
-	res := products{}
-	err := reader.Decode(&res)
+// result in the GitHubSpec struct
+func Parse(r io.Reader) (generator.GitHubSpec, error) {
+	spec, err := ioutil.ReadAll(r)
+
 	if err != nil {
-		return res, err
+		return generator.GitHubSpec{}, fmt.Errorf("fail to read from the input: %w", err)
 	}
+	res := generator.GitHubSpec{}
+	if err := yaml.Unmarshal(spec, &res); err != nil {
+		return res, fmt.Errorf("fail to unmarshal from the input: %w", err)
+	}
+
 	return res, nil
 }
--- a/generators/pkg/writer/write.go
+++ b/generators/pkg/writer/write.go
@@ -1,0 +1,38 @@
+// Package writer provides a method to write the generated task
+// to the yaml document.
+package writer
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"generators/pkg/generator"
+	"generators/pkg/parser"
+
+	"sigs.k8s.io/yaml"
+)
+
+// WriteToDisk writes the generated task
+// to the io.Writer
+func WriteToDisk(filename string, writer io.Writer) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return fmt.Errorf("unable to open the file from %s. check if the file exists: %w", filename, err)
+	}
+	defer file.Close()
+
+	spec, err := parser.Parse(file)
+	if err != nil {
+		return fmt.Errorf("unable to parse the file from %s: %w", filename, err)
+	}
+
+	task := generator.GenerateTask(spec)
+	data, err := yaml.Marshal(&task)
+	if err != nil {
+		return fmt.Errorf("unable to marshal the file from %s: %w", filename, err)
+	}
+
+	_, err = writer.Write(data)
+	return err
+}
--- a/generators/pkg/writer/write_test.go
+++ b/generators/pkg/writer/write_test.go
@@ -1,0 +1,26 @@
+package writer
+
+import (
+	"bytes"
+	"io/ioutil"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestWriteToDisk(t *testing.T) {
+	buf := new(bytes.Buffer)
+	if err := WriteToDisk("./testdata/spec.yaml", buf); err != nil {
+		t.Fatalf("error from 'WriteToDisk': %v", err)
+	}
+	got := buf.Bytes()
+
+	path := "./testdata/task.yaml"
+	want, err := ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatalf("fail to read file %s: %v", path, err)
+	}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("WriteToDisk mismatch (-want +got):\n %s", diff)
+	}
+}
--- a/generators/cmd/main.go
+++ b/generators/cmd/main.go
@@ -1,19 +1,7 @@
 package main
 
-import (
-	"fmt"
-	"log"
-	"os"
-
-	"generators/pkg/parser"
-)
+import "generators/cmd/cli/cmd"
 
 func main() {
-	fmt.Println("Yaml Parser Demo:")
-	products, err := parser.Parse(os.Stdin)
-	if err != nil {
-		log.Fatal(err)
-	}
-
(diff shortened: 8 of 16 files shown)


wlynch:
Looks like this is pulling in changes from #558 can you rebase so that those diffs aren't included?

Also,  I think for future PRs we can stop linking to the community issue. That was primarily to create the OWNERS file.

Thanks!

generators/cmd/cli/cmd/root.go
```diff
@@ -0,0 +1,22 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var RootCmd = &cobra.Command{
```
wlynch, line 10: We should leave things unexported as much as possible unless there's a struct need. I don't think there is one here, so we can go ahead and unexport this.

generators/cmd/cli/cmd/show.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var filename string
+var ShowCmd = &cobra.Command{
+	Use:   "show",
+	Short: "Show generated configuration.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(filename) == 0 {
```
wlynch, line 16: Currently this logic is not tested. Let's break this into a separate func that takes in a filename and a `io.Writer`- e.g. 

```go
func Write(filename string, w io.Writer) error {
...
}
```

This will also have the side effect of being able to be used by both `show` and `write` - one can pass in stdout as the writer while the other passes in another file.

generators/cmd/cli/cmd/show.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+	"generators/pkg/parser"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var filename string
+var ShowCmd = &cobra.Command{
```
wlynch, line 12: Also unexport this, for the same reason as rootCmd

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,20 @@
...
+	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"sigs.k8s.io/yaml"
+)
+
+// WriteToDisk writes the generated task
+// to the []byte
+func WriteToDisk(task v1beta1.Task) ([]byte, error) {
+	data, err := yaml.Marshal(&task)
+	if err != nil {
+		log.Fatal(err)
```
wlynch, line 17: You should avoid calling log.Fatal in library funcs, since `log.Fatal` will cause the program to immediately exit, which severely limits the ability for the func to be reused. In general you should return the error and let the caller decide how to handle it.

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,20 @@
+// Package writer provides a method to write the generated task
+// to the yaml document.
+package writer
+
+import (
+	"log"
+
+	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"sigs.k8s.io/yaml"
```
wlynch, line 9: Nice find! Perhaps we should standardize on this for reading as well. (if we do this, let's make the change in another PR).

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// WriteToDisk writes the generated task
+// to the []byte
+func WriteToDisk(task v1beta1.Task) ([]byte, error) {
```
wlynch, line 14: As is, this func is not adding additional benefit to callers.  When calling it, you would likely do something like:

```go
t := Task{...}
data, err := WriteToDisk(task)
if err != nil {
  return err
}
return data
```
which is near identical to the code within the func here. In most cases you would be able to just drop in replace it.

I suspect what we really want is to add more functionality to have it write to a file, which would make it different than just the yaml.Marshal. (this would also make the func name more fitting).

generators/pkg/writer/write_test.go
```diff
@@ -0,0 +1,52 @@
...
+			}}}}}
+
+	b, err := WriteToDisk(task)
+	if err != nil {
+		t.Fatalf("error from 'Writer': %v", err)
+	}
+
+	got := string(b)
+	want := `
+	    metadata:
```
wlynch, line 32: The reason why you might have been seeing indent errors, is because anything inside `` is treated as a string literal, which means that any indenting you put it to make it look indented w.r.t. the code will also be included in the resulting string.

We need to do one of 2 things:
1. Remove the extra whitespace from the want string in the code here.
2. Move this to its own yaml file in a testdata dir and read it in.
These are functionally equivalent so up to you which one you want to do.

Also I'm confused why we're not seeing presubmit failures for this. We should file an issue to make sure PR testing is set up.

generators/pkg/writer/write.go
YolandaDu1997: Do you mean the `parser`? Maybe change it to use this package as well?

generators/cmd/cli/cmd/show.go
YolandaDu1997: I am a little bit confused about this `Write` func and `WriteToDisk` func. For example if the command line is `tkn-gen show -f spec.yaml`, in the `Write`, I need to first parse the spec file then generate the Tekton Task and show the generated task as stdout. There is no need to call the `WriteToDisk`. But for the command `tkn-gen write -f spec.yaml`, in the `Write`, I need to call the `WriteToDisk` because it will write the output into a new yaml file. 

How could `Write` func be used by both `show` and `write`?

wlynch: Because both files and output streams (stdout, stderr) satisfy the [`io.Writer`](https://golang.org/pkg/io/#Writer) interface! :D 

Here's a concrete example: https://play.golang.org/p/8BWo0mWFmxE

generators/pkg/writer/write_test.go
YolandaDu1997: Done. I filed the issue #566 here.

generators/cmd/cli/cmd/show.go
YolandaDu1997: I change the `WriteToDisk` to:
```
func WriteToDisk(filename string, writer io.Writer) error {}
```
This will write the generated task from the input file to the io.Writer. It can be used by both `show` and `write` now. But only for one generated task. Next if I want to add support of the complete task and pipeline, it will change.

YolandaDu1997:
I have updated the `parser` and it now can parse the GithubSpec. Also, I have made changes to the cli command.

generators/cmd/cli/cmd/write.go
```diff
@@ -0,0 +1,43 @@
...
+var writeCmd = &cobra.Command{
+	Use:   "write",
+	Short: "Write generated configuration to disk.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(specFilename) == 0 {
+			cmd.Help()
+			return nil
+		}
+
+		if len(outputFilename) == 0 {
```
wlynch, line 21: You can make this a default value instead. e.g.

```go
writeCmd.Flags().StringVarP(&outputFilename, "output", "o", "gen-task.yaml", "generated config file")
```

generators/cmd/cli/cmd/write.go
```diff
@@ -0,0 +1,43 @@
...
+
+var specFilename string
+var outputFilename string
+var writeCmd = &cobra.Command{
+	Use:   "write",
+	Short: "Write generated configuration to disk.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(specFilename) == 0 {
+			cmd.Help()
+			return nil
```
wlynch, line 18: We probably want to return an error, since it will cause the program to return a non-zero exit code, which will signal to the user that there was an issue

IIRC, cobra will automatically print the help func if an error is returned, so we should be able to do something like:

```go
if len(specFilename) == 0 {
  return errors.New("no input file specified")
}
```

generators/cmd/cli/cmd/write.go
```diff
@@ -0,0 +1,43 @@
...
+			outputFilename = "./gen-task.yaml"
+		}
+
+		file, err := os.Create(outputFilename)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+
+		err = writer.WriteToDisk(specFilename, file)
```
wlynch, line 31: It's generally a good practice to use the inline syntax where possible:

```
if err := ... ; err != nil {
  ...
}
```

It helps limit scope of variables where possible to avoid potential issues when reusing var names.

wlynch: Alternatively, you can just `return writer.WriteToDisk`

generators/pkg/writer/output/gen-task.yaml
```diff
@@ -0,0 +1,16 @@
+metadata:
```
wlynch, line 1: Can we delete this file? (I don't think it's used anywhere, it's just an output artifact)

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return err
+	}
+
+	data, err := yaml.Marshal(&task)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(writer, "%s", data)
```
wlynch, line 39: It's probably cleaner to use the writer interface instead. I don't know if there will be any weird edge cases in string conversion using `%s`.

e.g.
```go
_, err := writer.Write(data)
return err
```

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"sigs.k8s.io/yaml"
+)
+
+// WriteToDisk writes the generated task
+// to the io.Writer
+func WriteToDisk(filename string, writer io.Writer) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
```
wlynch, line 21: Not sure what these errors look like in practice. Would it be clear to users what is failing, and what they would need to do to resolve?

If not, perhaps we should wrap these errors with some additional description of what was attempted.

YolandaDu1997: I use `fmt.Errorf()` to provide the description of errors.

generators/pkg/generator/generate_test.go
```diff
@@ -31,11 +31,7 @@ func TestGenerateTask(t *testing.T) {
 				"--destination=gcr.io/wlynch-test/kaniko-test",
 				"--verbosity=debug"},
 		}}}}
-	got, err := GenerateTask(spec)
-	if err != nil {
-		t.Fatalf("error from 'GenerateTask': %v", err)
-	}
-
+	got := GenerateTask(spec)
 	if diff := cmp.Diff(got, want); diff != "" {
```
dibyom, line 35: it should be `cmp.Diff(want,got)` for the printed diff to be of the form (-want/+got)

YolandaDu1997: sorry I' ll fix this

YolandaDu1997:
Done.

generators/pkg/writer/write.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// WriteToDisk writes the generated task
+// to the io.Writer
+func WriteToDisk(filename string, writer io.Writer) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return fmt.Errorf("unable to open the file from %s. check if the file exists: %w", filename, err)
+	}
```
wlynch, line 22:
```suggestion
	}
	defer file.Close()
```

generators/pkg/parser/parser.go
```diff
@@ -3,24 +3,27 @@
...
-	err := reader.Decode(&res)
+func Parse(r io.Reader) (generator.GitHubSpec, error) {
+	res := generator.GitHubSpec{}
+	spec, err := ioutil.ReadAll(r)
+	if err != nil {
+		return res, fmt.Errorf("fail to read from the input: %w", err)
+	}
+
+	err = yaml.Unmarshal(spec, &res)
 	if err != nil {
```
wlynch, line 24:
```suggestion
	res := generator.GitHubSpec{}
	if err := yaml.Unmarshal(spec, &res); err != nil {
```

generators/pkg/parser/parser.go
```diff
@@ -3,24 +3,27 @@
...
-}
-
 // Parse parses a yaml file from io.Reader and stores the
 // result in the products struct
-func Parse(r io.Reader) (products, error) {
-	reader := yaml.NewYAMLToJSONDecoder(r)
-	res := products{}
-	err := reader.Decode(&res)
+func Parse(r io.Reader) (generator.GitHubSpec, error) {
+	res := generator.GitHubSpec{}
```
wlynch, line 17: It's generally good practice to declare vars as close to where they are used. In this case, we're declaring res, but then doing nothing with in in the next statement. This would make more sense to be moved down to where the unmarshal occurs.

generators/pkg/writer/write_test.go
```diff
@@ -0,0 +1,36 @@
...
+	defer os.Remove(tempF.Name())
+
+	err = WriteToDisk("./testdata/spec.yaml", tempF)
+	if err != nil {
+		t.Fatalf("error from 'WriteToDisk': %v", err)
+	}
+
+	got, err := ioutil.ReadFile(tempF.Name())
+	if err != nil {
+		t.Fatal(err)
```
wlynch, line 25: Also annotate this and the other testing errors about what failed, so it's easier to see what failed from the go test logs.

generators/pkg/parser/parser.go
YolandaDu1997: I got it. But here because if `ioutil.ReadAll(r)` has an error, an empty struct and the error will be returned. I could change the return value to be `generator.GitHubSpec{}` instead of the var.

dibyom:
This is looking good to me! Just need to squash up the commits :)

generators/pkg/parser/parser.go
wlynch: Ah, I see! SGTM then :) (either way is fine)

generators/pkg/parser/parser.go
```diff
@@ -3,24 +3,26 @@
...
+// result in the GitHubSpec struct
+func Parse(r io.Reader) (generator.GitHubSpec, error) {
+	spec, err := ioutil.ReadAll(r)
+
 	if err != nil {
-		return res, err
+		return generator.GitHubSpec{}, fmt.Errorf("fail to read from the input: %w", err)
 	}
+	res := generator.GitHubSpec{}
+	if err = yaml.Unmarshal(spec, &res); err != nil {
```
wlynch, line 23:
```suggestion
	if err := yaml.Unmarshal(spec, &res); err != nil {
```

generators/cmd/cli/cmd/show.go
```diff
@@ -0,0 +1,30 @@
...
+var showCmd = &cobra.Command{
+	Use:   "show",
+	Short: "Show generated configuration.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(filename) == 0 {
+			cmd.Help()
+			return nil
+		}
+		err := writer.WriteToDisk(filename, os.Stdout)
+		if err != nil {
```
wlynch, line 20:
```suggestion
		if err := writer.WriteToDisk(filename, os.Stdout); err != nil {
```

wlynch:
Did not realize that the GitHub PR approve button would trigger the Prow approve 😅  It makes sense, but I don't remember it doing this before. Sorry about the noise.

generators/cmd/cli/cmd/show.go
YolandaDu1997: Didn't notice this, sorry:(

YolandaDu1997:
Done.

PR #575, opened 2020-06-29
Add support for tasks and pipelines

YolandaDu1997:
# Changes

This PR aims to add support for tasks and pipelines for [Tekton Generators](https://github.com/tektoncd/pipeline/issues/2590).

The changes includes:

1. Able to generate tasks needed from the complete github resource input.

2. Able to generate the pipeline from the given input.

# Submitter Checklist

review them:

for more details._


generators/pkg/generator/generate.go
```diff
@@ -3,20 +3,40 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              GitHubSpec `json:"spec"`
+}
+
 // GithubSpec defines Github spec
 type GitHubSpec struct {
 	URL   string         `json:"url,omitempty"`
 	Steps []v1beta1.Step `json:"steps,omitempty"`
 }
```
dibyom, line 27: we can put the two consts in a block:
```
const(
 GitCLoneURL = ...
 GitHubStatusURL = ...
)
```

generators/pkg/generator/generate.go
```diff
@@ -3,20 +3,40 @@
...
-// GenerateTask generates Tekton Task spec
-// from simplified GithubSpec configs.
-func GenerateTask(spec GitHubSpec) v1beta1.Task {
+const GitCloneURL = "https://raw.githubusercontent.com/tektoncd/catalog/master/git/git-clone.yaml"
+const GitHubStatusURL = "https://raw.githubusercontent.com/tektoncd/catalog/master/github/set_status.yaml"
+
+// GenerateTask generates Tekton Task
+// from simplified Github configs.
+func GenerateTask(github GitHub) v1beta1.Task {
 	var task v1beta1.Task
```
dibyom, line 34: slight simplification:
```
return v1beta1.Task {
  APIVersion: ...,
  Kind: "Task",
  Name:...,
 ....
}
```

generators/pkg/generator/generate.go
```diff
@@ -25,3 +45,94 @@ func GenerateTask(spec GitHubSpec) v1beta1.Task {
...
+func addTask(url string) (v1beta1.Task, error) {
+	task := v1beta1.Task{}
+	response, err := http.Get(url)
+	if err != nil {
+		return task, fmt.Errorf("fail to get git-clone task from %s: %w", url, err)
+	}
+	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return task, fmt.Errorf("fail to read reponse body: %v", err)
```
dibyom, line 58: If the function returned a *v1beta1.Task, all of the returns for the error cases can become `return nil, fmt.Errorf("....")`

generators/pkg/generator/generate_test.go
```diff
@@ -6,33 +6,285 @@ import (
...
... of a mounted volume.\n  if [[ -d \"$CHECKOUT_DIR\" ]] ; then\n    # Delete non-hidden files and directories\n    rm -rf \"$CHECKOUT_DIR\"/*\n    # Delete files and directories starting with . but excluding ..\n    rm -rf \"$CHECKOUT_DIR\"/.[!.]*\n    # Delete files and directories starting with .. plus any other character\n    rm -rf \"$CHECKOUT_DIR\"/..?*\n  fi\n}\n\nif [[ \"$(params.deleteExisting)\" == \"true\" ]] ; then\n  cleandir\nfi\n\ntest -z \"$(params.httpProxy)\" || export HTTP_PROXY=$(params.httpProxy)\ntest -z \"$(params.httpsProxy)\" || export HTTPS_PROXY=$(params.httpsProxy)\ntest -z \"$(params.noProxy)\" || export NO_PROXY=$(params.noProxy)\n\n/ko-app/git-init \\\n  -url \"$(params.url)\" \\\n  -revision \"$(params.revision)\" \\\n  -refspec \"$(params.refspec)\" \\\n  -path \"$CHECKOUT_DIR\" \\\n  -sslVerify=\"$(params.sslVerify)\" \\\n  -submodules=\"$(params.submodules)\" \\\n  -depth \"$(params.depth)\"\ncd \"$CHECKOUT_DIR\"\nRESULT_SHA=\"$(git rev-parse HEAD | tr -d '\\n')\"\nEXIT_CODE=\"$?\"\nif [ \"$EXIT_CODE\" != 0 ]\nthen\n  exit $EXIT_CODE\nfi\n# Make sure we don't add a trailing newline to the result!\necho -n \"$RESULT_SHA\" > $(results.commit.path)\n",
```
dibyom, line 194: We can make the Script field easier to read using raw Strings using `. See https://github.com/tektoncd/pipeline/blob/master/pkg/pod/pod_test.go#L485

generators/pkg/generator/generate.go
wlynch: I don't think we should fetch master, since this may result in unintended/inconsistent behavior. We should rely on the on cluster task for versioning for the time being.

generators/pkg/generator/generate.go
```diff
@@ -3,20 +3,40 @@
...
+
+// GenerateTask generates Tekton Task
+// from simplified Github configs.
+func GenerateTask(github GitHub) v1beta1.Task {
 	var task v1beta1.Task
-	task.Spec.Steps = spec.Steps
+	task.APIVersion = v1beta1.SchemeGroupVersion.String()
+	task.Kind = "Task"
+	task.Name = github.Name
+	task.Labels = map[string]string{"generator.tekton.dev": github.Name}
```
wlynch, line 38: We should append so we don't overwrite existing labels.

generators/pkg/generator/generate.go
```diff
@@ -25,3 +45,94 @@ func GenerateTask(spec GitHubSpec) v1beta1.Task {
...
+
+	ws := "source"
+	name := "build-from-git-repo"
+	tasksName := []string{"fetch-git-repo", "build-from-repo"}
+
+	pipeline := v1beta1.Pipeline{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: v1beta1.SchemeGroupVersion.String(),
+			Kind:       "Pipeline",
+		},
```
wlynch, line 83: @dibyom do you know if there's a way to generate this? This feels like the type of thing that k8s should know how to infer from the type.

generators/pkg/generator/generate.go
```diff
@@ -25,3 +45,94 @@ func GenerateTask(spec GitHubSpec) v1beta1.Task {
...
+	}
+
+	if err := yaml.Unmarshal(body, &task); err != nil {
+		return task, fmt.Errorf("unable to unmarshal git-clone task: %w", err)
+	}
+	task.Labels = map[string]string{"generator.tekton.dev": task.Name}
+	return task, nil
+}
+
+func GeneratePipeline(github GitHub) (v1beta1.Pipeline, []v1beta1.Task, error) {
```
wlynch, line 68: This func needs a comment. In particular, why are we returning both a Pipeline and a list of tasks?

generators/pkg/generator/generate.go
```diff
@@ -25,3 +45,94 @@ func GenerateTask(spec GitHubSpec) v1beta1.Task {
...
+							},
+						},
+					},
+					Workspaces: []v1beta1.WorkspacePipelineTaskBinding{
+						{
+							Name:      gitCloneTask.Spec.Workspaces[0].Name,
+							Workspace: ws,
+						},
+					},
+				},
```
wlynch, line 110: Isn't this the same as `gitCloneTask`? If not, why is this different, and why is this separated out? (we should make sure to add comments to document this!)

generators/pkg/generator/generate_test.go
```diff
@@ -6,33 +6,285 @@ import (
...
-	got := GenerateTask(spec)
+			},
+			}}}}
+	got := GenerateTask(github)
 	if diff := cmp.Diff(want, got); diff != "" {
 		t.Errorf("Tasks mismatch (-want +got):\n %s", diff)
 	}
 }
+
+func TestAddGitCloneTask(t *testing.T) {
```
wlynch, line 64: I don't think this is a useful test. This is prone to breaking if there is any upstream change to either of the tasks. I mentioned in another comment to not fetch remote configs and just rely on what's on the cluster. Once you do that this can be removed.

generators/pkg/writer/write.go
```diff
@@ -22,17 +22,39 @@ func WriteToDisk(filename string, writer io.Writer) error {
...
-	if err != nil {
-		return fmt.Errorf("unable to marshal the file from %s: %w", filename, err)
+	pipeline, tasks, err := generator.GeneratePipeline(github)
+
+	// write tasks
+	for _, task := range tasks {
+		_, err = writer.Write([]byte("---\n"))
+		if err != nil {
+			return err
+		}
```
wlynch, line 37:
```suggestion
		if _, err = writer.Write([]byte("---\n")); err != nil {
			return err
		}
```

Here and elsewhere in this file

generators/pkg/generator/generate.go
dibyom: Hmm..yeah I don't think we need to explicitly specify the Type meta fields (i.e. Kind and APIVersion). Though I'm not sure where that gets added automatically

generators/pkg/generator/generate.go
YolandaDu1997: Is it a good way to fix the commitId for the gitClone task? 
e.g.
```
https://raw.githubusercontent.com/tektoncd/catalog/2fa2e40357bfebbd739d475eab233972a8e16782/git/git-clone.yaml
```

generators/pkg/generator/generate.go
YolandaDu1997: I see. Then I need to change all my functions return a pointer instead of a value.

generators/pkg/generator/generate.go
YolandaDu1997: My idea is that when users want to `show` or `write` the generated config, the content will be the tasks used and a pipeline. This functions calls the `GenerateTask` and `addTask`. In this way I don't need to call these two functions again in the CLI to show the generated tasks. I am not sure if this strategy is good. 
The other one that I come up with is only return the pipeline here. But in the cli command, I need to call `GenerateTask` and `addTask` again.

generators/pkg/generator/generate.go
YolandaDu1997: This is the same as `gitCloneTask`. Here I refer the `gitCloneTask` and add the parameter value.

generators/pkg/generator/generate.go
```diff
@@ -3,25 +3,151 @@
...
+							Value: v1beta1.ArrayOrString{
+								Type:      v1beta1.ParamTypeString,
+								StringVal: u.Path,
+							},
+						},
+						{
+							Name: "STATE",
+							Value: v1beta1.ArrayOrString{
+								Type:      v1beta1.ParamTypeString,
+								StringVal: "pending",
```
wlynch, line 137: We probably don't want the final status to be pending- we want it to be the outcome of the pipeline. Pending would be useful as a first step to acknowledge that the pipeline is running.

generators/pkg/generator/generate_test.go
```diff
@@ -6,33 +6,169 @@ import (
...
+				},
+			},
+			Steps: []v1beta1.Step{
+				{
+					Container: corev1.Container{
+						Name:    "build",
+						Image:   "gcr.io/kaniko-project/executor:latest",
+						Command: []string{"/kaniko/executor"},
+						Args: []string{"--context=dir://$(workspaces.input.path)/src",
+							"--destination=gcr.io/wlynch-test/kaniko-test",
```
wlynch, line 36: Let's use fake values so people don't come bugging me personally about issues 🤣

generators/pkg/generator/generate_test.go
```diff
@@ -6,33 +6,169 @@ import (
...
+	got := GenerateTask(github)
 	if diff := cmp.Diff(want, got); diff != "" {
 		t.Errorf("Tasks mismatch (-want +got):\n %s", diff)
 	}
 }
+
+func TestGeneratePipeline(t *testing.T) {
+	ws := "source"
+	tasksName := []string{"fetch-git-repo", "build-from-repo", "final-set-status"}
+	want := &v1beta1.Pipeline{
```
wlynch, line 67: Is this supposed to match `testdata/pipeline.yaml`? As an alternative, you could read the result yaml from the testdata dir so you don't need to update multiple things for any test changes. (this is optional, feel free to push back or defer to another change if you want).

generators/pkg/writer/write.go
```diff
@@ -22,17 +21,34 @@ func WriteToDisk(filename string, writer io.Writer) error {
...
 	if err != nil {
 		return fmt.Errorf("unable to parse the file from %s: %w", filename, err)
 	}
 
-	task := generator.GenerateTask(spec)
+	task := generator.GenerateTask(github)
+	pipeline, err := generator.GeneratePipeline(github)
+
+	// write task
+	if _, err = writer.Write([]byte("---\n")); err != nil {
```
wlynch, line 33:
```suggestion
	if _, err := writer.Write([]byte("---\n")); err != nil {
```

here and instances below as well.

generators/pkg/generator/generate.go
YolandaDu1997: > If we are going with the cluster task approach, let's make sure we document somewhere that the user has to install a few tasks themselves.

Is it a good way to add instructions in the README file?

wlynch: Yup! That's probably the best spot for it.

wlynch:
Just needs the minor doc update + rebase!

YolandaDu1997:
Done.

generators/README.md
```diff
@@ -4,9 +4,22 @@ This project contains experimental code to create a tool for generating Tekton s
 
 See [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590) information and background.
 
+## Cluster Tasks
+The generated config would expect to use the cluster tasks adding to the pipeline. The cluster tasks include [`git-clone`](https://github.com/tektoncd/catalog/blob/master/git/git-clone.yaml) and [`github-set-status`](https://github.com/tektoncd/catalog/blob/master/github/set_status.yaml). The `git-clone` task is used to help clone a repo into the workspace. The `status-task` is used to help allow external services to mark GitHub commits with a state. 
+### Install the Task
+You can install the tasks with the specified revision(commit SHA) on the command line like this:
+```
+kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/<revision>/github/set_status.yaml
+kubectl apple -f https://raw.githubusercontent.com/tektoncd/catalog/<revision>/git/git-clone.yaml
```
wlynch, line 13: apply

generators/README.md
```diff
@@ -4,9 +4,22 @@ This project contains experimental code to create a tool for generating Tekton s
 
 See [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590) information and background.
 
+## Cluster Tasks
```
wlynch, line 7: We probably want to rephrase this, since there is actually something called a ClusterTask, and that's not a requirement for what we're doing here - https://pkg.go.dev/github.com/tektoncd/pipeline@v0.14.0/pkg/apis/pipeline/v1beta1?tab=doc#ClusterTask

YolandaDu1997:
Done!

dibyom:
Looks good 😃 
/approve

PR #587, opened 2020-07-10
Create Trigger for GitHub events

YolandaDu1997:
# Changes
This PR aims to create Tekton Trigger for the generated pipeline on GitHub events for [Tekton Generators](https://github.com/tektoncd/pipeline/issues/2590).

The changes includes:

1. Generate Trigger with the given pipeline.
2. Add support for writing the generated Trigger into disk. 
3. Add pull requests event github interceptor and update the status of commit.

# Submitter Checklist

review them:

for more details._


dibyom:
Haven't gone through this fully but at a high level this looks good to me!

generators/pkg/generator/generate.go
```diff
@@ -144,3 +155,163 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+				{
+					Name: p.Spec.Workspaces[0].Name,
+					VolumeClaimTemplate: &corev1.PersistentVolumeClaim{
+						Spec: corev1.PersistentVolumeClaimSpec{
+							AccessModes: []corev1.PersistentVolumeAccessMode{
+								corev1.ReadWriteOnce,
+							},
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									"storage": *resource.NewQuantity(1*1024*1024*1024, resource.BinarySI),
```
wlynch, line 207: Fine for now, but we should think about how we can make this configurable in the future.

generators/pkg/writer/write.go
```diff
@@ -45,10 +45,48 @@ func WriteToDisk(filename string, writer io.Writer) error {
...
+	data, err = yaml.Marshal(trigger.TriggerTemplate)
+	if err != nil {
+		return fmt.Errorf("unable to marshal the TriggerTemplate %s: %w", trigger.TriggerTemplate.Name, err)
+	}
+	if _, err := writer.Write(data); err != nil {
+		return err
+	}
+
+	// write EventListener
+	if _, err := writer.Write([]byte("---\n")); err != nil {
```
wlynch, line 82: The abundance of writer.Write + yaml.Marshal is starting to look repetitive. Would it make sense to combine this into a func where we just loop over each object?

generators/pkg/generator/generate.go
```diff
@@ -144,3 +155,163 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+					Name: "github-listener",
+					Interceptors: []*v1alpha1.EventInterceptor{
+						{
+							GitHub: &v1alpha1.GitHubInterceptor{
+								EventTypes: []string{"push"},
+								SecretRef: &v1alpha1.SecretRef{
+									SecretKey:  "secretToken",
+									SecretName: "github-secret",
+								},
+							},
```
wlynch, line 295: Similar to the disk space, we should also think of ways we can make this configurable.

Until then, we should make sure to document this behavior!

generators/pkg/generator/generate.go
```diff
@@ -144,3 +155,163 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+			Labels:    p.Labels,
+		},
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: v1alpha1.SchemeGroupVersion.String(),
+			Kind:       "EventListener",
+		},
+		Spec: v1alpha1.EventListenerSpec{
+			Triggers: []v1alpha1.EventListenerTrigger{
+				{
+					Name: "github-listener",
```
wlynch, line 286: nit: for everything but the event listener, we probably want to go with github-push. This will make it easier to add other event types (e.g. pull request).

generators/pkg/generator/generate.go
```diff
@@ -144,3 +155,163 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+				},
+			},
+		},
+	}
+
+	// create the eventlistener
+	el := &v1alpha1.EventListener{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: p.Namespace,
+			Name:      "github-listener",
```
wlynch, line 276: We should scope this to the input somehow. e.g. If I use the GitHub generator for 2 different repos, this is going to conflict.

generators/pkg/generator/generate.go
```diff
@@ -23,6 +27,13 @@ type GitHubSpec struct {
 	Steps []v1beta1.Step `json:"steps,omitempty"`
 }
 
+// Trigger includes the Tekton trigger resources
+type Trigger struct {
```
wlynch, line 31: Let's avoid exposing new types that may cause confusion with existing trigger types. I think it's fine to use this for grouping, but let's unexport this so it can't be used by other packages.

generators/pkg/writer/write.go
YolandaDu1997: I can break writer.Write + yaml.Marshal out to a separate function. For the objects that need to be written, e.g. Task, Pipeline, TriggerTemplate, ..., should I wrap them in an array like `[]interface{}` or in a struct to loop in the function?

YolandaDu1997:
Done.

generators/pkg/writer/write.go
wlynch: sgtm! other option is https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime?tab=doc#Object if you want access to type information.

YolandaDu1997:
Done.

generators/pkg/generator/generate_test.go
```diff
@@ -103,3 +105,85 @@ func TestGeneratePipeline(t *testing.T) {
...
+	path := "./testdata/triggerbinding.yaml"
+	file, err := ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatalf("fail to read file %s: %v", path, err)
+	}
+
+	tb := &v1alpha1.TriggerBinding{}
+	if err := yaml.Unmarshal(file, tb); err != nil {
+		t.Fatalf("fail to unmarshal from the input: %v", err)
+	}
```
wlynch, line 120: This section of code is also duplicative. We can break this out into a helper func to make this simpler. e.g.

```go
func unmarshal(t *testing.T, path string, i interface{}) {
	t.Helper()

	file, err := ioutil.ReadFile(path)
	if err != nil {
		t.Fatalf("fail to read file %s: %v", path, err)
	}

	if err := yaml.Unmarshal(file, i); err != nil {
		t.Fatalf("fail to unmarshal from the input: %v", err)
	}
}

func Test(t *testing.T) {
  tb := &v1alpha1.TriggerBinding{}
  unmarshal(t, "./testdata/triggerbinding.yaml", tb)
}
...

File: generators/pkg/writer/write.go [hunk 8]
```diff
@@ -26,28 +27,31 @@ func WriteToDisk(filename string, writer io.Writer) error {
...
-		return fmt.Errorf("unable to marshal the task %s: %w", task.Name, err)
+		return fmt.Errorf("unable to get the pipeline: %w", err)
 	}
-	if _, err := writer.Write(data); err != nil {
-		return err
+	trigger := generator.GenerateTrigger(pipeline)
+
+	// pack up all the objects
+	pac := []runtime.Object{task, pipeline, trigger.TriggerBinding, trigger.TriggerTemplate, trigger.EventListener}
+	for _, o := range pac {
```
wlynch, line 40: Should range over the objects in the write method instead?

This looks like it might be a good candidate for a [variadic function](https://golang.org/ref/spec#Passing_arguments_to_..._parameters). e.g. 

```go
func writeYaml(writer io.Writer, objs ...runtime.Object) error {
  for _, o := range objs {
    ...
}

File: generators/pkg/writer/write.go [hunk 9]
```diff
@@ -26,28 +27,31 @@ func WriteToDisk(filename string, writer io.Writer) error {
...
+		if err := wirteYaml(o, writer); err != nil {
+			return err
+		}
 	}
+	return nil
+}
 
-	// write pipeline
+func wirteYaml(o runtime.Object, writer io.Writer) error {
```
wlynch, line 48:
```suggestion
func writeYaml(o runtime.Object, writer io.Writer) error {
```

YolandaDu1997:
Done!

generators/README.md
```diff
@@ -4,8 +4,25 @@ This project contains experimental code to create a tool for generating Tekton s
 
 See [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590) information and background.
 
+## GitHub Token
```
wlynch, line 7: This is secret is actually the webhook secret - https://developer.github.com/webhooks/securing/

The secret described here would be the one for fetching source code / updating PRs.

PR #593, opened 2020-07-16
Add fieldmask utility to UpdateTaskRun to only update specific fields

yuege01:
Add fieldmask utility to UpdateTaskRun to only update specific fields in TaskRun

Added a FieldMask to UpdateTaskRunRequest proto message
Modified Api Server main.go to use fieldmask for updating
Added tests to test if specified fields are updated and unspecified fields are untouched

# Changes

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

results/cmd/api/main_test.go
```diff
@@ -114,14 +115,15 @@ func TestUpdateTaskRun(t *testing.T) {
...
+	r, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1beta1", Metadata: objectmeta}})
 	if err != nil {
 		t.Fatalf("could not create taskrun: %v", err)
 	}
 
 	// Update the created taskrun
 	objectmeta = &pb.ObjectMeta{Uid: "123459", Name: "ziweifan", Namespace: "tomorrow"}
-	r, err = srv.UpdateTaskRunResult(ctx, &pb.UpdateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1alpha1", Metadata: objectmeta}, ResultsId: r.GetResultsId()})
+	fmask, _ := fieldmask.New(&pb.TaskRun{}, "api_version", "metadata.uid")
+	r, err = srv.UpdateTaskRunResult(ctx, &pb.UpdateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1alpha1", Metadata: objectmeta}, ResultsId: r.GetResultsId(), FieldMask: fmask})
```
wlynch, line 126: hmmmm - I'm not sure what's going on here. 

The update mask includes uid and api_version, but we're modifying name. If this is passing, this means that name is still being modified even though it's not included in the mask.

We should verify this is working as intended.

results/cmd/api/main_test.go
```diff
@@ -114,14 +115,15 @@ func TestUpdateTaskRun(t *testing.T) {
...
-	r, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "1", Metadata: objectmeta}})
+	r, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1beta1", Metadata: objectmeta}})
 	if err != nil {
 		t.Fatalf("could not create taskrun: %v", err)
 	}
 
 	// Update the created taskrun
 	objectmeta = &pb.ObjectMeta{Uid: "123459", Name: "ziweifan", Namespace: "tomorrow"}
-	r, err = srv.UpdateTaskRunResult(ctx, &pb.UpdateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1alpha1", Metadata: objectmeta}, ResultsId: r.GetResultsId()})
+	fmask, _ := fieldmask.New(&pb.TaskRun{}, "api_version", "metadata.uid")
```
wlynch, line 125: There's a few different things we should test here:

1. No mask
2. Partial update with mask
   - Verify only fields that should be modified are, and that fields that were not included stay the same.
3. Mask with excess fields - e.g. setting a field in the mask but not setting a value in the request (this should unset the value).
4. Mask with empty fields.
5. Mask with invalid fields.
6. Interaction with repeated fields (see https://google.aip.dev/144#update-strategies)
6. Any more test cases you can think of.

I recommend representing this as a table driven test - https://github.com/golang/go/wiki/TableDrivenTests

results/proto/api.proto
```diff
@@ -32,6 +33,7 @@ message DeleteTaskRunRequest {
 message UpdateTaskRunRequest {
 	TaskRun task_run = 1;
 	string results_id = 2;
+	google.protobuf.FieldMask field_mask = 3;
```
wlynch, line 36: We should rename this to update_mask - See https://google.aip.dev/134

results/cmd/api/main.go
```diff
@@ -119,19 +120,49 @@ func (s *server) GetTaskRunResult(ctx context.Context, req *pb.GetTaskRunRequest
...
+			log.Fatal("unmarshaling error: ", err)
+			return nil, fmt.Errorf("failed to unmarshal taskrun: %w", err)
+		}
+	}
+	// Merge TaskRun from client into existing TaskRun based on fieldmask.
 	taskrunFromClient := req.GetTaskRun()
-	blobData, err := proto.Marshal(taskrunFromClient)
+	fieldMask := req.GetFieldMask()
+	msk, _ := mask.FromFieldMask(fieldMask, taskrun, false, false)
+	msk.Merge(taskrunFromClient, taskrun)
```
wlynch, line 153: Make sure to check the error for both of these calls.

results/cmd/api/main.go
```diff
@@ -119,19 +120,49 @@ func (s *server) GetTaskRunResult(ctx context.Context, req *pb.GetTaskRunRequest
...
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
+			return nil, fmt.Errorf("failed to unmarshal taskrun: %w", err)
+		}
+	}
+	// Merge TaskRun from client into existing TaskRun based on fieldmask.
 	taskrunFromClient := req.GetTaskRun()
-	blobData, err := proto.Marshal(taskrunFromClient)
+	fieldMask := req.GetFieldMask()
+	msk, _ := mask.FromFieldMask(fieldMask, taskrun, false, false)
```
wlynch, line 152:
```suggestion
	msk, _ := mask.FromFieldMask(fieldMask, taskrun, true, true)
```

isFieldNameJSON should be set to true since we want users to use the proto name format, not the Go struct JSON format.
isUpdateMask should be true since this is an update mask.

results/cmd/api/main.go
wlynch, line 153: It's hard to know from the variable names what's the purpose of each. I think `msk.Merge(taskrunFromClient, update)` might be clearer.

results/cmd/api/main_test.go
wlynch: Perhaps this is due to the FromFieldMask parameters?

results/cmd/api/main_test.go
```diff
@@ -138,7 +140,13 @@ func TestUpdateTaskRun(t *testing.T) {
...
 			t.Fatal("unmarshaling error: ", err)
 		}
-		if diff := cmp.Diff(taskrun.String(), r.GetTaskRun().String()); diff != "" {
+		if diff := cmp.Diff(taskrun.GetApiVersion(), r.GetTaskRun().GetApiVersion()); diff != "" {
+			t.Fatalf("Update Function not properly implemented: %v", diff)
+		}
+		if diff := cmp.Diff(taskrun.GetMetadata().GetUid(), r.GetTaskRun().GetMetadata().GetUid()); diff != "" {
+			t.Fatalf("Update Function not properly implemented: %v", diff)
+		}
+		if diff := cmp.Diff(taskrun.GetMetadata().GetName(), r.GetTaskRun().GetMetadata().GetName()); diff == "" {
```
wlynch, line 149:
```suggestion
		if diff := cmp.Diff(taskrun, r.GetTaskRun()); diff != "" {
```

yuege01: The point of this test is to check uid and apiversion are changed because they are specified in fieldmask, but name is not because it is not in fieldmask

results/cmd/api/main_test.go
yuege01: This is the same as what I intend to do in the last comment - updating only field in fieldmask

results/cmd/api/main.go
yuege01: The first should remain false, or it only accept UpperCase format of field, I tested it when I set it to true, it cannot find a field like api_version.

wlynch: I misread the godoc! You are correct. Only the isUpdateMask should be set to true then.

results/cmd/api/main_test.go
wlynch: ahhh. I assumed that the only thing that was being changed was the objectmeta. It wasn't clear to me that other things were being changed as well, because they were embedded in the Update call. We should break out any field updates to make it clearer which fields are being changed.

I suspect the table driven tests will also help make this clearer.

results/cmd/api/main.go
```diff
@@ -119,19 +120,49 @@ func (s *server) GetTaskRunResult(ctx context.Context, req *pb.GetTaskRunRequest
...
+			log.Println("Warning: multiple rows found")
+			break
+		}
+		rows.Scan(&taskrunblob)
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
+			return nil, fmt.Errorf("failed to unmarshal taskrun: %w", err)
+		}
+	}
+	// Merge TaskRun from client into existing TaskRun based on fieldmask.
```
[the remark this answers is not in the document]

results/cmd/api/main.go
```diff
@@ -119,19 +120,49 @@ func (s *server) GetTaskRunResult(ctx context.Context, req *pb.GetTaskRunRequest
...
+	msk, _ := mask.FromFieldMask(fieldMask, taskrun, false, false)
+	msk.Merge(taskrunFromClient, taskrun)
+	blobData, err := proto.Marshal(taskrun)
 	if err != nil {
 		log.Println("taskrun marshaling error: ", err)
 		return nil, fmt.Errorf("taskrun marshaling error: %w", err)
 	}
-	taskrunMeta := taskrunFromClient.GetMetadata()
+	// Update the entire row in database based on uid of taskrun.
+	statement, err := s.db.Prepare("UPDATE taskrun SET name = ?, namespace = ?, taskrunlog = ? WHERE results_id = ?")
```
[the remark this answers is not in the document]
yuege01: done

results/cmd/api/main_test.go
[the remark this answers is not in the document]

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main_test.go

results/cmd/api/main.go

dibyom:
Haven't gone through all of the tests yet, but the overall approach looks good to me. 
We should squash the commits before we merge though.

results/.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
wlynch, line 1: Looks like this file may have been checked in by accident. Let's remove?

It might also be helpful to add .vscode to your gitignore file!

results/cmd/api/main.go
```diff
@@ -161,3 +165,34 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
 	}
 	return nil, nil
 }
+
+// GetTaskRunByID is the helper function to get a TaskRun by results_id
+func GetTaskRunByID(s *server, id string) (*pb.TaskRun, error) {
```
wlynch, line 170:
```suggestion
func (s *server) getTaskRunByID(s *server, id string) (*pb.TaskRun, error) {
```

Instead of passing in the server as an arg, we can define the func on the struct itself.
I think we can also unexport this func, since this isn't something I would expect external users to call.

results/cmd/api/main.go
```diff
@@ -161,3 +165,34 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	resultsID, err := uuid.Parse(id)
+	if err != nil {
+		log.Fatal("failed to parse resultID string into resultsID UUID", err)
```
wlynch, line 173:
```suggestion
		log.Fatal("failed to parse resultID string into resultsID UUID: %v", err)
```

results/cmd/api/main.go
```diff
@@ -89,53 +90,56 @@ func (s *server) CreateTaskRunResult(ctx context.Context, req *pb.CreateTaskRunR
...
+			return nil, fmt.Errorf("failed to update taskrun: %w", err)
+		}
+	}
+	blobData, err := proto.Marshal(updateTaskRun)
 	if err != nil {
 		log.Println("taskrun marshaling error: ", err)
 		return nil, fmt.Errorf("taskrun marshaling error: %w", err)
 	}
-	taskrunMeta := taskrunFromClient.GetMetadata()
+	tx, _ := s.db.Begin()
```
wlynch, line 130: Check error

results/cmd/api/main.go
```diff
@@ -89,53 +90,56 @@ func (s *server) CreateTaskRunResult(ctx context.Context, req *pb.CreateTaskRunR
...
+		log.Printf("failed to update a existing taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to update a exsiting taskrun: %w", err)
+	}
+	taskrunMeta := updateTaskRun.GetMetadata()
 	if _, err := statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, req.GetResultsId()); err != nil {
+		tx.Rollback()
 		log.Printf("failed to execute update of a new taskrun: %v\n", err)
 		return nil, fmt.Errorf("failed to execute update of a new taskrun: %w", err)
 	}
+	tx.Commit()
```
wlynch, line 142: Thanks for adding this! We're also going to need to move the read into the transaction as well to ensure the atomic read-modify-write.

results/cmd/api/main_test.go
```diff
@@ -118,37 +119,287 @@ func TestUpdateTaskRun(t *testing.T) {
...
-	r, err = srv.UpdateTaskRunResult(ctx, &pb.UpdateTaskRunRequest{TaskRun: &pb.TaskRun{ApiVersion: "v1alpha1", Metadata: objectmeta}, ResultsId: r.GetResultsId()})
-	if err != nil {
-		t.Fatalf("could not update taskrun: %v", err)
-	}
-
-	// Validate by checking if we can get the updated taskrun
-	rows, err := srv.db.Query("SELECT taskrunlog FROM taskrun WHERE results_id = ?", r.GetResultsId())
-	if err != nil {
-		t.Fatalf("failed to query on database: %v", err)
+	// Validate by checking if ouput equlas expected TaskRun
```
wlynch, line 126: No changes needed here - just want to say this looks great and really thorough. Thanks! 😃

results/cmd/api/main_test.go
```diff
@@ -118,37 +119,287 @@ func TestUpdateTaskRun(t *testing.T) {
...
+			for rows.Next() {
+				var taskrunblob []byte
+				taskrun := &pb.TaskRun{}
+				rows.Scan(&taskrunblob)
+				if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+					t.Fatal("unmarshaling error: ", err)
+				}
+				if diff := cmp.Diff(taskrun.String(), tc.expect.String()); diff != "" {
+					t.Fatalf("Update Function not properly implemented: %v, %v, %v", diff, tc.name, tc.fieldmask.String())
+					t.Fatal(tc.name)
```
wlynch, line 399: I'm not sure if this will ever print given the fatal right before this. I think this can be removed though since we print the name in the last line.

results/cmd/api/main_test.go
yuege01: done

results/.vscode/launch.json
yuege01: done

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done

results/cmd/api/main_test.go

results/cmd/api/main.go

results/cmd/api/main.go
```diff
@@ -89,53 +90,60 @@ func (s *server) CreateTaskRunResult(ctx context.Context, req *pb.CreateTaskRunR
...
 	}
-	taskrunMeta := taskrunFromClient.GetMetadata()
+	statement, err := s.db.Prepare("UPDATE taskrun SET name = ?, namespace = ?, taskrunlog = ? WHERE results_id = ?")
+	if err != nil {
+		log.Printf("failed to update a existing taskrun: %v\n", err)
+		return nil, fmt.Errorf("failed to update a exsiting taskrun: %w", err)
+	}
+	taskrunMeta := updateTaskRun.GetMetadata()
 	if _, err := statement.Exec(taskrunMeta.GetName(), taskrunMeta.GetNamespace(), blobData, req.GetResultsId()); err != nil {
+		tx.Rollback()
```
wlynch, line 142: Do we need to rollback anywhere there is an error, or only on the exec?

results/cmd/api/main_test.go
```diff
@@ -118,37 +119,286 @@ func TestUpdateTaskRun(t *testing.T) {
...
-		}
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			r, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{TaskRun: tc.in})
+			if err != nil {
+				t.Fatalf("could not create taskrun: %v", err)
+			}
+			// Update the created taskrun
+			fieldMask := tc.fieldmask
+			if tc.name == "Test no Mask" {
```
wlynch, line 379: Now that you're passing in a pointer, you should be able to omit this.

results/cmd/api/main.go
yuege01: I think we only need to rollback only when we do update or any modification operations on database. Other errors do not damage the status of database

results/cmd/api/main_test.go

PR #596, opened 2020-07-20
Update the GitHub Config settings for Generators

YolandaDu1997:
# Changes

This PR aims to add more fields to the input Github config settings tohelp user specify their workflow with more flexibilty for [Tekton Generators](https://github.com/tektoncd/pipeline/issues/2590).

The changes includes:
1. Add fields `branch`, `storage`, `secretName`, `secretKey`, `serviceAccountName` to the `GitHubSpec` structure.
2. Add CEL in the triggers to specify the branch.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/generators/README.md
+++ b/generators/README.md
@@ -4,6 +4,11 @@ This project contains experimental code to create a tool for generating Tekton s
 
 See [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590) information and background.
 
+## GitHub token
+You will need to create a [GitHub Personal Access Token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token#creating-a-token) and set it in the Kubernetes scecret `github` in the key `token` like this:
+```
+kubectl create secret generic github --from-literal token="YOUR_GITHUB_PERSONAL_ACCESS_TOKEN"
+```
 ## Webhook secret
 You would expect to create a [Webhook secret token](https://developer.github.com/webhooks/securing/#setting-your-secret-token) and configure the GitHub webhook to use this value.
 You can contain this value in the Kubernetes secret like this :
@@ -12,7 +17,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: webhook-secret
-  type: Opaque
+type: Opaque
 stringData:
   secretToken: "YOUR-WEBHOOK-SECRET-TOKEN"
 ```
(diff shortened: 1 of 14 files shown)


generators/README.md
```diff
@@ -4,6 +4,11 @@ This project contains experimental code to create a tool for generating Tekton s
 
 See [tektoncd/pipeline/#2590](https://github.com/tektoncd/pipeline/issues/2590) information and background.
 
+## GitHub token
+You will need to create a [GitHub Personal Access Token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token#creating-a-token) and set it in the Kubernetes scecret `github` in the key `token` like this:
+```
+kubectl create secret generic github --from-literal token="YOUR_GITHUB_PERSONAL_ACCESS_KTOKEN"
```
wlynch, line 10:
```suggestion
kubectl create secret generic github --from-literal token="YOUR_GITHUB_PERSONAL_ACCESS_TOKEN"
```

generators/pkg/generator/generate.go
```diff
@@ -23,12 +23,17 @@ type GitHub struct {
...
 type GitHubSpec struct {
-	URL   string         `json:"url,omitempty"`
-	Steps []v1beta1.Step `json:"steps,omitempty"`
+	URL                string         `json:"url,omitempty"`
+	Branch             string         `json:"branch,omitempty"`
+	Storage            string         `json:"storage,omitempty"`
+	SecretName         string         `json:"secretName,omitempty"`
+	SecretKey          string         `json:"secretKey,omitempty"`
+	ServiceAccountName string         `json:"serviceAccountName,omitempty"`
+	Steps              []v1beta1.Step `json:"steps,omitempty"`
```
wlynch, line 32: It'd be great to document these fields in the read me - what are the expected values, what are there default values, etc.

generators/pkg/generator/generate.go
```diff
@@ -334,15 +400,20 @@ func GenerateTrigger(p *v1beta1.Pipeline) *trigger {
...
-									SecretKey:  "secretToken",
-									SecretName: "github-secret",
+									SecretKey:  g.Spec.SecretKey,
+									SecretName: g.Spec.SecretName,
 								},
 							},
 						},
+						{
+							CEL: &v1alpha1.CELInterceptor{
+								Filter: "body.head.ref == " + g.Spec.Branch,
```
wlynch, line 410:
```suggestion
								Filter: "body.base.ref == " + g.Spec.Branch,
```

I think we want base here, since this is the ref you want to merge into. e.g. for this PR, base == master and head == generators-pr.

generators/pkg/generator/generate.go
```diff
@@ -41,6 +46,14 @@ func GenerateTask(github *GitHub) *v1beta1.Task {
 		labels = make(map[string]string)
 	}
 	labels["generator.tekton.dev"] = github.Name
+
+	// DOCKER_CONFIG is used for kaniko to detect docker credential
+	env := &github.Spec.Steps[0].Env
+	*env = append(*env, corev1.EnvVar{
+		Name:  "DOCKER_CONFIG",
+		Value: "/tekton/home/.docker/",
+	})
```
wlynch, line 55: These look like options that are step specific. I think we should remove these, letting users set them manually in their steps if they want it. What do you think?

YolandaDu1997: I agree. The reason why I add it here is that when I apply the resources on the cluster, the pipelinerun failed since the DOCKER_CONFIG is required to allow kaniko to detect docker credential.

YolandaDu1997:
Done.

generators/pkg/writer/testdata/config.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   steps:
   - args:
     - --context=dir://$(workspaces.input.path)/src
-    - --destination=gcr.io/<use your project>/kaniko-test
+    - --destination=gcr.io/tekton-yolandadu/kaniko-test
```
dibyom, line 13: is this intentional?

dibyom: Also, Kaniko usually requires a secret or a service account that has enough credentials to push images. 
I think this is usually already configured for GKE. But other k8s installations might have to configure this

YolandaDu1997: It's probably because I paste from another config. Sure I can document it in README.

YolandaDu1997: By the way for the service account, can it be the same as the one required for the triggers?

dibyom:
This is looking good to me! Let's squash up the commits?

YolandaDu1997:
> This is looking good to me! Let's squash up the commits?

sure!

PR #599, opened 2020-07-23
Implement query API of Tekton Results API

yuege01:
# Changes
Created a new ListTaskRunsResult interface to handle queries from users
Added a filter field to ListTaskRunsRequest to specify a query string
Used cel-go library to process queries from users and return taskruns back conforming to those queries
Added a test to test if we can correctly query from a list of inserted taskruns using filter

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

results/cmd/api/main.go
```diff
@@ -161,3 +172,79 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
 	}
 	return nil, nil
 }
+
+// ListTaskRunsResult receives a ListTaskRunRequest from users and return to users a list of TaskRuns according to query
+func (s *server) ListTaskRunsResult(ctx context.Context, req *pb.ListTaskRunsRequest) (*pb.ListTaskRunsResponse, error) {
+	env, err := cel.NewEnv(
```
wlynch, line 178: Check the error.

results/cmd/api/main.go
```diff
@@ -161,3 +172,79 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+		cel.Types(&pb.TaskRun{}),
+		cel.Declarations(decls.NewIdent("taskrun", decls.NewObjectType("tekton.TaskRun"), nil)),
+	)
```
wlynch, line 181: This is always going to be the same for the server, regardless of the request, right? We should pull this into a `New()` func and set it once on server creation.

results/cmd/api/main.go
```diff
@@ -161,3 +172,79 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	ast, issues := env.Compile(req.GetFilter())
+	if issues != nil && issues.Err() != nil {
+		log.Printf("type-check error: %s", issues.Err())
+		return nil, fmt.Errorf("failed to list taskruns: %w", err)
```
wlynch, line 185: We should return a more descriptive error. e.g. was it cause by the user? what should they do to resolve it?

We should also make sure to set a gRPC status code in the response.

wlynch: We should make similar changes for the errors below.

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
 }
+func TestListTaskRuns(t *testing.T) {
+	// Create a temporary database
+	srv, err := setupTestDB(t)
+	if err != nil {
+		t.Fatalf("failed to setup db: %v", err)
+	}
+
+	// Connect to fake server and create a new taskrun
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
wlynch, line 448: Go tests have a timeout flag that can be configured, so we likely don't need to do this unless we have a particular reason to have a shorter per-request timeout.

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+	t4 := &pb.TaskRun{
+		ApiVersion: "v1beta1",
+		Metadata: &pb.ObjectMeta{
+			Uid:       "543535",
+			Name:      "newtaskrun",
+			Namespace: "demo",
+		},
+	}
+	t5 := &pb.TaskRun{
+		ApiVersion: "v1apha1",
```
wlynch, line 485:
```suggestion
		ApiVersion: "v1alpha1",
```

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+		Metadata: &pb.ObjectMeta{
+			Uid:       "543535",
+			Name:      "newtaskrun",
+			Namespace: "demo",
+		},
+	}
+	testList := []*pb.TaskRun{t1, t2, t3, t4, t5}
+	expect := []string{t1.String(), t3.String(), t4.String()}
+	sort.Strings(expect)
```
wlynch, line 494: It's a good practice to group variables together where they are used. Since this isn't used until the very last comparison, we should move these down there.

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+	for _, tt := range testList {
+		_, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
+			TaskRun: tt,
+		})
+		if err != nil {
+			t.Fatalf("could not create taskrun: %v", err)
+		}
+	}
+	if err != nil {
```
wlynch, line 503: I don't think this is checking any particular error. Remove?

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+		},
+	}
+	testList := []*pb.TaskRun{t1, t2, t3, t4, t5}
+	expect := []string{t1.String(), t3.String(), t4.String()}
+	sort.Strings(expect)
+	for _, tt := range testList {
+		_, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
+			TaskRun: tt,
+		})
+		if err != nil {
```
wlynch, line 499:
```suggestion
		if _, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
			TaskRun: tt,
		}; err != nil {
```

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			t.Fatalf("could not create taskrun: %v", err)
+		}
+	}
+	if err != nil {
+		t.Fatalf("could not create taskrun: %v", err)
+	}
+	// Test if we can find inserted taskruns
+	res, err := srv.ListTaskRunsResult(ctx, &pb.ListTaskRunsRequest{
+		Filter: `taskrun.api_version=="v1beta1" && taskrun.metadata.namespace=="demo"`,
```
wlynch, line 508: It'd be great to have more tests for different cases, logical operators, error cases, etc.

results/cmd/api/main_test.go
```diff
@@ -186,8 +432,92 @@ func TestDeleteTaskRun(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("could not list taskruns: %v", err)
+	}
+	var resList []string
+	for _, t := range res.GetItems() {
+		resList = append(resList, t.String())
+	}
+	sort.Strings(resList)
+	for i := 0; i < len(resList); i++ {
+		if diff := cmp.Diff(resList[i], expect[i]); diff != "" {
```
wlynch, line 519: We should compare the full object list using cmp.Diff.

https://pkg.go.dev/sort?tab=doc#Slice will likely be helpful here!

results/proto/api.proto
```diff
@@ -40,12 +42,14 @@ message GetTaskRunRequest {
 
 message ListTaskRunsRequest {
 	// TODO: document query format.
-	string query = 1;
-	// TODO: page_token, etc.
+	string filter = 1;
+	int32 page_size = 2;
+	string page_token = 3;
```
wlynch, line 47: Let's not include any pagination fields in the API until we support this.

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/proto/api.proto
yuege01: done

results/cmd/api/main.go

results/cmd/api/main_test.go

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+		log.Printf("failed to query on database: %v", err)
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
+	}
+	var taskRunList []*pb.TaskRun
+	for rows.Next() {
+		taskrun := &pb.TaskRun{}
+		var taskrunblob []byte
+		rows.Scan(&taskrunblob)
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
```
dibyom, line 197: this will crash the server because after Fatal will log and then call os.Exit().
We should continue to process

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+		taskRunList = append(taskRunList, taskrun)
+	}
+	// return all taskruns back to users if empty query is given
+	if req.GetFilter() == "" {
+		return &pb.ListTaskRunsResponse{Items: taskRunList}, nil
+	}
+	ast, issues := s.env.Compile(req.GetFilter())
+	if issues != nil && issues.Err() != nil {
+		log.Printf("type-check error: %s", issues.Err())
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to wrong type in the query, please try to enter a query with correct type again")
```
dibyom, line 210: It might also be useful to return the actual issues.Err() as part of the response so that the user knows exactly what went wrong

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
+	}
+	var taskRunList []*pb.TaskRun
+	for rows.Next() {
+		taskrun := &pb.TaskRun{}
+		var taskrunblob []byte
+		rows.Scan(&taskrunblob)
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Fatal("unmarshaling error: ", err)
+			return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
```
dibyom, line 198: Also `codes.Internal` might be a better error code here since this is an internal error ?

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	}
+	ast, issues := s.env.Compile(req.GetFilter())
+	if issues != nil && issues.Err() != nil {
+		log.Printf("type-check error: %s", issues.Err())
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to wrong type in the query, please try to enter a query with correct type again")
+	}
+	prg, err := s.env.Program(ast)
+	if err != nil {
+		log.Printf("program construction error: %s", err)
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
```
dibyom, line 215: This one looks like a genuine internal error, so codes.Internal might be useful here

results/cmd/api/main.go
dibyom, line 210: codes.InvalidArgument since this error happens if the users sends the wrong type?

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	}
+	// filter from all taskruns
+	var resList []*pb.TaskRun
+	for _, taskrun := range taskRunList {
+		out, _, err := prg.Eval(map[string]interface{}{
+			"taskrun": taskrun,
+		})
+		if err != nil {
+			log.Printf("failed to evaluate the expression: %v", err)
+			return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to wrong type in the query, please try to enter a query with correct type again")
```
dibyom, line 225: `codes.InvalidArgument` -> also might be useful to return the error message to the user so that they can debug easily?

results/cmd/api/main.go
```diff
@@ -193,10 +254,9 @@ func (s *server) getTaskRunByID(id string) (*pb.TaskRun, error) {
 		}
 		rows.Scan(&taskrunblob)
 		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
-			log.Fatal("unmarshaling error: ", err)
+			log.Println("unmarshaling error: ", err)
```
dibyom, line 257: use `Printf` ?

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+			return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
+
+		}
+		taskRunList = append(taskRunList, taskrun)
+	}
+	// return all taskruns back to users if empty query is given
+	if req.GetFilter() == "" {
+		return &pb.ListTaskRunsResponse{Items: taskRunList}, nil
+	}
+	ast, issues := s.env.Compile(req.GetFilter())
```
dibyom, line 207: small optimization -> We should be able to move the following lines where we setup CEL and check the filter before we fetch the taskruns from DB. This way we fail fast, and do not have to make a DB call when the filter is invalid.

results/cmd/api/main.go
```diff
@@ -43,29 +46,36 @@ func main() {
...
+	}
+	srv := &server{db: db, env: env}
 	// Listen for gRPC requests.
 	lis, err := net.Listen("tcp", port)
 	if err != nil {
 		log.Fatalf("failed to listen: %v", err)
 	}
 	s := grpc.NewServer()
 	pb.RegisterResultsServer(s, srv)
+	reflection.Register(s)
```
dibyom, line 65: question: why do we need this?

results/cmd/api/main_test.go
```diff
@@ -87,9 +96,7 @@ func TestGetTaskRun(t *testing.T) {
 	}
 
 	// Connect to fake server and create a new taskrun
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-	r, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
+	r, err := srv.CreateTaskRunResult(nil, &pb.CreateTaskRunRequest{
```
dibyom, line 99: why did we switch from context.Background() to nil?

results/cmd/api/main_test.go
```diff
@@ -100,10 +107,9 @@ func TestGetTaskRun(t *testing.T) {
 	if err != nil {
 		t.Fatalf("could not create taskrun: %v", err)
 	}
-	t.Logf("Created taskrun: %s", r.String())
 
 	// Test if we can find inserted taskrun
-	res, err := srv.GetTaskRunResult(ctx, &pb.GetTaskRunRequest{ResultsId: r.GetResultsId()})
+	res, err := srv.GetTaskRunResult(nil, &pb.GetTaskRunRequest{ResultsId: r.GetResultsId()})
```
dibyom, line 112: same question as above?

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			expect: []*pb.TaskRun{t1, t3, t4},
+		},
+		{
+			name:   "test logical &&",
+			list:   testList,
+			filter: `taskrun.api_version=="v1beta1" && taskrun.metadata.namespace=="demo"`,
+			expect: []*pb.TaskRun{t3, t4},
+		},
+		{
+			name:   "test logical ||",
```
dibyom, line 510: we may be able to combine some of these tests e.g. test a complex query that includes an logical AND, OR as well as startWith?

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+		},
+		{
+			name:   "test invalid field",
+			list:   testList,
+			filter: `task.name=="newtaskrun"`,
+		},
+		{
+			name:   "test invalid value type",
+			list:   testList,
+			filter: `taskrun.name==newtaskrun`,
```
dibyom, line 541: Is this an invalid type due to the missing `""`?

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			list:   testList,
+			filter: `taskrun.metadata.name.startsWith("new")`,
+			expect: []*pb.TaskRun{t4, t5},
+		},
+		{
+			name:   "test endWith",
+			list:   testList,
+			filter: `taskrun.metadata.name.endsWith("run")`,
+			expect: []*pb.TaskRun{t1, t3, t4, t5},
+		},
```
wlynch, line 526: From the perspective of our code, these more or less behave the same. They can all be categorized as "filters that result in multiple true responses by cel-go". Using the simple query is probably sufficient.

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+		{
+			name:   "test empty filter",
+			list:   testList,
+			filter: "",
+			expect: testList,
+		},
+		{
+			name:   "test invalid field",
+			list:   testList,
+			filter: `task.name=="newtaskrun"`,
```
wlynch, line 536: nice! I think testing invalid fields outside of our defined top level field would also be valuable (e.g. `taskrun.doesnotexist`)

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+	}
+	t5 := &pb.TaskRun{
+		ApiVersion: "v1alpha1",
+		Metadata: &pb.ObjectMeta{
+			Uid:       "543535",
+			Name:      "newtaskrun",
+			Namespace: "official",
+		},
+	}
+	testList := []*pb.TaskRun{t1, t2, t3, t4, t5}
```
wlynch, line 490: Let's rename this to be a little more descriptive (since this isn't actually a list of tests). Perhaps `taskruns`?

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			t.Fatalf("could not create taskrun: %v", err)
+		}
+	}
+	// Test if we can find inserted taskruns
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			res, err := srv.ListTaskRunsResult(nil, &pb.ListTaskRunsRequest{
+				Filter: tc.filter,
+			})
+			if tc.name == "test invalid field" || tc.name == "test invalid value type" {
```
wlynch, line 563: Instead of switching on the field, we should just include the expected code in the test case.

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+				taskList := res.GetItems()
+				sort.SliceStable(taskList, func(i, j int) bool {
+					return taskList[i].String() < taskList[j].String()
+				})
+				sort.SliceStable(tc.expect, func(i, j int) bool {
+					return tc.expect[i].String() < tc.expect[j].String()
+				})
+				if diff := cmp.Diff(taskList, tc.expect, protocmp.Transform()); diff != "" {
+					t.Fatalf("could not get the same taskrun: %v", diff)
+				}
```
wlynch, line 580: Nothing to do here. Just wanted to say thanks for keeping slice ordering in mind! :D

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			name:   "test invalid value type",
+			list:   testList,
+			filter: `taskrun.name==newtaskrun`,
+		},
+		{
+			name:   "test value not existing in the server record",
+			list:   testList,
+			filter: `taskrun.metadata.name=="notaskrun"`,
+			expect: nil,
+		},
```
wlynch, line 548: Other interesting tests that come to mind:
- What if the user enters a non-boolean (but valid) filter?
- What if a user enters in a random word instead of a boolean expression? (e.g. `tekton`)
- Does case sensitivity matter?

results/cmd/api/main.go
yuege01: Without this line, I cannot use grpc-cli:)

yuege01: I will add a comment for this

results/cmd/api/main_test.go
yuege01: Go tests have a timeout flag that can be configured, so we likely don't need to do this unless we have a particular reason to have a shorter per-request timeout.

results/cmd/api/main_test.go
yuege01: Go tests have a timeout flag that can be configured, so we likely don't need to do this unless we have a particular reason to have a shorter per-request timeout.

results/cmd/api/main_test.go
yuege01: Yes, I will clarify it in the test name

results/cmd/api/main_test.go
yuege01: I decide to use context.Background only instead of nil

results/cmd/api/main_test.go
yuege01: same as above

results/cmd/api/main_test.go
yuege01: I have thought for a while and decide to delete some of these tests, basically they are the same as a feature of cel. So a simple query test is enough

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go
yuege01: done

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go
yuege01: done

results/cmd/api/main_test.go
yuege01: done

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
 	return nil, nil
 }
 
+// ListTaskRunsResult receives a ListTaskRunRequest from users and return to users a list of TaskRuns according to the query
+func (s *server) ListTaskRunsResult(ctx context.Context, req *pb.ListTaskRunsRequest) (*pb.ListTaskRunsResponse, error) {
+	// get all taskruns from database
+	rows, err := s.db.Query("SELECT taskrunlog FROM taskrun")
+	if err != nil {
+		log.Printf("failed to query on database: %v", err)
+		return nil, status.Errorf(codes.NotFound, "No TaskRuns found for the query string due to internal server error, please try to enter the query again")
```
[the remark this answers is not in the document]

results/cmd/api/main.go
```diff
@@ -170,16 +180,67 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	}
+	var taskRunList []*pb.TaskRun
+	for rows.Next() {
+		taskrun := &pb.TaskRun{}
+		var taskrunblob []byte
+		rows.Scan(&taskrunblob)
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
```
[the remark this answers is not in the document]

results/cmd/api/main.go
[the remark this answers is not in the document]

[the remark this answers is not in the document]
yuege01: Note: Also I have to put code `prg, err := s.env.Program(ast)` after database query and checking empty filter, or there will be a nullpointer error crashing the server.

results/cmd/api/main.go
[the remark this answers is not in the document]

results/cmd/api/main.go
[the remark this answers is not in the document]

results/cmd/api/main_test.go
```diff
@@ -29,9 +33,7 @@ func TestCreateTaskRun(t *testing.T) {
 	}
 
 	// connect to fake server and do testing
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-	if _, err := srv.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
+	if _, err := srv.CreateTaskRunResult(nil, &pb.CreateTaskRunRequest{
```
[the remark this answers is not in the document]

results/cmd/api/main_test.go
```diff
@@ -56,10 +58,17 @@ func setupTestDB(t *testing.T) (*server, error) {
...
 	if err != nil {
 		t.Fatalf("failed to open the results.db: %v", err)
 	}
+	env, err := cel.NewEnv(
+		cel.Types(&pb.TaskRun{}),
+		cel.Declarations(decls.NewIdent("taskrun", decls.NewObjectType("tekton.TaskRun"), nil)),
+	)
+	if err != nil {
+		t.Fatalf("failed to create environment for filter: %v", err)
+	}
```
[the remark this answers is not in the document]

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			expect: nil,
+		},
+	}
+	for _, ts := range testList {
+		if _, err := srv.CreateTaskRunResult(nil, &pb.CreateTaskRunRequest{
+			TaskRun: ts,
+		}); err != nil {
+			t.Fatalf("could not create taskrun: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			})
+			if tc.name == "test invalid field" || tc.name == "test invalid value type" {
+				if err != nil && status.Code(err) != codes.NotFound {
+					t.Fatalf("failed on test %v: %v", status.Code(err), err)
+				}
+			} else {
+				if err != nil {
+					t.Fatalf("could not list taskruns: %v", err)
+				}
+				taskList := res.GetItems()
```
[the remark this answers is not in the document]

results/cmd/api/main_test.go

results/cmd/api/main_test.go

results/cmd/api/main_test.go
yuege01: Thank you!

results/cmd/api/main_test.go
yuege01: done

results/cmd/api/main_test.go

results/cmd/api/main_test.go
yuege01: done

results/cmd/api/main_test.go
```diff
@@ -437,3 +439,146 @@ func TestDeleteTaskRun(t *testing.T) {
...
+		Metadata: &pb.ObjectMeta{
+			Uid:       "543535",
+			Name:      "newtaskrun",
+			Namespace: "official",
+		},
+	}
+	testList := []*pb.TaskRun{t1, t2, t3, t4, t5}
+	tt := []struct {
+		name   string
+		list   []*pb.TaskRun
```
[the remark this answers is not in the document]

results/cmd/api/main.go
```diff
@@ -170,16 +180,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	for rows.Next() {
+		var taskrunblob []byte
+		if err := rows.Scan(&taskrunblob); err != nil {
+			log.Printf("failed to scan a row in query results: %v", err)
+			return nil, status.Errorf(codes.Internal, "failed to parse the query string due to internal server error, please try to enter the query again: %v", err)
+		}
+		taskrun := &pb.TaskRun{}
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Printf("unmarshaling error: %v", err)
+			return nil, status.Errorf(codes.Internal, "failed to parse the query string due to internal server error, please try to enter the query again: %v", err)
```
wlynch, line 207: I'd recommend returning a more detailed error, ideally describing what step is failing. Most users will see the Internal error status and give up, but being able to see a log and see what step failed is very useful when debugging why something went wrong.

results/cmd/api/main.go
```diff
@@ -170,16 +180,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	ast, issues := s.env.Compile(req.GetFilter())
+	if issues != nil && issues.Err() != nil && req.GetFilter() != "" {
+		log.Printf("type-check error: %s", issues.Err())
+		return nil, status.Errorf(codes.InvalidArgument, "No TaskRuns found for the query string due to wrong type in the query, please try to enter a query with correct type again: %v", issues.Err())
+	}
+	// get all taskruns from database
+	rows, err := s.db.Query("SELECT taskrunlog FROM taskrun")
+	if err != nil {
+		log.Printf("failed to query on database: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse the query string due to internal server error, please try to enter the query again: %v", err)
```
wlynch, line 195: InvalidArgument usually indicates things went wrong because of the user input, but here we aren't using the query string / any other user input.

We should change the status code and the error message, since an error here would indicate an issue with the server. Because of this, perhaps internal makes sense?

results/cmd/api/main.go
```diff
@@ -170,16 +180,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	}
+	// return all taskruns back to users if empty query is given
+	if req.GetFilter() == "" {
+		return &pb.ListTaskRunsResponse{Items: taskRunList}, nil
+	}
+	// filter from all taskruns
+	prg, err := s.env.Program(ast)
+	if err != nil {
+		log.Printf("program construction error: %s", err)
+		return nil, status.Errorf(codes.InvalidArgument, "No TaskRuns found for the query string due to wrong type in the query, please try to enter a query with correct type again: %v", issues.Err())
```
wlynch, line 220: Will the error always be caused by a wrong type, or are there other errors that may occur here? If so, we should change the error message to better match what's failing.

results/cmd/api/main_test.go
```diff
@@ -437,3 +443,133 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			filter: `taskrun.MetaData.name=="notaskrun"`,
+		},
+	}
+	// Test if we can find inserted taskruns
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			res, err := srv.ListTaskRunsResult(ctx, &pb.ListTaskRunsRequest{
+				Filter: tc.filter,
+			})
+			if err != nil && status.Code(err) != codes.InvalidArgument {
```
wlynch, line 559: We should be more explicit here and check for the returned status. It's difficult to determine whether something is passing because they returned no values vs they returned InvalidArgument.

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main_test.go

results/cmd/api/main.go
wlynch: I don't think this was addressed. This would help reduce some between main.go and main_test.go! :)

results/cmd/api/main_test.go
```diff
@@ -437,3 +443,144 @@ func TestDeleteTaskRun(t *testing.T) {
...
+				t.Fatalf("failed on test %v: %v", tc.name, err)
+			}
+			gotList := res.GetItems()
+			sort.SliceStable(gotList, func(i, j int) bool {
+				return gotList[i].String() < gotList[j].String()
+			})
+			sort.SliceStable(tc.expect, func(i, j int) bool {
+				return tc.expect[i].String() < tc.expect[j].String()
+			})
+			if diff := cmp.Diff(gotList, tc.expect, protocmp.Transform()); res != nil && status.Code(err) != codes.OK && diff != "" {
```
wlynch, line 580: If I'm understanding this, this will ignore checks in cases where the response is nil or there was no returned error. This effectively skips diffs for successful responses. I think we can just simplify this to check the responses, regardless of whether they are nil or not. Status checking can be handled in another statement (likely the one above).

```suggestion
			if diff := cmp.Diff(gotList, tc.expect, protocmp.Transform()); diff != "" {
```

results/cmd/api/main_test.go
```diff
@@ -437,3 +443,144 @@ func TestDeleteTaskRun(t *testing.T) {
...
+			expectStatus: codes.InvalidArgument,
+		},
+	}
+	// Test if we can find inserted taskruns
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			res, err := srv.ListTaskRunsResult(ctx, &pb.ListTaskRunsRequest{
+				Filter: tc.filter,
+			})
+			if res == nil && status.Code(err) != codes.InvalidArgument {
```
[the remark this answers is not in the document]

results/cmd/api/main_test.go

results/cmd/api/main.go
```diff
@@ -170,16 +173,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	ast, issues := s.env.Compile(req.GetFilter())
+	if issues != nil && issues.Err() != nil && req.GetFilter() != "" {
+		log.Printf("type-check error: %s", issues.Err())
+		return nil, status.Errorf(codes.InvalidArgument, "Error occurred during filter parse step, no TaskRuns found for the query string due to invalid field, invalid function to evaluate filter or missing double quotes around field value, please try to enter a query with correct type again: %v", issues.Err())
+	}
+	// get all taskruns from database
+	rows, err := s.db.Query("SELECT taskrunlog FROM taskrun")
+	if err != nil {
+		log.Printf("failed to query on database: %v", err)
+		return nil, status.Errorf(codes.Internal, "Error occurred during preparation for search of query results step, failed to parse the query string due to internal server error, please try to enter the query again or check if connection to server is set up: %v", err)
```
wlynch, line 188:
```suggestion
		return nil, status.Errorf(codes.Internal, "failed to query results: %v", err)
```

results/cmd/api/main.go
```diff
@@ -170,16 +173,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+	}
+	var taskRunList []*pb.TaskRun
+	for rows.Next() {
+		var taskrunblob []byte
+		if err := rows.Scan(&taskrunblob); err != nil {
+			log.Printf("failed to scan a row in query results: %v", err)
+			return nil, status.Errorf(codes.Internal, "Error occurred during query results parsing step, failed to parse the query string due to internal server error, please try to enter the query again or check if connection to server is set up: %v", err)
```
wlynch, line 195:
```suggestion
			return nil, status.Errorf(codes.Internal, "failed to read result data: %v", err)
```

results/cmd/api/main.go
```diff
@@ -170,16 +173,71 @@ func (s *server) DeleteTaskRunResult(ctx context.Context, req *pb.DeleteTaskRunR
...
+		}
+		taskrun := &pb.TaskRun{}
+		if err := proto.Unmarshal(taskrunblob, taskrun); err != nil {
+			log.Printf("unmarshaling error: %v", err)
+			return nil, status.Errorf(codes.Internal, "Error occurred during query results parsing step, failed to parse the query string due to internal server error, please try to enter the query again or check if connection to server is set up: %v", err)
```
wlynch, line 200:
```suggestion
			return nil, status.Errorf(codes.Internal, "failed to parse result data: %v", err)
```

dibyom:
Yeah, I think once the error messages are simplified, we can squash and merge!

results/cmd/api/main.go

results/cmd/api/main.go

results/cmd/api/main.go

dibyom:
minor issue: the commit message is very long and seems to include the commit messages for all squashed commits. What I'd recommend is removing most of those and just adding a few lines that describe the commit. The first few lines of the current commit message seems to be good. 
```
Implement query API of Tekton Results API
Created a new ListTaskRunsResult interface to handle queries from users
Added a filter field to ListTaskRunsRequest to specify a query string
Used cel-go library to process queries from users and return taskruns back conforming to those queries
Added a test to test if we can correctly query from a list of inserted taskruns using filter
```

yuege01:
> minor issue: the commit message is very long and seems to include the commit messages for all squashed commits. What I'd recommend is removing most of those and just adding a few lines that describe the commit. The first few lines of the current commit message seems to be good.
> 
> ```
> Implement query API of Tekton Results API
> Created a new ListTaskRunsResult interface to handle queries from users
> Added a filter field to ListTaskRunsRequest to specify a query string
> Used cel-go library to process queries from users and return taskruns back conforming to those queries
> Added a test to test if we can correctly query from a list of inserted taskruns using filter
> ```

yeah, I've simplified commit messages, I just wanted to make it more descriptive:)

PR #600, opened 2020-07-23
Add support for PipelineRun in pkg

YolandaDu1997:
# Changes

This PR aims to add the alternative configs (pipelinerun) for users to run the pipeline for [Tekton Generators](https://github.com/tektoncd/pipeline/issues/2590).

The changes include:

1. Add support to generate pipelinerun for users to run. Because we want to provide more ways for users to run, not only the triggers.
2. Change the StringVal `$(params.gitrevision)` to `github.Spec.Revision` in the params of pipeline, because param substitution wasn't working properly with finally for the time being. This change can be removed when this param substitution bug is fixed in the future.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/generators/pkg/generator/generate.go
+++ b/generators/pkg/generator/generate.go
@@ -5,6 +5,7 @@ package generator
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
@@ -41,7 +42,12 @@ type trigger struct {
 
 // GenerateTask generates Tekton Task
 // from simplified Github configs.
-func GenerateTask(github *GitHub) *v1beta1.Task {
+func GenerateTask(github *GitHub) (*v1beta1.Task, error) {
+	// check input validation
+	if github == nil {
+		return nil, fmt.Errorf("invalid github configuration")
+	}
+
 	labels := github.Labels
 	if labels == nil {
 		labels = make(map[string]string)
@@ -66,13 +72,18 @@ func GenerateTask(github *GitHub) *v1beta1.Task {
 			},
 			Steps: github.Spec.Steps,
 		},
-	}
+	}, nil
 
 }
 
 // GeneratePipeline generates Tekton Pipeline
 // from simplified Github configs.
 func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
+	// check input validation
+	if github == nil {
+		return nil, fmt.Errorf("invalid github configuration")
+	}
+
 	ws := "source"
 	name := github.Name + "-pipeline"
 	tasksName := []string{"fetch-git-repo", "build-from-repo", "final-set-status"}
@@ -80,6 +91,9 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
 	u, err := url.Parse(github.Spec.URL)
 	if err != nil {
 		return nil, fmt.Errorf("fail to parse the url %s: %w", github.Spec.URL, err)
+	}
+	if github.Spec.Revision == "" {
+		github.Spec.Revision = "master"
 	}
 
 	pipeline := &v1beta1.Pipeline{
@@ -162,14 +176,18 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
 							Name: "REPO_FULL_NAME",
 							Value: v1beta1.ArrayOrString{
 								Type:      v1beta1.ParamTypeString,
-								StringVal: u.Path,
+								StringVal: strings.TrimPrefix(u.Path, "/"),
 							},
 						},
 						{
 							Name: "SHA",
 							Value: v1beta1.ArrayOrString{
-								Type:      v1beta1.ParamTypeString,
-								StringVal: "$(params.gitrevision)",
+								Type: v1beta1.ParamTypeString,
+								// TODO: because for the time being, param substitution isn't working properly with finally,
+								// see the issue here: https://github.com/tektoncd/pipeline/issues/2906
+								// StringVal here should be changed to "$(params.gitrevision)" in the future
+								// when this param substitution bug is fixed in the future
+								StringVal: github.Spec.Revision,
 							},
 						},
 						{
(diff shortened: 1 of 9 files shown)


generators/cmd/cli/cmd/pipelinerun/pipelinerun.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/spf13/cobra"
+)
+
+// Command includes the pipelinerun commands
+func Command(kubeconfig string) *cobra.Command {
+	prCmd := &cobra.Command{
+		Use:   "pipelinerun",
+		Short: "Manage the generated config with PipelineRun",
+	}
+	prCmd.AddCommand(
```
dibyom, line 13: nice!

generators/pkg/generator/generate.go
```diff
@@ -162,14 +163,14 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
-								StringVal: u.Path,
+								StringVal: strings.TrimPrefix(u.Path, "/"),
 							},
 						},
 						{
 							Name: "SHA",
 							Value: v1beta1.ArrayOrString{
 								Type:      v1beta1.ParamTypeString,
-								StringVal: "$(params.gitrevision)",
+								StringVal: github.Spec.Revision,
```
wlynch, line 173: We leave a comment / TODO about why we need to do this (e.g. param substitution wasn't working properly with finally).

generators/pkg/generator/generate.go
```diff
@@ -179,6 +180,20 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+								StringVal: "https://tekon/dashboard/taskrun/log",
+							},
+						},
+						{
+							Name: "DESCRIPTION",
+							Value: v1beta1.ArrayOrString{
+								Type:      v1beta1.ParamTypeString,
+								StringVal: "test",
+							},
+						},
```
wlynch, line 196: GitHub API documentation says these are optional fields, so we should omit these until we have a meaningful value to put here.

generators/pkg/generator/generate.go
```diff
@@ -194,6 +209,148 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+		format = diskSize.Format
+	} else {
+		value = 1024 * 1024 * 1024
+		format = resource.BinarySI
+	}
+
+	if g.Spec.ServiceAccountName == "" {
+		g.Spec.ServiceAccountName = "default"
+	}
+
```
wlynch, line 311: This seems duplicative with PipelineRun. Can we reuse code / pull this into a func?

generators/pkg/generator/generate.go
```diff
@@ -194,6 +209,148 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// GenerateTaskRun generates Tekton TaskRun
+// from simplified Github configs.
+func GenerateTaskRun(g *GitHub) *v1beta1.TaskRun {
```
wlynch, line 286: Is this used anywhere outside of tests? If not, let's hold off on adding this / break it out into another branch to add in a later PR.

generators/pkg/generator/generate.go
```diff
@@ -194,6 +209,148 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+					Name: "gitrevision",
+					Value: v1beta1.ArrayOrString{
+						Type:      v1beta1.ParamTypeString,
+						StringVal: g.Spec.Revision,
+					},
+				},
+			},
+			Workspaces: []v1beta1.WorkspaceBinding{
+				{
+					Name: p.Spec.Workspaces[0].Name,
```
wlynch, line 265: Are we guaranteed that this will always be present? We may want to add validation to check that this and other fields exist.

wlynch: To provide more context, if workspaces aren't set, this is going to cause the program to panic and crash.

generators/pkg/generator/generate_test.go
```diff
@@ -90,6 +90,28 @@ func TestGeneratePipeline(t *testing.T) {
...
 
+func TestGeneratePipelineRun(t *testing.T) {
+	p := &v1beta1.Pipeline{}
+	unmarshal(t, "./testdata/pipeline.yaml", p)
+
+	got := GeneratePipelineRun(p, github)
+	want := &v1beta1.PipelineRun{}
+	unmarshal(t, "./testdata/run.yaml", want)
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("Pipeline mismatch (-want +got):\n %s", diff)
```
wlynch, line 101:
```suggestion
		t.Errorf("PipelineRun mismatch (-want +got):\n %s", diff)
```

generators/pkg/generator/generate_test.go
```diff
@@ -90,6 +90,28 @@ func TestGeneratePipeline(t *testing.T) {
 	}
 }
 
+func TestGeneratePipelineRun(t *testing.T) {
```
wlynch, line 93: It would also be great to capture error / edge cases as well.

generators/pkg/generator/generate.go
YolandaDu1997: I think this workspace is present because I wire up workspaces for the GitHub repo in the generated task and pipeline. For other fields to check, maybe we can add the validation at parsing? For example, when parsing the input config, if users don't provide the url which is a required field, we could print an error.

generators/pkg/generator/generate.go
YolandaDu1997: But if I ignore these two fields, I will run into this error in the pipelinerun:
```
invalid input params for task github-set-status: missing values for these params which have no default values: [TARGET_URL DESCRIPTION]
```

YolandaDu1997:
Done!

generators/pkg/generator/generate.go
```diff
@@ -162,14 +166,15 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
...
+								StringVal: strings.TrimPrefix(u.Path, "/"),
 							},
 						},
 						{
 							Name: "SHA",
 							Value: v1beta1.ArrayOrString{
-								Type:      v1beta1.ParamTypeString,
-								StringVal: "$(params.gitrevision)",
+								Type: v1beta1.ParamTypeString,
+								// TODO: the change from "$(params.gitrevision)" can be removed when the param substitution works with finally
```
wlynch, line 176: > the change from "$(params.gitrevision)"

Someone reading the code for the first time won't know what this change was unless they dig back through the history. We should point them towards what to change it to in the future, either by linking to an issue or just indicating in the comment what the future value should be.

Let's also include a link to the underlying param substitution issue.

generators/pkg/generator/generate.go
```diff
@@ -194,27 +199,75 @@ func GeneratePipeline(github *GitHub) (*v1beta1.Pipeline, error) {
 	return pipeline, nil
 }
 
-// Generate the trigger with the given generated pipeline
-func GenerateTrigger(p *v1beta1.Pipeline, g *GitHub) *trigger {
+// Generate the pipelinerun with the given generated pipeline
+func GeneratePipelineRun(p *v1beta1.Pipeline, g *GitHub) (*v1beta1.PipelineRun, error) {
+	if p.Spec.Workspaces == nil {
```
wlynch, line 204: We should also check if `Pipeline` and `GitHub` are nil.

YolandaDu1997:
Done!

YolandaDu1997:
Done.

PR #604, opened 2020-07-27
Hub: Add cronjob to sync catalog changes to db at regular interval

SM43:
This will run a cron job every 10 minutes which will sync catalog
changes to db.

Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/hub/api/config/db-migration/15-db-cronjob.yaml
+++ b/hub/api/config/db-migration/15-db-cronjob.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: hub-sync-db
+  namespace: tekton-hub-preview
+  labels:
+    app: api
+spec:
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: sync-db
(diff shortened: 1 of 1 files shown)


hub/api/config/db-migration/15-db-cronjob.yaml
```diff
@@ -0,0 +1,23 @@
...
+    app: api
+spec:
+  schedule: "*/10 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: sync-db
+              image: byrnedo/alpine-curl:0.1.8
```
sthaha, line 17: please take note when using images. it is better to use official images whenever possible - https://hub.docker.com/r/curlimages/curl

hub/api/config/db-migration/15-db-cronjob.yaml
```diff
@@ -0,0 +1,23 @@
...
+              args:
```
sthaha, line 18: this is going to a lot more legible if we used `[ ... ]`
e.g.: `[ -s, -q, -X, POST, api/resources/sync ]` 
We can directly access the service instead of the ingress, can't we?

PR #609, opened 2020-07-29
Add update to watcher and results_id to annotation field of taskrun

yuege01:
This pull request implements a feature that checks if a taskrun exists in results server and create a new one if it does not or update existing one otherwise
Set up connection to gke cluster and add a results_id returned from results server to taskrun in the cluster
Update taskrun in results server if a results_id key exists in the taskrun received from cluster by watcher

# Changes

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

results/cmd/watcher/main.go
```diff
@@ -108,12 +113,58 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+		configOverrides := &clientcmd.ConfigOverrides{}
+		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+		config, err := kubeConfig.ClientConfig()
+		if err != nil {
+			r.logger.Error("Error creating client config for TaskRun %s: %v", name, err)
+		}
+		client, err := taskinterface.NewForConfig(config)
+		if err != nil {
+			r.logger.Error("Error creating client for TaskRun %s: %v", name, err)
+		}
```
wlynch, line 148: These are things that can be set up once for the reconciler, instead of on each reconcile event. We should do this on reconciler creation instead.

results/cmd/watcher/main.go
```diff
@@ -108,12 +113,58 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+	patches := []jsonpatch.JsonPatchOperation{{
+		Operation: "add",
+		Path:      "/metadata/annotations/results_id",
+		Value:     val,
+	}}
+	patchBytes, err := json.Marshal(patches)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal patch bytes in order to cancel: %v", err)
+	}
+	return patchBytes, nil
```
wlynch, line 169:
```suggestion
	return json.Marshal(patches)
```

results/cmd/watcher/main.go
```diff
@@ -108,12 +113,58 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+		}
+
+		// Connect to gke cluster and add results_id to annotations field of TaskRun
+		loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+		configOverrides := &clientcmd.ConfigOverrides{}
+		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+		config, err := kubeConfig.ClientConfig()
+		if err != nil {
+			r.logger.Error("Error creating client config for TaskRun %s: %v", name, err)
+		}
```
wlynch, line 144: For this and other err checks, I think we also need to return the error as well? otherwise this will continue (and likely  panic).

results/cmd/watcher/main.go
```diff
@@ -108,12 +113,58 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+	} else {
+		res, err := r.client.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
+			TaskRun: p,
+		})
+		if err != nil {
+			r.logger.Error("Error creating TaskRun %s: %v", name, err)
+			return err
+		}
+
+		// Connect to gke cluster and add results_id to annotations field of TaskRun
```
dibyom, line 137: Two improvements here:
1. Creating the kubeconfig can be simplified.
2. We should create the client once and pass it in as a field in the reconciler struct

results/cmd/watcher/main.go
dibyom: yeah we need to return an err here.

results/cmd/watcher/main.go
```diff
@@ -108,12 +113,58 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
-	if _, err := r.client.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
-		TaskRun: p,
-	}); err != nil {
-		r.logger.Error("Error updating TaskRun %s: %v", name, err)
-		return err
-	}
 
+	// Create a TaskRun if it does not exist in results server, update existing one otherwise.
+	annotation := p.GetMetadata().GetAnnotations()
+	if val, ok := annotation["results_id"]; ok {
```
dibyom, line 119: we might want to use something more descriptive like `results.tekton.dev/id` as the annotation name.

results/cmd/watcher/main.go
dibyom: Example for 1: 
-  https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go
- https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
```diff
@@ -76,9 +82,11 @@ func main() {
...
 
 type reconciler struct {
-	logger        *zap.SugaredLogger
-	client        pb.ResultsClient
-	taskRunLister listers.TaskRunLister
+	logger            *zap.SugaredLogger
+	client            pb.ResultsClient
+	taskRunLister     listers.TaskRunLister
+	pipelineclientset versioned.Interface
+	fakeserver        *server.Server
```
wlynch, line 89: The server doesn't really need to have a concept of a real vs fake client. If you look at the logic below, they are treated identically. Instead, we should make a client that conforms to the `ResultsClient` interface that's pointed at a local server instance for tests.

e.g. for tests:
```go
s := grpc.NewServer()
pb.RegisterResultsService(s, &Server{...}) // local test server

lis, err := net.Listen("tcp", "localhost:0")
if err != nil {
  ...
}
defer lis.Close()
go s.Serve(lis)
defer s.Stop()
```

results/cmd/watcher/main.go
```diff
@@ -100,20 +108,76 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+		r.logger.Infof("Creating a new TaskRun %s/%s(results_id: %s)", namespace, tr.Name, res.GetResultsId())
+	}
 	return nil
 }
+
+// AnnotationPath creates a jsonpatch path used for adding results_id to TaskRun annotations field.
+func annotationPath(val string) ([]byte, error) {
+	patches := []jsonpatch.JsonPatchOperation{{
+		Operation: "add",
+		Path:      "/metadata/annotations/results.tekton.dev~1id",
```
wlynch, line 175: What does this `~1` syntax mean? Is there a way to dervive it from a common const to share it with the string above?

results/cmd/watcher/main.go
```diff
@@ -100,20 +108,76 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
 	}
-	if _, err := r.client.CreateTaskRunResult(ctx, &pb.CreateTaskRunRequest{
-		TaskRun: p,
-	}); err != nil {
-		r.logger.Error("Error updating TaskRun %s: %v", name, err)
-		return err
-	}
 
+	// Create a TaskRun if it does not exist in results server, update existing one otherwise.
+	annotation := p.GetMetadata().GetAnnotations()
```
wlynch, line 121: nit: We only ever use this once, so feel free to inline below.

results/cmd/watcher/main.go
```diff
@@ -100,20 +108,76 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+				return err
+			}
+		}
+		path, err := annotationPath(res.GetResultsId())
+		if err != nil {
+			r.logger.Error("Error jsonpatch for TaskRun %s : %v", name, err)
+			return err
+		}
+		r.pipelineclientset.TektonV1beta1().TaskRuns(namespace).Patch(name, types.JSONPatchType, path)
+		r.logger.Infof("Creating a new TaskRun %s/%s(results_id: %s)", namespace, tr.Name, res.GetResultsId())
```
wlynch, line 166:
```suggestion
		r.logger.Infof("Creating a new TaskRun result %s/%s(results_id: %s)", namespace, tr.Name, res.GetResultsId())
```

results/cmd/watcher/main.go
wlynch: Doesn't look like this change was applied.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/tektoncd/experimental/results/pkg/server"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+)
+
+var (
+	srv *server.Server
```
wlynch, line 26: Let's avoid declaring global variables, and instead keep server instances local to their tests.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+		fakeserver:        srv,
+		pipelineclientset: pipelineclientset,
+	}
+	impl := controller.NewImpl(c, c.logger, pipeline.PipelineRunControllerName)
+	taskRunInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    impl.Enqueue,
+		UpdateFunc: controller.PassNew(impl.Enqueue),
+	})
+	return impl
+}
```
wlynch, line 66: This looks really similar to the logic in main for setting up the controller. Consider refactoring this into a common function.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	d := []*v1beta1.TaskRun{
+		&taskRun,
+	}
+	// Create a new TaskRun in the fake client, completed TaskRun should be updated with a results id field in annotations.
+	tr, err := getNewTaskRun(taskRun, d, t)
+	if err != nil {
+		t.Fatalf("Failed to get completed TaskRun %s: %v", taskRun.Name, err)
+	}
+	_, ok := tr.Annotations["results.tekton.dev/id"]
+	if !ok {
```
wlynch, line 107:
```suggestion
	if _, ok := tr.Annotations["results.tekton.dev/id"]; !ok {
```

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	impl := controller.NewImpl(c, c.logger, pipeline.PipelineRunControllerName)
+	taskRunInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    impl.Enqueue,
+		UpdateFunc: controller.PassNew(impl.Enqueue),
+	})
+	return impl
+}
+
+// getNewTaskRun sends TaskRun data to reconciler and then retrieves completed TaskRun
+func getNewTaskRun(taskRun v1beta1.TaskRun, d []*v1beta1.TaskRun, t *testing.T) (*v1beta1.TaskRun, error) {
```
wlynch, line 69:
```suggestion
func reconcileTaskRun(taskRun v1beta1.TaskRun, d []*v1beta1.TaskRun, t *testing.T) (*v1beta1.TaskRun, error) {
```

It was unclear when first reading this that calling this function would result in modifications to the task run.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	testAssets, cancel := getFakeClients(t, d)
```
wlynch, line 70: If I'm reading this correctly, this creates new controller/client instances each time this function is called. This feels like it should be done once per setup, and we should simply pass the clients into this func to do the reconcile + Get. What do you think?

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		t.Fatalf("Failed to get completed TaskRun %s: %v", taskRun.Name, err)
+	}
+	_, ok := tr.Annotations["results.tekton.dev/id"]
+	if !ok {
+		t.Fatalf("Expected completed TaskRun %s should be updated with a results_id field in annotations", taskRun.Name)
+	}
+	d = []*v1beta1.TaskRun{
+		tr,
+	}
```
wlynch, line 112: Looks like we're only ever calling this with 1 taskrun. Let's move the list wrapping into the func to make it cleaner.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	return impl
+}
+
+// getNewTaskRun sends TaskRun data to reconciler and then retrieves completed TaskRun
+func getNewTaskRun(taskRun v1beta1.TaskRun, d []*v1beta1.TaskRun, t *testing.T) (*v1beta1.TaskRun, error) {
+	testAssets, cancel := getFakeClients(t, d)
+	defer cancel()
+	c := testAssets.Controller
+	clients := testAssets.Clients
+	if err := c.Reconciler.Reconcile(context.Background(), getRunName(&taskRun)); err != nil {
```
wlynch, line 74: Pass in a context instead of creating a new one.

results/cmd/watcher/results.sql
```diff
@@ -0,0 +1,7 @@
+-- results_id serve as auto-generated key by sqltite to uniquely identify a taskrun log
```
wlynch, line 1: We shouldn't copy the database schema. We should reuse the existing one to ensure that the schemas stay consistent with each other.

results/pkg/server/server.go
```diff
@@ -0,0 +1,272 @@
...
+	mask "go.chromium.org/luci/common/proto/mask"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// Server dsadsad
+type Server struct {
+	pb.UnimplementedResultsServer
+	Env *cel.Env
+	Db  *sql.DB
```
wlynch, line 31: Once you move the tests back to be in this package, you likely won't need to export these since you already have the `New` function below.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+func getRunName(tr *v1beta1.TaskRun) string {
+	return strings.Join([]string{tr.Namespace, tr.Name}, "/")
+}
+
+// getFakeClients create a fake client to send test data to reconciler
+func getFakeClients(t *testing.T, tr []*v1beta1.TaskRun) (test.Assets, context.CancelFunc) {
+	t.Helper()
+	ctx, _ := ttesting.SetupFakeContext(t)
```
wlynch, line 36: I haven't seen this before. What does this give us over a standard context.Background?

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx, cancel := context.WithCancel(ctx)
```
wlynch, line 37: Any particular reason we need the context cancel?

Alternatively, you could use t.Cleanup instead of propagating the defer up the callstack.

results/cmd/watcher/main.go
yuege01: It's just an escape character representing /

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+	clients, _ := test.SeedTestData(t, ctx, d)
+	cmw := configmap.NewInformedWatcher(clients.Kube, "")
+	return test.Assets{
+		Controller: newController(ctx, cmw, t),
+		Clients:    clients,
+	}, cancel
+}
+
+// newController creates a Controller with provided context and configmap
+func newController(ctx context.Context, cmw configmap.Watcher, t *testing.T) *controller.Impl {
```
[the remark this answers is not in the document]
yuege01: It will be used in t.Cleanup(conn.Close()). :)

results/cmd/watcher/main_test.go
yuege01: It seems it is required by creating fakecontext and getting it work

results/cmd/watcher/main_test.go
yuege01: Talked about this in the meeting

results/cmd/watcher/main.go
```diff
@@ -100,20 +108,76 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 		return err
 	}
 
-	r.logger.Infof("Sending update for %s/%s (uid %s)", namespace, tr.Name, tr.UID)
+	r.logger.Infof("Recieving new TaskRun %s/%s", namespace, tr.Name)
```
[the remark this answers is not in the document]
yuege01: done

results/cmd/api/main_test.go
```diff
@@ -41,45 +39,9 @@ func TestCreateTaskRun(t *testing.T) {
...
-	if _, err := statement.Exec(); err != nil {
-		t.Fatalf("failed to execute the taskrun table creation statement statement: %v", err)
-	}
-	return new(db)
-}
-
 func TestGetTaskRun(t *testing.T) {
 	// Create a temporary database
-	srv, err := setupTestDB(t)
+	srv, err := server.SetupTestDB(t)
```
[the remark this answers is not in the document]

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/results.sql
yuege01: done

results/pkg/server/server.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: and this cancel will be used in testReconciler(), btw, I used defer in testReconciler() and cleanup in all helper functions.

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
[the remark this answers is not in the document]
yuege01: I changed definition of newController so it can be shared by both watcher and test, now t is no needed.

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: We talked over meeting that we could use SeedTestData(), but I looked at code, this function() is exactly the one creating client and overwriting old one, plus to use it, we have to create some other configurations. So if we use it, it is basically the same as calling getFakeClients(). I also searched for other code in Tekton repository, for different data in the same test, new client is created for each data set as well.

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+func TestReconcile(t *testing.T) {
+	srv, err := server.SetupTestDB(t)
+	if err != nil {
+		t.Fatalf("Failed to create fake server: %v", err)
+	}
+	s := grpc.NewServer()
+	pb.RegisterResultsServer(s, srv) // local test server
+	lis, err := net.Listen("tcp", port)
+	if err != nil {
+		//
```
dibyom, line 73: t.Fatalf here for tests?

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	defer lis.Close()
```
dibyom, line 75: minor: might be useful to extract the lines that setup the grpc server to its own helper function.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+	defer conn.Close()
+	taskRun := v1beta1.TaskRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "Tekton",
+			Namespace:   "default",
+			Annotations: map[string]string{"demo": "demo"},
+		},
+	}
+	// Create a new TaskRun in the fake client, completed TaskRun should be updated with a results id field in annotations.
+	tr, err := reconcileTaskRun(taskRun, []*v1beta1.TaskRun{&taskRun}, t, conn)
```
dibyom, line 91: I think @wlynch already mentioned this. But we are calling reconcileTaskRun twice in this function. And inside that function, we are initializing the fake clients and the controller each time. Which means that there are two sets of controllers and clients. What we should do instead is call `getFakeClients` once, get the controller from the returned testAssets and then call c.Reconciler.Reconcile twice. In other words, we do the setup only once but call Reconcile twice to verify we are setting the results id annotation.

dibyom: I understand that functionally this won't be too different from what we have now because in the second call to `reconcileTaskRun`, we are seeding the fake client with a Taskrun which already has a results_id set. What this will help with is clarity and readability i.e. when I read the test, it becomes easier for me to understand what is going on. Its like there are two test cases here in the same test:

1. We check that when we reconcile a TaskRun that has no results_id set, we set the results_id. i.e we seed fake clients with the initial taskrun, call reconcile, and then we can get the taskrun to check that it has a results_id annotation.

2. We check that when we reconcile a TaskRun that already has a results_id set, we do not set a new results_id but use the already set results_id.

I think it might be clearer to have two tests (or two subtests in the same test) to make it clearer what is going on.

yuege01: Thanks for the suggestion, for the second one, I tried to use a pre-set results_id, the issue is for UpdateTaskRun(), if it cannot find a taskrun with specified results_id, it will return an error, so the test with a hardcoding TaskRun data will immediately fail. This is the reason why I use a TaskRun created before.

yuege01: I agree with what you said, I think I can create just a new client instead of both client and controller for the two consecutive tests.

yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+	lis, err := net.Listen("tcp", port)
+	if err != nil {
+		t.Fatalf("failed to listen: %v", err)
+	}
+	go s.Serve(lis)
+	t.Cleanup(func() {
+		lis.Close()
+		s.Stop()
+	})
+	return grpc.Dial(*apiAddr, grpc.WithInsecure(), grpc.WithBlock())
```
wlynch, line 75:
```suggestion
	return grpc.Dial(lis.Addr().String(), grpc.WithInsecure(), grpc.WithBlock())
```

We should make sure to use the server created above. As is, this might connect to a different already running server.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func setupServerConnection(t *testing.T) (*grpc.ClientConn, error) {
+	srv, err := server.SetupTestDB(t)
+	if err != nil {
+		t.Fatalf("Failed to create fake server: %v", err)
+	}
+	s := grpc.NewServer()
+	pb.RegisterResultsServer(s, srv) // local test server
+	lis, err := net.Listen("tcp", port)
```
wlynch, line 66:
```suggestion
	lis, err := net.Listen("tcp", ":0")
```

Let the system auto assign a port to avoid conflicts with running servers.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, err
+	}
+	tr, err := clients.Pipeline.TektonV1beta1().TaskRuns(taskRun.Namespace).Get(taskRun.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return tr, nil
+}
+
+func setupServerConnection(t *testing.T) (*grpc.ClientConn, error) {
```
wlynch, line 59: `*grpc.ClientConn` is a bit ambiguous. Let's return a [`ResultClient`](https://pkg.go.dev/github.com/tektoncd/experimental/results@v0.0.0-20200511131837-5e91598ff70d/proto/proto?tab=doc#ResultsClient) instead and pass that around.

wlynch: Also it might make sense to rename this func as `newResultClient` (or similar) instead.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+	srv, err := server.SetupTestDB(t)
```
wlynch, line 60: Thoughts on renaming this func to `NewTestServer`?

results/pkg/server/server.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// create database
+	_ "github.com/mattn/go-sqlite3"
+	pb "github.com/tektoncd/experimental/results/proto/proto"
+	mask "go.chromium.org/luci/common/proto/mask"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// Server dsadsad
```
wlynch, line 27: Replace with actual doc comment.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+func getRunName(tr *v1beta1.TaskRun) string {
+	return strings.Join([]string{tr.Namespace, tr.Name}, "/")
+}
+
+// getFakeClients create a fake client to send test data to reconciler
+func getFakeClients(t *testing.T, tr []*v1beta1.TaskRun, conn *grpc.ClientConn) test.Assets {
+	t.Helper()
+	ctx, _ := ttesting.SetupFakeContext(t)
```
wlynch, line 33: Should we be using the context through the entire test? We might want to pull this out and use this instead of context.Background.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Controller: newController(ctx, cmw, conn),
+		Clients:    clients,
+	}
+}
+
+// reconcileTaskRun sends TaskRun data to reconciler and then retrieves completed TaskRun
+func reconcileTaskRun(ctx context.Context, asset test.Assets, taskRun v1beta1.TaskRun) (*v1beta1.TaskRun, error) {
+	c := asset.Controller
+	clients := asset.Clients
+	if err := c.Reconciler.Reconcile(ctx, getRunName(&taskRun)); err != nil {
```
wlynch, line 49: Let's try and be consistent with our usage of pointers as much as possible. In particular, it looks like for one of the calls to reconcile you're deferencing a pointer to then pass it in here to then pass a pointer reference.

I think some of this would be simplified if we used *v1beta1.TaskRun as the param type instead

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	asset := getFakeClients(t, []*v1beta1.TaskRun{&taskRun}, conn)
+	ctx := context.Background()
+	
+	// Create a new TaskRun in the fake client, completed TaskRun should be updated with a results id field in annotations.
+	tr, err := reconcileTaskRun(ctx, asset, taskRun)
+	if err != nil {
+		t.Fatalf("Failed to get completed TaskRun %s: %v", taskRun.Name, err)
+	}
+	if _, ok := tr.Annotations[idName]; !ok {
```
wlynch, line 101: We should also test that the corresponding result was created.

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	// Update a TaskRun with results id, completed TaskRun should remain unchanged.
+	newtr, err := reconcileTaskRun(ctx, asset, *tr)
+	if err != nil {
+		t.Fatalf("Failed to get completed TaskRun %s: %v", taskRun.Name, err)
+	}
+	if diff := cmp.Diff(tr, newtr); diff != "" {
+		t.Fatalf("Expected completed TaskRun should remain unchanged when it has a results_id in annotations: %v", diff)
+	}
+}
```
wlynch, line 112: This does a good job of testing the new build flow, but we should also add a test for the update flow as well.

yuege01: Added a update-flow test

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/pkg/server/server.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
yuege01: done

results/cmd/watcher/main_test.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		res, err := client.GetTaskRunResult(ctx, &pb.GetTaskRunRequest{ResultsId: tr.Annotations[idName]})
+		if err != nil {
+			t.Fatalf("Expected completed TaskRun %s not created in api server", taskRun.Name)
+		}
+		pr, err := convert.ToProto(updatetr)
+		if err != nil {
+			t.Fatalf("failed to convert to proto: %v", err)
+		}
+		if diff := cmp.Diff(pr.String(), res.GetTaskRun().String()); diff != "" {
```
wlynch, line 150: We should compare the structs directly, not the strings. Theres a chance that the string function may not contain all the information needed to check for equality.
```suggestion
		if diff := cmp.Diff(pr, res.GetTaskRun()); diff != "" {
```

yuege01: got it, by the way, to compare unexported struct like proto, like in test for api, we have to use ``diff := cmp.Diff(pr, res.GetTaskRun(), protocmp.Transform())``

yuege01: done

PR #615, opened 2020-07-31
Change return type of ListTaskRunsResult

yuege01:
Change return type of ListTaskRunsResult to a list of TaskRunResults instead of TaskRuns

Modified api.proto for ListTaskRunsResponse

# Submitter Checklist

review them:

for more details._


yuege01:
@wlynch @dibyom

results/cmd/api/main_test.go
```diff
@@ -555,7 +555,10 @@ func TestListTaskRuns(t *testing.T) {
 			if tc.expectStatus != status.Code(err) {
 				t.Fatalf("failed on test %v: %v", tc.name, err)
 			}
-			gotList := res.GetItems()
+			var gotList []*pb.TaskRun
+			for _, taskrunResult := range res.GetItems() {
+				gotList = append(gotList, taskrunResult.GetTaskRun())
+			}
```
wlynch, line 561: Only looking at the taskruns means that we are ignoring the extra data being returned by TaskRunResults. We should modify the expected results to include the TaskRunResults so we can verify that it's being set properly.

We'll need a way to optionally make ID values deterministic for testing (easiest thing to do is likely just making the ID generator func configurable by the package, and setting it to something deterministic in tests).

results/cmd/api/main.go
```diff
@@ -65,6 +65,7 @@ type server struct {
 	env *cel.Env
 	db  *sql.DB
 }
+type newID func() string
```
wlynch, line 68: Doesn't look like this is being used. Remove.

PR #618, opened 2020-08-03
Fix resource modified time

sthaha:
# Changes

Previously, API sync service used to use update_at time to infer the
date a resource was last modified. The sync service also used to parse
the entire catalog if requested even if it was already parsed and
updated in DB.

This patch fixes it by exiting early if the SHA/HEAD of the catalog last
scanned is the same. It also adds a modified_at field to compute and
store the last modified time of the resource instead of relying on the
updated_at so that even if the catalog gets updated, the modified_at
still remains the same.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

# Submitter Checklist

review them:

- [🙅  ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/hub/api/pkg/git/repo.go
+++ b/hub/api/pkg/git/repo.go
@@ -7,9 +7,9 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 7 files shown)


hub/api/pkg/git/repo.go
```diff
@@ -200,16 +214,37 @@ func (r Repo) parseResourceVersion(filePath string, kind string) (*TekonResource
...
 		Description: description,
 		Path:        relPath,
+		ModifiedAt:  modified,
 	}
 	return ret, nil
 }
 
+func (r *Repo) lastModifiedTime(path string) (time.Time, error) {
+	gitPath, _ := filepath.Rel(r.Path, path)
+	//r.Log.Infof(">>> finding %s modified for repo at %s", gitPath, r.Path)
```
PuneetPunamiya, line 236: Are these comments necessary, if not can we remove it ??

PR #621, opened 2020-08-05
Hub: Fixes trailing spaces in api

PuneetPunamiya:
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/hub/api/pkg/git/repo.go
+++ b/hub/api/pkg/git/repo.go
@@ -221,10 +221,10 @@ func (r Repo) parseResourceVersion(filePath string, kind string) (*TekonResource
 	}
 
 	ret := &TekonResourceVersion{
-		Version:     version,
-		DisplayName: displayName,
-		Tags:        strings.Split(tags, ","),
-		Description: description,
+		Version:     strings.TrimSpace(version),
+		DisplayName: strings.TrimSpace(displayName),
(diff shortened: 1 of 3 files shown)


hub/api/pkg/sync/sync.go
```diff
@@ -170,7 +170,7 @@ func (s *SyncService) updateResources(job model.SyncJob, repo git.Repo, res []gi
 			for _, t := range v.Tags {
 				tag := model.Tag{Name: t, CategoryID: others.ID}
```
SM43, line 171: here also spaces needs to be trimmed

SM43: `tag` is getting inserted so it should be trimmed

hub/api/pkg/sync/sync.go
```diff
@@ -170,7 +170,7 @@ func (s *SyncService) updateResources(job model.SyncJob, repo git.Repo, res []gi
...
 
-				txn.Model(&model.Tag{}).Where(&model.Tag{Name: t}).FirstOrCreate(&tag)
+				txn.Model(&model.Tag{}).Where(&model.Tag{Name: strings.TrimSpace(t)}).FirstOrCreate(&tag)
```
sthaha, line 173: this doesn't still address the actual cause, does it? the v.Tags contain spaces. I think the Trim must be applied to the parsing logic.

PR #629, opened 2020-09-14
Hub: use sha256 instead of tags

sthaha:
Previously Hub db image used postgres:13 which used to get updated
frequently and caused db to fail with the following error

```
PostgreSQL Database directory appears to contain a database; Skipping initialization

[1] FATAL:   database files are incompatible with server
[1] DETAIL:  The database cluster was initialized with CATALOG_VERSION_NO 202005171, but the server was compiled with CATALOG_VERSION_NO 202007201.
[1] HINT:    It looks like you need to initdb.
[1] LOG:     database system is shut down

```
This used to happen because the docker image used to get updated
frequently and overwrote the same tag.

Hopefully pinning down the image using its SHA should fix this 🤞.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/hub/api/config/12-db-deployment.yaml
+++ b/hub/api/config/12-db-deployment.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
         - name: db
-          image: postgres:13
+          image: postgres:13@sha256:260a98d976574b439712c35914fdcb840755233f79f3e27ea632543f78b7a21e
           imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 5432


SM43:
If the image is updated would it fetch a new one? the DB deployment has `imagePullPolicy: IfNotPresent`.

sthaha:
> If the image is updated would it fetch a new one? the DB deployment has `imagePullPolicy: IfNotPresent`.

Evidently it did for some reason because the db crash coincided with the update to the tag (3 days back) and the (re)deployment showed that the db has been deployed 3 days back.

vdemeester (approved):

PuneetPunamiya:
Thanks 
/lgtm

PR #634, opened 2020-09-22
Results: Configure API Server and Watcher to run against remote servers.

wlynch:
# Changes

Largely a rehash of the work Jason did in #539, but focuses on any
remote DB, not just one running locally in the cluster.

- Configures API Server to run against an remote MySQL database.
  - Primarily uses environment variables instead of flags for
    configuration, since this just happened to be easier. I'm not opposed
    to moving this back to flags at a later point in time.
- Configures Result Watcher to dial into remote API Server using TLS +
  authentication credentials.
- Slightly modifies storage schema to make it MySQL friendly (still
  uses sqlite for local unit tests)

With these, I was able to get a working e2e demo running, with the
API Server running on Cloud Run + Cloud SQL and the watcher running
locally.

# Submitter Checklist

review them:

for more details._


results/pkg/grpc/creds.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"google.golang.org/api/idtoken"
+	"google.golang.org/grpc/credentials"
+)
+
+// Google provides an authenticated transport for use against Google APIs
```
ImJasonH, line 11: I think we need to do this in a provider-agnostic way from the beginning, or we risk having to refactor all of this before we can promote it to a real top-level project. I'd rather do this without creds, insecurely, than bake in GCP support like this.

Let's keep this code stashed somewhere for later, though, since I think we can use it when the general solution exists.

results/cmd/api/main.go
```diff
@@ -18,38 +18,53 @@ package main
...
+	}
+	// Connect to the MySQL database.
+	// DSN derived from https://github.com/go-sql-driver/mysql#dsn-data-source-name
+	dbURI := fmt.Sprintf("%s:%s@%s(%s)/%s?parseTime=true", user, pass, os.Getenv("DB_PROTOCOL"), os.Getenv("DB_ADDR"), os.Getenv("DB_NAME"))
+	db, err := sql.Open("mysql", dbURI)
 	if err != nil {
 		log.Fatalf("failed to open the results.db: %v", err)
 	}
 	defer db.Close()
+
```
ImJasonH, line 49: I have no idea where `// Create cel environment for filter` comes from and at this point I'm afraid to ask. 😆

ImJasonH (commented):
/lgtm
/approve

PR #636, opened 2020-10-06
Initial implementation of task looping controller

GregDritschler:
The Task Loop controller allows users to run a Task in a loop with varying parameter values.
This functionality is provided by a controller that implements the Tekton Custom Task interface.

The initial implementation has the following limitations.

* The value of only one Task parameter can be varied between TaskRun executions.

* Each TaskRun is executed sequentially.  The second TaskRun is created only after the first TaskRun completes, and so on.

* If a TaskRun fails, the execution of the TaskLoop stops.  TaskRuns for remaining iteration values are not created.

* Task results are not collected into Run results (run.status.results).
  However the results of each TaskRun can be seen in the TaskRun status under run.status.extraFields.

* Run does not support specifying workspaces, pipeline resources, a service account name or a pod template.

* There are no metrics specific to Run.

# Changes

# Submitter Checklist

review them:

for more details._


GregDritschler:
@afrittoli @bobcatfish @ImJasonH

The install/uninstall sections of the readme will be filled in once we figure that part out.

I plan to work on enabling parallel execution next since many people have mentioned that but I want to get this big hunk of stuff in first.

ImJasonH:
This is great! 🎉 

/lgtm

afrittoli (commented):
Thanks!!
/lgtm

PR #641, opened 2020-10-08
Fix unit tests

GregDritschler:
taskloop_validation_test.go fails due to a recent change in task validation.
It can no longer assume that only one FieldError is returned.
It needs to use Error() to collapse the FieldErrors into a string.

tasklooprun_test.go fails with compilation errors.
MarkRunning was renamed MarkRunRunning.
It also needs to use the experimental test package rather than the pipelines one.

These errors got through because we don't yet have automation set up for this repo.

# Changes

Fix taskloop_validation_test.go to use Error().

Fix tasklooprun_test.go to use correct package and fix the incorrect function name.

# Submitter Checklist

review them:

for more details._


afrittoli (commented):
/lgtm
/approve

PR #647, opened 2020-10-16
Adds PipelineRun message

XinruZhang:
# Changes
1. Creates pipelinerun.proto file to store PipelineRun message
2. Split out ObjectMeta and Condition messages to common.proto file
   to avoid repeat definitions

# Submitter Checklist

review them:

for more details._


results/proto/pipelinerun.proto
```diff
@@ -0,0 +1,77 @@
...
+import "common.proto";
+import "taskrun.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "./proto";
+
+message PipelineRun {
+    string api_version = 1;
+	string kind = 2; // TODO: must always be TaskRun?
```
wlynch, line 14: Fix indentation (here and other places in the file)

wlynch, line 14: Can probably remove the TODO (or switch to PipelineRun to reflect the right type)

results/proto/pipelinerun.proto
```diff
@@ -0,0 +1,77 @@
...
+
+	ObjectMeta metadata = 3;
+
+	PipelineRunSpec spec = 4;
+	PipelineRunStatus status = 5;
+}
+
+message PipelineRunSpec {
+    PipelineSpec task_spec = 5;
```
wlynch, line 23: Any reason why these proto numbers don't start at 1?

results/proto/pipelinerun.proto
```diff
@@ -0,0 +1,77 @@
...
+	google.protobuf.Timestamp start_time = 4;
+    google.protobuf.Timestamp completion_time = 5;
+    
+    // The type of this field in go struct is map[string] *PipelineRunTaskRunStatus
+    map<string, PipelineRunTaskRunStatus> taskruns = 6;
+    PipelineSpec pipeline_spec = 7;
+}
+
+message PipelineRunTaskRunStatus {
+    string pipeline_task_nam = 1;
```
wlynch, line 75:
```suggestion
    string pipeline_task_name = 1;
```

We should probably add some sort of test to make sure this converts correctly. With TaskRuns there were a few cases where the Go struct name was not in the default format protobuf expected, causing issues when converting.

wlynch (changes requested):

results/pkg/watcher/reconciler/controller_test.go
```diff
@@ -132,6 +132,7 @@ func TestReconcile(t *testing.T) {
 		tr.UID = "234435"
 		asset.Clients.Pipeline.TektonV1beta1().TaskRuns(taskRun.Namespace).Update(tr)
 		updatetr, err := reconcileTaskRun(ctx, asset, tr)
+		updatetr.ResourceVersion = tr.ResourceVersion
```
wlynch, line 135: This seems unrelated to the rest of the proto changes. Why is this necessary?

results/proto/pipelinerun.proto
```diff
@@ -0,0 +1,82 @@
...
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "./proto";
+
+message PipelineRun {
+  string api_version = 1;
+  string kind = 2;
+
+	ObjectMeta metadata = 3;
```
wlynch, line 16: Fix indenting.

wlynch (changes requested):
Mostly lgtm! Few minor comments.

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: That's because the new version of k8s will change objects' ResourceVersion after they are updated, so we need to equal them manually before comparing them. See [concurrency-control-and-consistency](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency). Should we also make sure the ResourceVersion field is changed?

wlynch (approved):
/lgtm

PR #649, opened 2020-10-20
Update presubmit-tests scripts to include task-loops

GregDritschler:
# Changes

A presubmit-tests.sh script is added to task-loops and the main presubmit-tests.sh script is updated to call it.

# Submitter Checklist

review them:

for more details._


GregDritschler:
All tests are passing for task-loops including the e2e (integration) tests.

I've changed the main `presubmit-tests` script in experimental to run a sub-project's tests only if
* files were changed inside that sub-project
* files were changed outside all sub-projects (which could have a global effect)

Since this PR is changing the main `presubmit-tests` script which is outside all of the sub-projects, it is running the tests in all of the subprojects as expected.

The test logs indicate why a subproject's test is being run.  For example `webhooks-extension` is tested because of the change to a file outside the sub-projects.

```
============================
==== WEBHOOKS-EXTENSION ====
============================
test/presubmit-tests.sh is modified so testing all projects
```

On the other hand `task-loops` is tested because of the change to a file inside that sub-project.

```
====================
==== TASK-LOOPS ====
====================
task-loops/test/e2e-common.sh is modified so testing project task-loops
```

@afrittoli Please review when you have a chance.

test/presubmit-tests.sh
```diff
@@ -26,20 +26,41 @@ source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/presubmit-te
...
+        # If changed file is outside all project folders then test every folder.
+        echo $file | grep -q -v $inanyprojectregex && echo "$file is modified so testing all projects" && return 0
+    done
+    return 1
+}
+
+# Get list of changed files
+initialize_environment
+
+projects="webhooks-extension catalogs octant-plugin tekdoc results generators task-loops"
```
afrittoli, line 61: I think the complete list here should be `catalogs commit-status-tracker generators helm hub oci octant-plugin pipeline results task-loops tekdoc tekton-listener webhooks-extension`

afrittoli (commented):
Thanks for this!

In my mind I was expecting to have different CI jobs, each which would only run if the corresponding folder was changed. But this setup would be more complex to achieve and maintain, so I think this is a better solution for the issue.

I noticed that the webhook unit tests were executed for some reason, but apart from that everything else seemed to be working ok. I'd be happy to merge this and address remaining issues in follow-up patches.

/cc @vdemeester 
/approve

We need to mention this in the WG so that people is aware of the change - so that it does not come to a surprise to folks contributing to `experimental`.

test/presubmit-tests.sh
GregDritschler: I used the list of projects in the existing presubmit-tests script.  The other subprojects not in the list don't have a presubmit-tests script so nothing would happen for them.

GregDritschler:
> We need to mention this in the WG so that people is aware of the change - so that it does not come to a surprise to folks contributing to experimental.

Which WG?

test/presubmit-tests.sh
GregDritschler: Oh wait, I see.  Even though those projects don't have tests, they need to be in the list so that changes to those projects aren't considered "global" which would cause every project's tests to run.  I've updated the list.

PR #650, opened 2020-10-23
Update the OCI experimental tool to match the current Tekton bundle f…

pierretasci:
# Changes

Updates the OCI tool to use the new Tekton Bundle format. This allows users to test out the functionality with this tool before an official Tekton Bundle tool is built.

# Submitter Checklist

review them:


oci/go.mod
```diff
@@ -13,14 +13,12 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.4.0 // indirect
 	github.com/spf13/cobra v0.0.5
-	github.com/tektoncd/pipeline v0.10.1
+	github.com/tektoncd/pipeline v0.11.2
```
vdemeester, line 16: Any reason for it to be `v0.11.2` instead of a later version ?

vdemeester (approved):

oci/go.mod
pierretasci: That's the version when I first wrote this modification to the experimental tool  👀

vdemeester: ok, let's get this in anyway, we can bump this in a follow-up :wink:

PR #651, opened 2020-10-23
Reconciles PipelineRun for Results API

XinruZhang:
# Changes
1. Changes convert function name:  ToProto -> ToTaskRunProto
2. Reconciles PipelineRun
3. Adds related test

# Submitter Checklist

review them:

for more details._


XinruZhang:
This PR is based on #647

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,60 +59,94 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
-	tr, err := r.pipelineclientset.TektonV1beta1().TaskRuns(namespace).Get(name, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		// The resource no longer exists, in which case we stop processing.
-		r.logger.Infof("task run %q in work queue no longer exists", key)
-		return nil
-	} else if err != nil {
-		r.logger.Errorf("Error retrieving Result %q: %s", name, err)
+	// Get the resource object with this namespace/name
+	taskRun, terr := r.pipelineclientset.TektonV1beta1().TaskRuns(namespace).Get(name, metav1.GetOptions{})
+	pipelineRun, perr := r.pipelineclientset.TektonV1beta1().PipelineRuns(namespace).Get(name, metav1.GetOptions{})
```
wlynch, line 67: I don't think this works as expected, because the reconcile func is tied to the TaskRun informer - e.g. we only ever get notified of TaskRun changes, and not PipelineRun. This works in your test because we're normalizing the object names to `namespace/id` which we then attempt on both clients - this will break down in cases where the TaskRun name == PipelineRun name.

We should break this up into 2 funcs or structs and handle pipeline runs separately from task runs, and group similar behavior into shared funcs where appropriate - this should also reduce the various type switching we're doing.

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,60 +59,94 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+		r.logger.Errorf("Error retrieving TaskRun %q: %s", name, terr)
+		return terr
+	}
+	if errors.IsNotFound(terr) && perr != nil {
+		r.logger.Errorf("Error retrieving PipelineRun %q: %s", name, perr)
+		return perr
+	}
 
-	r.logger.Infof("Receiving new Result %s/%s", namespace, tr.Name)
+	var result *pb.Result = nil
```
wlynch, line 82:
```suggestion
	var result *pb.Result
```

Pointers default to nil, so you can omit the trailing `= nil`

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,60 +59,94 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
 
-	// Send the new status of the Result to the API server.
-	p, err := convert.ToProto(tr)
-	if err != nil {
-		r.logger.Errorf("Error converting to proto: %v", err)
-		return err
+	val := ""
```
wlynch, line 84: val is a bit too generic here. Is there a more descriptive name we can use?

resultID?

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,60 +59,94 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
-		res.Name = val
+	if ok {
+		result.Name = val
 		if _, err := r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
 			Name:   val,
-			Result: res,
+			Result: result,
 		}); err != nil {
-			r.logger.Errorf("Error updating TaskRun %s: %v", name, err)
+			r.logger.Errorf("Error updating Tekton Result(%s): %s\n%v", resultType, val, err)
```
wlynch, line 126: We can use zap to automatically tag all logs with theses values - https://pkg.go.dev/go.uber.org/zap#Logger.With e.g.

```go
log := r.logger.With(zap.String("type", "PipelineRun"), zap.String("result", val))
log.Errorf(...)
```

wlynch (changes requested):

results/pkg/watcher/reconciler/controller_test.go
```diff
@@ -57,6 +55,20 @@ func reconcileTaskRun(ctx context.Context, asset test.Assets, taskRun *v1beta1.T
 	return tr, nil
 }
 
+// reconcileTaskRun sends TaskRun data to reconciler and then retrieves completed TaskRun
+func reconcilePipelineRun(ctx context.Context, asset test.Assets, pipelineRun *v1beta1.PipelineRun) (*v1beta1.PipelineRun, error) {
+	c := asset.Controller
+	clients := asset.Clients
+	if err := c.Reconciler.Reconcile(ctx, strings.Join([]string{pipelineRun.Namespace, pipelineRun.Name}, "/")); err != nil {
```
wlynch, line 62: Discovered this trick recently: https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineRun.GetNamespacedName

```suggestion
	if err := c.Reconciler.Reconcile(ctx, pipelineRun.GetNamespacedName()); err != nil {
```

results/pkg/watcher/reconciler/controller_test.go
```diff
@@ -81,35 +93,76 @@ func newResultsClient(t *testing.T) pb.ResultsClient {
...
+		if _, ok := tr.Annotations[IDName]; !ok {
 			t.Fatalf("Expected completed TaskRun %s should be updated with a results_id field in annotations", taskRun.Name)
 		}
-		if _, err := client.GetResult(ctx, &pb.GetResultRequest{Name: tr.Annotations[idName]}); err != nil {
+		if _, err := client.GetResult(ctx, &pb.GetResultRequest{Name: tr.Annotations[IDName]}); err != nil {
 			t.Fatalf("Expected completed TaskRun %s not created in api server", taskRun.Name)
 		}
 	})
 
+	t.Run("test-create-pipelinerun", func(t *testing.T) {
```
wlynch, line 151: many of these tests look very similar. Could we use a table driven test here? https://github.com/golang/go/wiki/TableDrivenTests

results/pkg/watcher/reconciler/controller_test.go
```diff
@@ -157,4 +225,39 @@ func TestReconcile(t *testing.T) {
...
+			Name: pr.Annotations[IDName],
+			Executions: []*pb.Execution{{
+				Execution: &pb.Execution_PipelineRun{p},
+			}},
+		}
+		if diff := cmp.Diff(want, res, protocmp.Transform()); diff != "" {
+			t.Fatalf("Expected completed PipelineRun should be upated in api server: %v", diff)
+		}
+	})
 }
```
wlynch, line 263: Another test that would be useful is an additive Result - e.g. have a PipelineRun, then add a TaskRun in the same result. As a bonus, let's have them share the same name so we can make sure to cover that edge case as well.

results/pkg/watcher/reconciler/controller_test.go
```diff
@@ -157,4 +225,39 @@ func TestReconcile(t *testing.T) {
...
 	})
+
+	// Update a PipelineRun with new value and results id, completed PipelineRun should be updated.
+	t.Run("test-update-pipelinerun", func(t *testing.T) {
+		pr, err := reconcilePipelineRun(ctx, asset, pipelineRun)
+		if err != nil {
+			t.Fatalf("Failed to get completed PipelineRun %s: %v", pipelineRun.Name, err)
+		}
+		pr.UID = "234435"
+		asset.Clients.Pipeline.TektonV1beta1().PipelineRuns(pipelineRun.Namespace).Update(pr)
```
wlynch, line 236: Check the error.

wlynch (changes requested):

results/pkg/watcher/reconciler/controller.go
```diff
@@ -3,54 +3,62 @@ package reconciler
...
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
-	_ "knative.dev/pkg/system/testing"
 )
 
 const (
-	path   = "/metadata/annotations/results.tekton.dev~1id"
-	idName = "results.tekton.dev/id"
+	Path   = "/metadata/annotations/results.tekton.dev~1id"
```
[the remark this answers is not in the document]
XinruZhang: Ok!

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,70 +67,152 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		r.logger.Errorf("invalid resource key: %s", key)
-		return nil
+		return err
+	}
+
+	pEMsg, prErr := r.reconcilePipelineRun(ctx, namespace, name)
+	tEMsg, trErr := r.reconcileTaskRun(ctx, namespace, name)
```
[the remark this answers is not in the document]
XinruZhang: Thanks for the advice! That's quite helpful! I've separated them in the latest commit.

I was very swinging on that. On the one hand, the two separate reconcilers have almost the same code, which is the reason I choose not to separate them. However, on the other hand, two reconcilers can bring much more flexibility, and the codes are also easier to read.

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,70 +67,152 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+	return false
+}
+
+func getReconcileError(emsgs []string, errs ...error) (string, error) {
+	if isReconcileSuccess(errs...) {
+		return "", nil
+	}
+	for idx, err := range errs {
+		if !errors.IsNotFound(err) {
+			return emsgs[idx], err
```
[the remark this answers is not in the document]
XinruZhang: Got it!

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,70 +67,152 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+		r.logger.With(zap.String("namespace", namespace), zap.String("name", name)).Error(emsg)
 	}
+	return err
+}
+
+func isReconcileSuccess(errs ...error) bool {
+	for _, err := range errs {
+		if err == nil {
+			return true
+		}
```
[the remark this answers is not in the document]
XinruZhang: Ok, didn't consider that. Since the reconciler is separated now, I guess we can abandon this piece of code.

results/pkg/watcher/reconciler/controller.go
```diff
@@ -3,54 +3,62 @@ package reconciler
...
+		UpdateFunc: controller.PassNew(impl.Enqueue),
+	})
 	return impl
 }
 
 type reconciler struct {
 	logger            *zap.SugaredLogger
 	client            pb.ResultsClient
 	taskRunLister     listers.TaskRunLister
+	pipelineRunLister listers.PipelineRunLister
```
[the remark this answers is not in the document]
XinruZhang: Not used, ok!

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,70 +67,152 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
 
 	// Get the Task Run resource with this namespace/name
-	tr, err := r.pipelineclientset.TektonV1beta1().TaskRuns(namespace).Get(name, metav1.GetOptions{})
+	taskRun, err := r.pipelineclientset.TektonV1beta1().TaskRuns(namespace).Get(name, metav1.GetOptions{})
+
 	if errors.IsNotFound(err) {
 		// The resource no longer exists, in which case we stop processing.
-		r.logger.Infof("task run %q in work queue no longer exists", key)
-		return nil
+		return fmt.Sprintf("TaskRun %q/%q in work queue no longer exists", namespace, name), err
```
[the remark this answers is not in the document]
XinruZhang: Great! Thanks!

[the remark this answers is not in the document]
XinruZhang: Yeah, you are right

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: Got it!

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: No problem!

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: I have a question here. If we separate the reconcilers, is there any chance we would meet this corner case?

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: Got it

results/pkg/watcher/reconciler/controller.go
```diff
@@ -59,70 +67,152 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
...
+func getReconcileError(emsgs []string, errs ...error) (string, error) {
+	if isReconcileSuccess(errs...) {
+		return "", nil
+	}
+	for idx, err := range errs {
+		if !errors.IsNotFound(err) {
+			return emsgs[idx], err
+		}
+	}
+	return "Can't find valid Tekton Result resource", errors.NewNotFound(v1beta1.Resource("Tekton Result"), "Tekton Result")
```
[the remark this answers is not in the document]
XinruZhang: Great~ Thanks for such detailed comments! I learned a lot!
I'm gonna resolve this conversation because we're going to separate the reconcilers, so this misleading error message will no longer exist :)

results/pkg/watcher/reconciler/controller_test.go
wlynch: We wouldn't have to worry about the naming corner case, but we should make sure to test the additive behavior (e.g. TaskRuns and PipelineRuns are able to be added to the same Result)

wlynch (changes requested):
Looking good! Few more final nits, mostly around refactoring code to reuse across packages.

results/pkg/watcher/reconciler/taskrun/taskrun.go
```diff
@@ -0,0 +1,103 @@
...
+		if _, err = r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
+			Name:   resultID,
+			Result: trResult,
+		}); err != nil {
+			logger.Errorf("Error updating TaskRun: %v", err)
+			return err
+		}
+		logger.Infof("Updating TaskRun, result id: %s", resultID)
+	} else {
+		if trResult, err = r.client.CreateResult(ctx, &pb.CreateResultRequest{
```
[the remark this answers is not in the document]
XinruZhang: Got

results/pkg/api/server/server.go
```diff
@@ -252,7 +255,9 @@ func SetupTestDB(t *testing.T) (*Server, error) {
 	t.Cleanup(func() {
 		db.Close()
 	})
-	schema, err := ioutil.ReadFile("../../../schema/results.sql")
+	_, b, _, _ := runtime.Caller(0)
+	basepath := filepath.Dir(b)
```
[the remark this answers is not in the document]
XinruZhang: The relative path is resolved dynamically in runtime. In this case, we called this function `SetupTestDB` in taskrun_test.go and pipelinerun_test.go, and they are one layer deeper than the server.go. Therefore the program can't find the file based on the relative path.

results/pkg/watcher/reconciler/controller_test.go
XinruZhang: Oh! That's a great Test Case!

I didn't get that you mean we should be able to hold both PipelineRun and TaskRun in the same result, I thought you want to check whether it would work if they share the same name.

results/pkg/watcher/reconciler/taskrun/taskrun.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	logger.Info("Recieving new TaskRun")
+	trProto, err := convert.ToTaskRunProto(tr)
+	if err != nil {
+		logger.Errorf("Error converting TaskRun to its corresponding proto: %v", err)
+		return err
+	}
+
+	if resultID, ok := trProto.GetMetadata().GetAnnotations()[common.IDName]; ok {
```
wlynch, line 53: This looks highly duplicated with pipelinerun, when all that differs seems to be the Execution. We can abstract and combine the duplicative logic?

results/pkg/watcher/reconciler/common/common.go
```diff
@@ -0,0 +1,86 @@
...
+const (
+	port = ":0"
+)
+
+const (
+	Path   = "/metadata/annotations/results.tekton.dev~1id"
+	IDName = "results.tekton.dev/id"
+)
+
+func NewResultsClient(t *testing.T) pb.ResultsClient {
```
wlynch, line 27: Pull test-only packages into an internal package(s) - https://golang.org/doc/go1.4#internalpackages

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -0,0 +1,137 @@
...
+	if _, ok := pr.Annotations[common.IDName]; !ok {
+		t.Fatalf("Expected completed PipelineRun %s should be updated with a results_id field in annotations", tt.pipelineRun.Name)
+	}
+	if _, err := tt.client.GetResult(tt.ctx, &pb.GetResultRequest{Name: pr.Annotations[common.IDName]}); err != nil {
+		t.Fatalf("Expected completed PipelineRun %s not created in api server", tt.pipelineRun.Name)
+	}
+}
+
+func (tt *PipelineRunTest) testUnchangePipelineRun(t *testing.T) {
+	pr, err := tt.asset.Clients.Pipeline.TektonV1beta1().PipelineRuns(tt.pipelineRun.Namespace).Get(tt.pipelineRun.Name, metav1.GetOptions{})
```
wlynch, line 74: What's the difference between pr and tt.pipelineRun?

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -0,0 +1,137 @@
...
+func (tt *PipelineRunTest) testUpdatePipelineRun(t *testing.T) {
+	pr, err := common.ReconcilePipelineRun(tt.ctx, tt.asset, tt.pipelineRun)
+	if err != nil {
+		t.Fatalf("Failed to get completed PipelineRun %s: %v", tt.pipelineRun.Name, err)
+	}
+	pr.UID = "234435"
+	if _, err = tt.asset.Clients.Pipeline.TektonV1beta1().PipelineRuns(tt.pipelineRun.Namespace).Update(pr); err != nil {
+		t.Fatalf("Failed to update PipelineRun %s: %v", tt.pipelineRun.Name, err)
+	}
+	updatepr, err := common.ReconcilePipelineRun(tt.ctx, tt.asset, pr)
```
wlynch, line 96: Is there a purpose to the double Reconcile here, or is 1 sufficient?

results/pkg/watcher/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -0,0 +1,99 @@
...
+			remotePr := execution.GetPipelineRun()
+			if remotePr != nil && remotePr.Metadata.Namespace == pr.Namespace && remotePr.Metadata.Name == pr.Name {
+				found = true
+				result.Executions[idx] = &pb.Execution{Execution: &pb.Execution_PipelineRun{prProto}}
+			}
+		}
+		if !found {
+			result.Executions = append(result.Executions, &pb.Execution{Execution: &pb.Execution_PipelineRun{prProto}})
+		}
+		if _, err = r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
```
wlynch, line 69:
```suggestion
		if _, err := r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
```

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func (tt *PipelineRunTest) testUpdatePipelineRun(t *testing.T) {
+	pr, err := common.ReconcilePipelineRun(tt.ctx, tt.asset, tt.pipelineRun)
+	if err != nil {
+		t.Fatalf("Failed to get completed PipelineRun %s: %v", tt.pipelineRun.Name, err)
+	}
+	pr.UID = "234435"
+	if _, err = tt.asset.Clients.Pipeline.TektonV1beta1().PipelineRuns(tt.pipelineRun.Namespace).Update(pr); err != nil {
```
wlynch, line 93:
```suggestion
	if _, err := tt.asset.Clients.Pipeline.TektonV1beta1().PipelineRuns(tt.pipelineRun.Namespace).Update(pr); err != nil {
```

results/pkg/watcher/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+}
+
+func getFakeClients(t *testing.T, tr []*v1beta1.TaskRun, pr []*v1beta1.PipelineRun, client pb.ResultsClient) (test.Assets, test.Assets, context.Context) {
+	t.Helper()
+	ctx, _ := ttesting.SetupFakeContext(t)
+	d := test.Data{
+		TaskRuns:     tr,
+		PipelineRuns: pr,
+	}
```
wlynch, line 81: We have 3 similar getFakeClients, and AFAICT they primarily differ by the `test.Data`. Can we replace the tr/pr params with `test.Data` and have a shared client creation?

results/pkg/watcher/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Controller: taskrun.NewController(ctx, cmw, client),
+			Clients:    clients,
+		}, test.Assets{
+			Controller: pipelinerun.NewController(ctx, cmw, client),
+			Clients:    clients,
+		}, ctx
+}
+
+func (tt *ReconcilerTest) testUpdatePipelineRunToTheExistedResult(t *testing.T) {
+	tt = newReconcilerTest(t)
```
wlynch, line 95: It seems strange to self-update this value. Perhaps we should do something like

```suggestion
func TestUpdatePipelineRunToTheExistedResult(t *testing.T) {
	tt := newReconcilerTest(t)
```

results/pkg/watcher/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			logger.Errorf("Error creating PipelineRun Result: %v", err)
+			return err
+		}
+		path, err := common.AnnotationPath(prResult.GetName(), common.Path, "add")
+		if err != nil {
+			logger.Errorf("Error jsonpatch for PipelineRun Result %s: %v", prResult.GetName(), err)
+			return err
+		}
+		r.pipelineclientset.TektonV1beta1().PipelineRuns(pr.Namespace).Patch(pr.Name, types.JSONPatchType, path)
```
wlynch, line 94: Check the error

results/pkg/watcher/reconciler/taskrun/taskrun.go
```diff
@@ -0,0 +1,99 @@
...
+			remoteTr := execution.GetTaskRun()
+			if remoteTr != nil && remoteTr.Metadata.Namespace == tr.Namespace && remoteTr.Metadata.Name == tr.Name {
+				found = true
+				result.Executions[idx] = &pb.Execution{Execution: &pb.Execution_TaskRun{trProto}}
+			}
+		}
+		if !found {
+			result.Executions = append(result.Executions, &pb.Execution{Execution: &pb.Execution_TaskRun{trProto}})
+		}
+		if _, err = r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
```
wlynch, line 69:
```suggestion
		if _, err := r.client.UpdateResult(ctx, &pb.UpdateResultRequest{
```

results/pkg/watcher/reconciler/taskrun/taskrun_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func (tt *TaskRunTest) testUpdateTaskRun(t *testing.T) {
+	tr, err := common.ReconcileTaskRun(tt.ctx, tt.asset, tt.taskRun)
+	if err != nil {
+		t.Fatalf("Failed to get completed TaskRun %s: %v", tt.taskRun.Name, err)
+	}
+	tr.UID = "234435"
+	if _, err = tt.asset.Clients.Pipeline.TektonV1beta1().TaskRuns(tt.taskRun.Namespace).Update(tr); err != nil {
```
wlynch, line 93:
```suggestion
	if _, err := tt.asset.Clients.Pipeline.TektonV1beta1().TaskRuns(tt.taskRun.Namespace).Update(tr); err != nil {
```

wlynch (changes requested):

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
XinruZhang: The one in the first line is to get the newest version of pr from both the result client and the pipeline client, and the second one is for testing the `update` operation.

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
XinruZhang: `tt.pipelineRun` is the seed test data, while `pr` being the newest version from the client. There might be some change in the remote client.

results/pkg/watcher/reconciler/taskrun/taskrun.go
XinruZhang: Yeah,, thanks for bring this out! I've tried to pull this logic out as a common function, however, when it comes to the Execution, I need to specify the type, just like `execution.GetPipelineRun()` and `&pb.Execution{Execution: &pb.Execution_PipelineRun{prProto}}`, and they appear a lot, so pulling the logic out as a common function requires a lot of type switches. 
Moreover, the logic seems like the same in the current stage, but in the future, we may want to make different changes for different type. Therefore I prefer we just keep them sparated.

wlynch (approved):
Thanks for your patience with all the changes!

/lgtm

PR #626, opened 2020-09-02
Using CronJobs to automatically clean up completed Runs

psschwei:
# Changes

This PR fixes #479 by providing documentation on how to use a CronJob to clean up old TaskRuns / PipelineRuns. It is based on what @afrittoli implemented in [tektoncd/plumbing#442](https://github.com/tektoncd/plumbing/pull/442).

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/experimental/blob/master/CONTRIBUTING.md)
for more details._

PR #660, opened 2020-10-27
Adding cluster resource to cleanup job

psschwei:
# Changes

As per a [comment](https://github.com/tektoncd/experimental/pull/626#pullrequestreview-517810540) in #626 , adding the ability to select the cluster that the cleanup job runs on. It uses the [kubeconfig create task](https://github.com/tektoncd/catalog/tree/master/task/kubeconfig-creator/0.1) to select the cluster, then the existing cleanup task to do the cleanup.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pipeline/cleanup/binding.yaml
+++ b/pipeline/cleanup/binding.yaml
@@ -4,7 +4,21 @@ metadata:
   name: cleanup-details
 spec:
   params:
-  - name: keep
-    value: $(body.params.cleanup.keep) 
   - name: namespace
     value: $(body.params.target.namespace)
+  - name: keep
+    value: $(body.params.cleanup.keep)
+  - name: filename
+    value: $(body.params.auth.filename)
+  - name: name
+    value: $(body.params.auth.name)
+  - name: url
+    value: $(body.params.auth.url)
+  - name: username
+    value: $(body.params.auth.username)
+  - name: cadata
+    value: $(body.params.auth.cadata)
+  - name: clientKeyData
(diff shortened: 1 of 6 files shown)


pipeline/cleanup/binding.yaml
```diff
@@ -4,7 +4,21 @@ metadata:
...
+  - name: url
+    value: $(body.params.auth.url)
+  - name: username
+    value: $(body.params.auth.username)
+  - name: cadata
+    value: $(body.params.auth.cadata)
+  - name: clientKeyData
+    value: $(body.params.auth.clientkeydata)
+  - name: clientCertificateData
+    value: $(body.params.auth.clientcertdata)
```
afrittoli, line 24: We should probably rely on a secret already existing on the cluster here, so we can avoid passing this sensitive content as parameters.

pipeline/cleanup/cleanup-template.yaml
```diff
@@ -6,40 +6,102 @@ spec:
...
     spec:
       serviceAccountName: tekton-cleaner
-      taskSpec:
-        params:
+      pipelineSpec:
+        workspaces:
+          - name: shared-workspace
+        tasks:
+          - name: kubeconfig-creator
+            taskRef: kubeconfig-creator
```
afrittoli, line 42: Uhm, it really surprises me that the `kubeconfig-creator` task would take secrets like the token as ca data as parameters.

PR #662, opened 2020-10-29
[notifiers] GitHub App Notifier

wlynch:
# Changes

This controller listens for Tekton TaskRuns that are annotated with
specific integration values, and converts the TaskRun into a GitHub
CheckRun with corresponding status and logs.

This is similar in motivation to https://github.com/tektoncd/experimental/tree/master/commit-status-tracker, but uses the GitHub Checks API directly in order to provide finer statuses and richer content.

Still a WIP, things still to do:
- Add k8s deployment configs
- Add settings to control behavior:
  - What gets published (e.g. should logs always be published?)
  - Installation restrictions (e.g. currently you can trick the controller to
    publish to any installation of the app by manually mutating the annotation,
    ideally we would have an ACL somewhere that limits the orgs/users a
    given namespace can access).

This does not do anything to grant the running TaskRun access to the App
credentials (e.g. for cloning the repo) -
https://github.com/tektoncd/catalog/tree/master/task/github-app-token/0.1
can be used for this purpose for the time being (though note this
doesn't support fine-grained installation permissions).

Partially addresses https://github.com/tektoncd/pipeline/issues/999 for GitHub App status notifications.

# Submitter Checklist

review them:

for more details._


chlunde:
Does this replace `commit-status-tracker`?

> Installation restrictions (e.g. currently you can trick the controller to
> publish to any installation of the app by manually mutating the annotation,
> ideally we would have an ACL somewhere that limits the orgs/users a
> given namespace can access).

I noticed the same issue in `commit-status-tracker`. One "solution" would be to *force* status context to be namespace + pipeline name, possibly with a cluster prefix.

This way pipeline X belonging to team A/repo A could never flip a fail to a success in another team/repo, it could just report an *additional* fail/success, and it would be pretty clear who did it..

wlynch:
> Does this replace commit-status-tracker?

No. commit-status-tracker uses the go-scm interface to interact with multiple types of SCM providers (e.g. GitHub OAuth, Bitbucket, GitLab, etc.). This is intentionally focused on GitHub Apps to take advantage of CheckRun-specific features.

> One "solution" would be to force status context to be namespace + pipeline name, possibly with a cluster prefix.

CheckRuns key on the numerical ID and not the name, so you could still overwrite an existing CheckRun.

That said, I like your idea since this can be combined with [required status checks](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-required-status-checks) to prevent this kind of tampering. I'll add it in! :D

chlunde:
> > One "solution" would be to force status context to be namespace + pipeline name, possibly with a cluster prefix.
> 
> CheckRuns key on the numerical ID and not the name, so you could still overwrite an existing CheckRun.

Ah, I see. So this would work for `commit-status-tracker` and the Status API, but not the newer Checks API, without adding a GET to verify ExternalID before `Checks.UpdateCheckRun`.

afrittoli:
It would be nice to update https://github.com/tektoncd/experimental/blob/e0257bddc847a2d09c7d6e0042a9a5bbd2b1ef65/test/presubmit-tests.sh#L61 and add a presubmit.sh script so that CI could run the tests. Apart from that, it would be nice to merge this, even if you don't intend to continue working on it.

wlynch:
> It would be nice to update
> 
> https://github.com/tektoncd/experimental/blob/e0257bddc847a2d09c7d6e0042a9a5bbd2b1ef65/test/presubmit-tests.sh#L61
> 
> and add a presubmit.sh script so that CI could run the tests. Apart from that, it would be nice to merge this, even if you don't intend to continue working on it.

Done!

afrittoli (commented):
/approve
/lgtm

PR #670, opened 2020-11-13
[results] Paginate the List Request

### LINKED ISSUE #654: Results: List pagination
We need to add pagination to the List APIs - https://google.aip.dev/158

The `page_size` and `page_token` fields are already present in the protos, but we need to add the functionality to the API Server.

/assign @XinruZhang

XinruZhang:
Implements the pagination functionality that meets the List API
requirements: https://google.aip.dev/158.
- Encodes the offset as the PageToken.
- Sets the default PageSize as 50.
- The filter is defined in CEL and the default database is MySQL,
  we can't query MySQL directly under the condition of which CEL 
  specifies, so we query a batch from the DB each time, and then 
  manually filter them to compose the final results.

Closes #654 

# Changes

# Submitter Checklist

review them:

for more details._


results/pkg/api/server/server.go
```diff
@@ -271,11 +264,175 @@ func SetupTestDB(t *testing.T) (*Server, error) {
...
+		} else {
+			filteredResults, nextIndex, err = s.getFilteredResult(batchResults, ast, pageSize-got)
+		}
+		if err != nil {
+			return nil, "", err
+		}
+		results = append(results, filteredResults...)
+		filteredNum := int32(len(filteredResults))
+		if filteredNum != 0 {
+			ratio = float32(filteredNum) / float32(batchSize)
```
wlynch, line 422: The ratio heuristic is a bit difficult to reason about as, so at minimum we should add documentation around how this works. IIUC, the idea here is to fetch less items as we approach the number of items we want to return - which is a great idea!

I'm slightly concerned that as we approach the limit, we are trading making more smaller network calls in favor of sending less data. We'll probably want to set a minimum batch size and/or using a simpler selection process to account for this. e.g. would max(remaining results, min batch size) be a sufficient?

results/pkg/api/server/server_test.go
```diff
@@ -316,21 +351,29 @@ func TestListResults(t *testing.T) {
...
+		filter        string
+		pageSize      int32
+		pageToken     string
+		nextPageToken string
+		expect        []*pb.Result
+		expectStatus  codes.Code
 	}{
 		{
-			name:         "test simple query",
+			name:         "test sime query for taskrun",
```
wlynch, line 363:
```suggestion
			name:         "test simple query for taskrun",
```

results/pkg/api/server/server_test.go
```diff
@@ -374,24 +417,70 @@ func TestListResults(t *testing.T) {
...
 			if tc.expectStatus != status.Code(err) {
 				t.Fatalf("failed on test %v: %v", tc.name, err)
 			}
 			gotList := res.GetResults()
+			nextPageToken := res.GetNextPageToken()
+			var expectedNextOffset int64 = 0
+			var gotNextOffset int64 = 0
+			if tc.nextPageToken != "" {
+				expectedNextOffset, _ = decodeToken(tc.nextPageToken)
+				gotNextOffset, _ = decodeToken(nextPageToken)
```
wlynch, line 471: check errors

results/pkg/api/server/server_test.go
```diff
@@ -374,24 +417,70 @@ func TestListResults(t *testing.T) {
...
+				PageSize:  tc.pageSize,
+				PageToken: tc.pageToken,
 			})
 			if tc.expectStatus != status.Code(err) {
 				t.Fatalf("failed on test %v: %v", tc.name, err)
 			}
 			gotList := res.GetResults()
+			nextPageToken := res.GetNextPageToken()
+			var expectedNextOffset int64 = 0
+			var gotNextOffset int64 = 0
```
wlynch, line 468:
```suggestion
			var expectedNextOffset int64
			var gotNextOffset int64
```

results/pkg/api/server/server_test.go
```diff
@@ -374,24 +417,70 @@ func TestListResults(t *testing.T) {
...
+			if tc.nextPageToken != "" {
+				expectedNextOffset, _ = decodeToken(tc.nextPageToken)
```
wlynch, line 470: We should have a separate test for Token encoding/decoding. This test isn't sufficient on its own, since if decodeToken always returned a static value this would still pass.

results/pkg/api/server/server_test.go
```diff
@@ -374,24 +417,70 @@ func TestListResults(t *testing.T) {
...
+				gotNextOffset, _ = decodeToken(nextPageToken)
+			}
+			if !cmp.Equal(gotNextOffset, expectedNextOffset) {
```
wlynch, line 473: cmp.Equal isn't need for primitive comparison.

results/pkg/api/server/server_test.go
```diff
@@ -374,24 +417,70 @@ func TestListResults(t *testing.T) {
...
+				t.Fatalf("NextPageToken mismatch, expected:%v, got:%v", expectedNextOffset, gotNextOffset)
+			}
 			sort.SliceStable(gotList, func(i, j int) bool {
 				return gotList[i].String() < gotList[j].String()
 			})
 			sort.SliceStable(tc.expect, func(i, j int) bool {
 				return tc.expect[i].String() < tc.expect[j].String()
 			})
+			t.Logf("%v\n", gotList)
```
wlynch, line 482: Either include a more useful label to distinguish these logs or remove.

results/pkg/api/server/server_test.go
```diff
@@ -406,3 +495,24 @@ func TaskRunResult(tr *pb.TaskRun) *pb.Result {
...
 }
+
+func PipelineRunResult(pr *pb.PipelineRun) *pb.Result {
+	return &pb.Result{
+		Executions: []*pb.Execution{{
+			Execution: &pb.Execution_PipelineRun{pr},
+		}},
+	}
+}
+
```
wlynch, line 506: This isn't being used, lets remove.

results/pkg/api/server/server_test.go
```diff
@@ -406,3 +495,24 @@ func TaskRunResult(tr *pb.TaskRun) *pb.Result {
...
+	if tr != nil {
+		executions = append(executions, &pb.Execution{Execution: &pb.Execution_TaskRun{tr}})
+	}
+	if pr != nil {
+		executions = append(executions, &pb.Execution{Execution: &pb.Execution_PipelineRun{pr}})
+	}
+	return &pb.Result{
+		Executions: executions,
+	}
+}
```
wlynch, line 518: Making this variadic would let you pass in any combination of Task/PipelineRuns.

```suggestion
func Result(in ...proto.Message) *pb.Result {
	executions := make([]*pb.Execution, 0, len(in))
	for _, m := range in {
		switch x := m.(type) {
		case pb.TaskRun:
			executions = append(executions, &pb.Execution{Execution: &pb.Execution_TaskRun{x}})
		case pb.PipelineRun:
			executions = append(executions, &pb.Execution{Execution: &pb.Execution_PipelineRun{x}})
		default:
			panic("unknown message: %v", m)
		}
	}	
	return &pb.Result{
		Executions: executions,
	}
}
```

wlynch (changes requested):

results/pkg/api/server/server.go
XinruZhang: The key idea is
- the more results remained after filtering, the smaller the next batchSize is.
- no matter how many remained results are, the batchSize only depends on the PageSize and the `ratio`

The following two equations could explain well.
- batchSize = min(PageSize/ratio, max page size)
- ratio = batchFilteredSize/batchSize

Therefore, there wouldn't be the small network calls problem, because the ratio is always less than 1.

The reason why max(remaining results, min batch size) could not work well is, if the ratio is very small, then we may need to query too many times to comprise the final results, and it's hard to determine the value of min batch size.

Hope I expressed myself clearly :)

If needed, I guess we could optimize this function in the future on two aspects:
1. the ratio can be more intelligent(like average over all batch queries).
2. give a maximum of the batchSize(which should be the maximum of the PageSize, the current max size is the maximum of int32)

results/pkg/api/server/server_test.go
XinruZhang: Thanks for the great question!

I did this just to make it easy for human to understand what behind the token. I should just compare the `tc.nextPageToken` and `nextPageToken` :(

As for the seperate test for Token encoding/decoding, I guess the tests  above about `pagination`--different pageToken and pageSize leads to different results and nextPageToken--could be enough?  At least returning a static value will certainly fail on thoses tests.

XinruZhang:
/cc @ImJasonH 
As we mentioned last time, you can take a look at this to know what's going on.

XinruZhang:
This PR is based on #671

results/pkg/api/server/server.go
```diff
@@ -24,6 +27,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	listResultsDefaultPageSize int32 = 50
+	listResultsMaximumPageSize int32 = int32((^uint32(0)) >> 1)
```
ImJasonH, line 32: Can this be some smaller number like 1000 or 10000? If you do intend to support page sizes up to this size it would help readability to just write out the number and add a comment describing why it's that number.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+			taskrun = *(t)
+		}
+		if p := e.GetPipelineRun(); p != nil {
+			pipelinerun = *(p)
+		}
+		out, _, err := prg.Eval(map[string]interface{}{
+			"taskrun":     taskrun,
+			"pipelinerun": pipelinerun,
+		})
+		if err != nil && !strings.Contains(err.Error(), "no such attribute") && !strings.Contains(err.Error(), "undeclared reference to") {
```
ImJasonH, line 233: String matching error messages can be brittle. Is there another way we can detect this error scenario?

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+func matchCelFilter(r *pb.Result, prg cel.Program) (bool, error) {
+	if prg == nil {
+		return true, nil
+	}
+	for _, e := range r.Executions {
+		var (
+			taskrun     pb.TaskRun
+			pipelinerun pb.PipelineRun
+		)
+		if t := e.GetTaskRun(); t != nil {
```
ImJasonH, line 223: Could this just be `taskRun := e.GetTaskRun()` and have the program handle nils?

ImJasonH (commented):
I like the algorithm of choosing batch sizes based on observed retention, I wonder if we can quantify the speed improvement it gives us over always querying some constant number of rows. Constant query sizes has the advantage of consistent and predictable performance per query. We might find out after this is in real use that that's more important than cleverly optimizing the number of queries, but for now this change looks good.

results/schema/results.sql
```diff
@@ -1,4 +1,5 @@
 CREATE TABLE results (
-	name varchar(256) PRIMARY KEY,
+	id INTEGER PRIMARY KEY AUTOINCREMENT,
```
wlynch, line 2: We should avoid using incrementing integers for primary keys - this caps the size of the database based on the size of the int and this can limit our write throughput for distributed workloads if all writes are required to go to the tail.

I'd recommend using a UUID.

wlynch: This also has the consequence of making CreateResult requests non-unique (e.g. I can call CreateResult with the same name multiple times).

If we're going to change the key, we should probably tackle this in another PR to make sure we tackle the edge cases. Otherwise, can we leave the key as is for now?

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
-		for _, e := range r.Executions {
-			out, _, err := prg.Eval(map[string]interface{}{
-				"taskrun": e.GetTaskRun(),
-			})
-			if err != nil {
-				log.Printf("failed to evaluate the expression: %v", err)
-				return nil, status.Errorf(codes.InvalidArgument, "Error occurred during filter evaluation step, no Results found for the query string due to invalid field, invalid function to evaluate filter or missing double quotes around field value, please try to enter a query with correct type again: %v", err)
+		return 0, err
+	}
+	p := &pb.Result{}
```
wlynch, line 265: We probably shouldn't use the full result proto as the pagination token to encode 1 value. Let's create a new proto message (in it's own internal package) that only contains the information we need.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+	if err := proto.Unmarshal(result, p); err != nil {
+		log.Printf("unmarshaling error: %v", err)
+		return 0, status.Errorf(codes.Internal, "failed to parse result data: %v", err)
+	}
+	var primaryKey uint64
+	if err := s.db.QueryRow("SELECT id FROM results where name=?", p.GetName()).Scan(&primaryKey); err != nil {
```
wlynch, line 271: This is susceptible to errors if the underlying result that was used for the pagination token is deleted, so we probably shouldn't depend on this.

If we create the internal proto as suggested above, we can just encode the primary key as part of the response and skip this step altogether.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
 
-	// filter from all results
-	prg, err := s.env.Program(ast)
+func (s Server) decodeTokenToPrimaryKey(token string) (uint64, error) {
+	encodedToken, err := base64.RawURLEncoding.DecodeString(token)
+	if err != nil {
+		return 0, err
+	}
+	result := make([]byte, base64.RawURLEncoding.DecodedLen(len(encodedToken)))
+	_, err = base64.RawURLEncoding.Decode(result, encodedToken)
```
wlynch, line 261: if _, err := ....

results/pkg/api/server/server.go
```diff
@@ -234,6 +374,15 @@ func (s Server) getResultByID(name string) (*pb.Result, error) {
 	return result, nil
 }
 
+func (s Server) getNextPrimaryKey(pk uint64) (bool, uint64) {
```
wlynch, line 377: This should be done transactionally with the list request.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+	} else {
+		lastPK = startPK - 1
+	}
+	var ratio float32 = 1 // the retention rate after filtering a batch of results.
+	// If didn't get enought results.
+	for int32(len(results)) < pageSize {
+		var (
+			batchSize     int32 // batchSize = math.Ceil(pageSize/ratio), has a maximum.
+			batchGot      int32 // <= batchSize, the actual size of a batch query.
+			batchFiltered int32 // <= batchGot, the filtered size of a batch query.
```
wlynch, line 303: nit: batchMatched?

Filtered sounds like the number of elements we filtered out.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+//                  batchSize = pageSize/last_ratio
+// The less the previous ratio is, the bigger the upcoming batch_size is. Then the queried time
+// is significantly decreased.
+func (s Server) getFilteredPaginatedResults(pageSize int32, startPK uint64, prg cel.Program) (results []*pb.Result, lastPK uint64, err error) {
+	if startPK == 0 {
+		lastPK = startPK
+	} else {
+		lastPK = startPK - 1
+	}
+	var ratio float32 = 1 // the retention rate after filtering a batch of results.
```
wlynch, line 297: it's not immediately obvious to me what retention rate means - we should describe it in more detail.

results/pkg/api/server/server.go
```diff
@@ -155,54 +163,186 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
+			batchSize     int32 // batchSize = math.Ceil(pageSize/ratio), has a maximum.
+			batchGot      int32 // <= batchSize, the actual size of a batch query.
+			batchFiltered int32 // <= batchGot, the filtered size of a batch query.
+		)
+		if math.Ceil(float64(pageSize)/float64(ratio)) > float64(listResultsMaximumPageSize) {
+			batchSize = listResultsMaximumPageSize
+		} else {
+			batchSize = int32(math.Ceil(float64(pageSize) / float64(ratio)))
+		}
+		rows, err := s.db.Query("SELECT id, data from results where id > ? limit ? ", lastPK, batchSize)
```
wlynch, line 310: Something I realized from the other db query below- these should all happen in the same transaction.

results/pkg/api/server/server.go
wlynch: We did some digging into this off-thread: CEL doesn't play nice with nil pointers for proto types (unclear if this is intentional behavior or a bug). We should probably document why we're doing this.

wlynch (changes requested):

results/schema/results.sql
XinruZhang: Okay! But we need an ordered field to support pagination, right? I tried to add a timestamp field, but then I thought that there might be multiple records added to the database in the same time, in such case, timestamp is invalid, therefore I turned to an auto-increment integer primary key. Is there any common way to implement this?

results/pkg/api/server/server.go
XinruZhang: Oh, that's great!

results/pkg/api/server/server.go
XinruZhang: Sounds good!

results/pkg/api/server/server.go
XinruZhang: Oh got you, that's great~

results/pkg/api/server/server.go
```diff
@@ -26,6 +28,15 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	listResultsDefaultPageSize int32 = 50
+	listResultsMaximumPageSize int32 = 10000
+)
+
+var uid = func() string {
+	return fmt.Sprintf("%v%s", time.Now().UnixNano(), uuid.New())
```
wlynch, line 37: uuid.New().String() is fine - UUIDs already include time information as a nonce.

results/proto/v1alpha1/api.proto
```diff
@@ -180,4 +180,9 @@ message ListResultsRequest {
 message ListResultsResponse {
   repeated Result results = 1;
   string next_page_token = 2;
+}
+
+message ListPageIdentifier {
+	string resultName = 1;
+  string filter = 2;
```
wlynch, line 187: We should break this message out into its own `internal` proto package so that it's not included in the API definition.

results/proto/v1alpha1/api.proto
```diff
@@ -180,4 +180,9 @@ message ListResultsRequest {
 message ListResultsResponse {
   repeated Result results = 1;
   string next_page_token = 2;
+}
+
+message ListPageIdentifier {
+	string resultName = 1;
```
wlynch, line 186: formatting.

results/pkg/api/server/server.go
```diff
@@ -157,51 +168,71 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
-	var resp []*pb.Result
-	for _, r := range results {
-		if ok, err := matchCelFilter(r, prg); err != nil {
-			return nil, err
-		} else if ok {
-			resp = append(resp, r)
+	var tx *sql.Tx
+	if tx, err = s.db.Begin(); err != nil {
+		return nil, err
+	}
```
wlynch, line 211: This is a case where it's more idiomatic to not use the inline if syntax - rule of thumb is to use it when the used values are only needed within the if.

```suggestion
	tx, err := s.db.Begin()
	if err != nil {
		return nil, err
	}
```

results/pkg/api/server/server.go
```diff
@@ -157,51 +168,71 @@ func (s *Server) ListResultsResult(ctx context.Context, req *pb.ListResultsReque
...
-			log.Printf("unmarshaling error: %v", err)
-			return nil, status.Errorf(codes.Internal, "failed to parse result data: %v", err)
+		if req.GetFilter() != pageIdentifier.GetFilter() {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("use a different CEL `filter` from the last page."))
 		}
-		results = append(results, r)
 	}
 
+	var prg cel.Program
+	var results []*pb.Result
```
wlynch, line 196: Try to define variables as close as to where they are used as possible.

results/pkg/api/server/server.go
```diff
@@ -229,9 +260,114 @@ func matchCelFilter(r *pb.Result, prg cel.Program) (bool, error) {
...
+// EncodePageResult encodes a ListPageIdentifier to PageToken
+func encodePageResult(pi *pb.ListPageIdentifier) (token string, err error) {
+	var tokenByte []byte
+	if tokenByte, err = proto.Marshal(pi); err != nil {
+		return "", err
+	}
+	encodedResult := make([]byte, base64.RawURLEncoding.EncodedLen(len(tokenByte)))
+	base64.RawURLEncoding.Encode(encodedResult, tokenByte)
+	token = base64.RawURLEncoding.EncodeToString(encodedResult)
+	return token, nil
```
wlynch, line 272:
```suggestion
	return base64.RawURLEncoding.EncodeToString(encodedResult), nil
```

wlynch (commented):
Almost there! just a few nits.

wlynch (approved):
/lgtm

PR #676, opened 2020-12-01
[results] Implement v1alpha2 Create/GetResult.

wlynch:
# Changes

Implements the CreateResult and GetResult methods of the v1alpha2
interface.

Includes request validation, and some basic error handling.
Unfortunately since SQL error codes are somewhat impl specific, we will
need to do some additional work to fill out error handling (particularly
for MySQL). For now, focus on gorm/sqlite errors for unit testing.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/results/pkg/api/server/test/db.go
+++ b/results/pkg/api/server/test/db.go
@@ -1,0 +1,56 @@
+package test
+
+import (
+	"database/sql"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+// NewDB set up a temporary database for testing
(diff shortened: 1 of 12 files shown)


results/pkg/api/server/test/db.go
```diff
@@ -0,0 +1,56 @@
+package test
```
dibyom, line 1: In other tekton projects, we've been following the pattern of putting all test helpers/e2e tests in a separate test/ folder that is in the same level as pkg/

dibyom (commented):
/approve

results/pkg/api/server/test/db.go
wlynch: I'm not sure we want to do that here - I suspect that we may want to split the API Server / watcher into separate modules in the future, so keeping this closer to the rest of the API Server code will make that transition easier if/when that happens.

PR #681, opened 2020-12-02
[results] Implements the ETAG to solve the race condition problem

### LINKED ISSUE #616: Resolving race condition in watcher in Tekton Results API
[Design Proposal](https://docs.google.com/document/d/1-XBYQ4kBlCHIHSVoYAAf_iC01_by_KoK2aRVO0t8ZQ0/edit)

Currently there is a race condition which could happen when a large number of new TaskRuns are sent to Reconciler in the watcher, for example, during a new TaskRun is being created in results server by a CreateTaskRun  request in Reconciler, another TaskRun could already finish creation and step over this one and take results_id from it.

To clear this issue, we need to design a thread-safe scheme to prevent such race condition to happen.

XinruZhang:
# Changes
There might be multiple update requests for one result in the same period, which could lead to the race condition problem.

To avoid that, we need to make sure the result in our update request untouched before being updated to the database.  We achieve this by the `etag`. A valid update request must carry the ETag of the result that to be updated. We encode the serialized result to a string as its etag.

The etag is like a summary of a result's content. Once the content changed, the etag would be different. Check [here](https://developers.google.com/gdata/docs/2.0/reference#ResourceVersioning) for more information.

Closes #616 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/results/pkg/api/server/v1alpha1/server.go
+++ b/results/pkg/api/server/v1alpha1/server.go
@@ -48,6 +48,13 @@ func (s *Server) CreateResult(ctx context.Context, req *pb.CreateResultRequest) 
 	name := uuid.New().String()
 	r.Name = fmt.Sprintf("%s/results/%s", req.GetParent(), name)
 
+	if etag, err := getResultETag(r); err != nil {
+		log.Printf("etag generating error: %v", err)
+		return nil, fmt.Errorf("failed to generate the etag for a result: %w", err)
+	} else {
+		r.Etag = etag
+	}
+
 	// serialize data and insert it into database.
 	b, err := proto.Marshal(r)
 	if err != nil {
@@ -123,6 +130,17 @@ func (s Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (
 	}
 	if r.GetCreatedTime() != prev.GetCreatedTime() {
 		return prev, status.Error(codes.InvalidArgument, "created time cannot be changed")
+	}
+
+	// Check if there's any race condition problem.
+	if prev.GetEtag() != r.GetEtag() {
+		return nil, status.Error(codes.Aborted, "failed on the race condition, the content of this result has been modified.")
+	}
+	if etag, err := getResultETag(r); err != nil {
+		log.Printf("etag generating error: %v", err)
+		return nil, fmt.Errorf("failed to generate the etag for a result: %w", err)
+	} else {
+		r.Etag = etag
 	}
 
 	// Write result back to database.
@@ -399,6 +417,18 @@ func (s Server) getResultByID(name string) (*pb.Result, error) {
 		return nil, status.Error(codes.NotFound, "result not found")
 	}
 	return result, nil
+}
+
+func getResultETag(result *pb.Result) (etag string, err error) {
+	tmpEtag := result.GetEtag()
+	result.Etag = ""
+	if b, err := proto.Marshal(result); err == nil {
+		dst := make([]byte, base64.RawURLEncoding.EncodedLen(len(b)))
+		base64.RawURLEncoding.Encode(dst, b)
+		etag = base64.RawURLEncoding.EncodeToString(dst)
+	}
+	result.Etag = tmpEtag
+	return
 }
 
 // SetupTestDB set up a temporary database for testing
--- a/results/pkg/api/server/v1alpha1/server_test.go
+++ b/results/pkg/api/server/v1alpha1/server_test.go
@@ -72,6 +72,16 @@ func TestGetResult(t *testing.T) {
 }
 
 func TestUpdateResult(t *testing.T) {
+	t.Run("Test etag generation", func(t *testing.T) {
+		r1 := Result(&ppb.TaskRun{})
+		etag1 := obtainResultEtag(t, r1)
+		r1.Annotations = map[string]string{"foo": "bar"}
+		etag2 := obtainResultEtag(t, r1)
+		if etag1 == etag2 {
+			t.Fatalf("different results can't have the same etag")
+		}
+	})
+
 	// Create a temporary database
 	srv, err := SetupTestDB(t)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


results/pkg/api/server/v1alpha1/server.go
```diff
@@ -48,6 +48,13 @@ func (s *Server) CreateResult(ctx context.Context, req *pb.CreateResultRequest)
 	name := uuid.New().String()
 	r.Name = fmt.Sprintf("%s/results/%s", req.GetParent(), name)
 
+	if etag, err := getResultETag(r); err != nil {
+		log.Printf("etag generating error: %v", err)
+		return nil, fmt.Errorf("failed to generate the etag for a result: %w", err)
+	} else {
+		r.Etag = etag
+	}
```
wlynch, line 56:
```suggestion
	etag, err := getResultETag(r)
	if err != nil {
		log.Printf("etag generating error: %v", err)
		return nil, fmt.Errorf("failed to generate the etag for a result: %w", err)
	}
	r.Etag = etag
```

results/pkg/api/server/v1alpha1/server.go
```diff
@@ -125,6 +132,17 @@ func (s Server) UpdateResult(ctx context.Context, req *pb.UpdateResultRequest) (
 		return prev, status.Error(codes.InvalidArgument, "created time cannot be changed")
 	}
 
+	// Check if there's any race condition problem.
+	if prev.GetEtag() != r.GetEtag() {
+		return nil, status.Error(codes.Aborted, "failed on the race condition, the content of this result has been modified.")
```
wlynch, line 137: https://google.aip.dev/154 suggests using FAILED_PRECONDITION instead.

results/pkg/api/server/v1alpha1/server.go
```diff
@@ -401,6 +419,18 @@ func (s Server) getResultByID(name string) (*pb.Result, error) {
 	return result, nil
 }
 
+func getResultETag(result *pb.Result) (etag string, err error) {
+	tmpEtag := result.GetEtag()
+	result.Etag = ""
+	if b, err := proto.Marshal(result); err == nil {
+		dst := make([]byte, base64.RawURLEncoding.EncodedLen(len(b)))
+		base64.RawURLEncoding.Encode(dst, b)
+		etag = base64.RawURLEncoding.EncodeToString(dst)
```
wlynch, line 428: I'm slightly worried that this will inflate the size of a record, since we're effectively duplicating the data.

This is probably fine for a first pass, but we should look into a slimmed down identifier in the future (perhaps update timestamp?)

results/pkg/api/server/v1alpha1/server.go
```diff
@@ -401,6 +419,18 @@ func (s Server) getResultByID(name string) (*pb.Result, error) {
 	return result, nil
 }
 
+func getResultETag(result *pb.Result) (etag string, err error) {
```
wlynch, line 422: Avoid naming parameters unless it helps identify what's being returned (e.g. if you're returning multiple things of the same type).

This can often make things harder to read and mask errors. In this case, it's actually making it harder to see that this func will never return an err since the value is scoped to the if statement (https://play.golang.org/p/MVnMta2Heme for a simplified example).

wlynch, line 422: Even though it's not an exported func, let's add godoc to explain what this is doing so it's clear when you want to use this vs `result.GetEtag()`

results/pkg/api/server/v1alpha1/server.go
```diff
@@ -401,6 +419,18 @@ func (s Server) getResultByID(name string) (*pb.Result, error) {
...
+	tmpEtag := result.GetEtag()
+	result.Etag = ""
+	if b, err := proto.Marshal(result); err == nil {
+		dst := make([]byte, base64.RawURLEncoding.EncodedLen(len(b)))
+		base64.RawURLEncoding.Encode(dst, b)
+		etag = base64.RawURLEncoding.EncodeToString(dst)
+	}
+	result.Etag = tmpEtag
```
wlynch, line 430: Instead of having callers set the Etag, we can just set `result.Etag = etag` here.

results/pkg/api/server/v1alpha1/server_test.go
```diff
@@ -72,6 +72,16 @@ func TestGetResult(t *testing.T) {
...
+	t.Run("Test etag generation", func(t *testing.T) {
+		r1 := Result(&ppb.TaskRun{})
+		etag1 := obtainResultEtag(t, r1)
+		r1.Annotations = map[string]string{"foo": "bar"}
+		etag2 := obtainResultEtag(t, r1)
+		if etag1 == etag2 {
+			t.Fatalf("different results can't have the same etag")
+		}
+	})
+
```
wlynch, line 84: Let's break this into it's own test.

results/pkg/api/server/v1alpha1/server_test.go
```diff
@@ -72,6 +72,16 @@ func TestGetResult(t *testing.T) {
 }
 
 func TestUpdateResult(t *testing.T) {
+	t.Run("Test etag generation", func(t *testing.T) {
+		r1 := Result(&ppb.TaskRun{})
+		etag1 := obtainResultEtag(t, r1)
+		r1.Annotations = map[string]string{"foo": "bar"}
+		etag2 := obtainResultEtag(t, r1)
+		if etag1 == etag2 {
+			t.Fatalf("different results can't have the same etag")
```
wlynch, line 81: We probably want a more exact test here that compares the expected etag output. Technically this could still pass as long as the etag generation was non-deterministic. 2 test cases come to mind:

1) Result with no etag, verify this is filled in and set properly.
2) Result with etag, verify this is overridden.

results/pkg/watcher/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -119,6 +119,7 @@ func (tt *PipelineRunTest) testUpdatePipelineRun(t *testing.T) {
 		Executions: []*pb.Execution{{
 			Execution: &pb.Execution_PipelineRun{p},
 		}},
+		Etag: res.Etag,
```
wlynch, line 122: We should avoid tautological tests - this would pass regardless of the etag value. We should either generate this from the wanted result, or loosen the etag restriction to make it not required for requests.

wlynch (changes requested):
/assign wlynch

XinruZhang:
moved out to the new repo.

PR #686, opened 2020-12-03
[results] Fix flaky reconciler tests.

wlynch:
# Changes

I don't fully understand why this was sometimes passing before, but
the likely guess is that the reconciler was updating the pointer
reference somewhere, or test ordering was interfering with results with
the same name.

This resolves the problem by doing 2 things:

1. Always use a fresh results server / reconciler for each test. This
helps prevent test crossover / ordering from affecting the output.
2. Do an initial reconcile instead of a Get to ensure that all the
necessary IDs are set.

Tested that `go test -count 50 ./pkg/watcher/reconciler/...` works,
which should give us decent confidence that this is fixed.

# Submitter Checklist

review them:

for more details._


dibyom (approved):
/lgtm

(we create new reconcilers each time for the tests in triggers/pipelines too(

PR #692, opened 2021-01-19
[CEL] Common Expression Language Custom Task

jerop:
# Changes

This is the initial of an experimental project that provides support for Common Expression Language (CEL) in Tekton Pipelines.

The functionality is provided by a controller that implements the Custom Task interface. Its use cases include evaluating complex expressions to be used in [`WhenExpressions`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#guard-task-execution-using-whenexpressions) in subsequent `Tasks` to guard their execution.

To evaluate a CEL expressions using `Custom Tasks`, we need to define a [`Run`](https://github.com/tektoncd/pipeline/blob/master/docs/runs.md) type with `apiVersion: cel.tekton.dev/v1alpha1` and `kind: CEL`. The `Run` takes the CEL expressions to be evaluated as `Parameters`.

As the `Run` executes, its `status` field accumulates information about the execution status of the `Run` in general. If the evaluation is successful, it will also contain the `Results` of the evaluation under `status.results` with the corresponding names of the CEL expressions as provided in the `Parameters`. If no CEL expressions are provided, any CEL expression is invalid or there's any other error, the `CEL` `Custom Task` will fail and the details will be included in `status.conditions`. 

Related issues:
- https://github.com/tektoncd/community/issues/304
- https://github.com/tektoncd/pipeline/issues/3149
- https://github.com/tektoncd/pipeline/issues/2812
- https://github.com/tektoncd/pipeline/issues/2219
- https://github.com/tektoncd/pipeline/issues/2127
- https://github.com/tektoncd/pipeline/issues/1393

# Submitter Checklist

review them:


cel/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Install
+
+Install and configure `ko`
+
+```
+ko apply -f celcontroller.yaml
+```
+
+This will build and install the controller on your cluster, in the namespace `cel`
```
ImJasonH, line 23: nit: Can this be `tekton-cel-task` or something? There might be other things on the cluster related to CEL and we shouldn't namesquat it.

cel/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Usage
+
+Two resources are required to evaluate a `CEL` expression using a `Custom Task` are:
+- `CELTask` which defines the API version and identifies the resource object as `CEL`
+- `CELRun` which provides the `CEL` expressions to be evaluated and executes the `CELTask`
+  to evaluate the `CEL` expressions
+
+### Configuring a `CELRun`
```
ImJasonH, line 32: There isn't (or, there shouldn't be) a type called `CELRun`, this should just be a `v1alpha1.Run` type with `apiVersion: cel.tekton.dev/v1alpha1` and `kind: CEL`.

cel/README.md
```diff
@@ -0,0 +1,221 @@
...
+```
+ko apply -f celcontroller.yaml
+```
+
+This will build and install the controller on your cluster, in the namespace `cel`
+
+## Usage
+
+Two resources are required to evaluate a `CEL` expression using a `Custom Task` are:
+- `CELTask` which defines the API version and identifies the resource object as `CEL`
```
ImJasonH, line 28: Unless we expand on this work some more, we don't actually need to define a `CELTask` CRD type, we can just require that `Run`s that evaluate a CEL expression have the apiVersion+kind, and no `name`, and a `param` named `expression`.

ImJasonH (commented):
This looks great! Just a couple comments about docs.

cel/README.md
jerop: sure, updated it

cel/README.md
jerop: yes yes, updated the docs

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Create a program environment configured with the standard library of CEL functions and macros
+	env, err := cel.NewEnv(cel.Declarations())
+	if err != nil {
+		logger.Errorf("cel.NewEnv: %v", err)
+		return err
+	}
+
+	// Iterate through the CEL expressions
+	for _, param := range o.Spec.Params {
```
ImJasonH, line 73: I think we want to pull out the param named `expression`, fail if it isn't found, and only evaluate that param value. This should also fail if any other params are defined, since that's probably a typo/bug.

I don't think we want/need to support multiple expressions in one run, since it's a bit unclear how it should work, and it's probably clearer (though much more verbose) to just have multiple CEL-typed Runs if you want that.

jerop: > since it's a bit unclear how it should work, and it's probably clearer (though much more verbose) to just have multiple CEL-typed Runs if you want that.

the way I have it implemented now is that the `Run` will evaluate each `Parameter` and produce a `Result` with the corresponding name (i.e. parameter name == result name), did that to reduce the verbosity and evaluate multiple expressions in one pod when appropriate, what do you think of that?

ImJasonH: Yeah, I think I'm coming around to it. It is better than having N single-expression Runs.

One possible confusing thing is that if a CEL Run has three expression params and one is an invalid expression, will it fail but still produce results for the other two expressions? Will it still evaluate params defined after the invalid expression? There are some options:

1. Fail and don't report any results if any expression param is invalid.
2. (this is what this PR does) Evaluate each expression in the order they're defined in `params`, fail when one is invalid and only report results for previous valid evaluations.
3. Evaluate each expression, fail when one is invalid but keep evaluating params and report results for any that are valid expressions. You can report either the first, or _all_, invalid expressions in `status`.

Either way I don't think it should block this PR, it'd be better to get usage and feedback.

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,121 @@
...
+		if param.Value.StringVal == "" {
+			logger.Errorf("Invalid CEL expression parameter %s in Run %s/%s", param.Name, o.Namespace, o.Name)
+			o.Status.MarkRunFailed("InvalidExpression",
+				"CEL expression parameter %s was invalid in Run %s/%s", param.Name, o.Namespace, o.Name)
```
ImJasonH, line 77: It doesn't really add much to include the name+namespace of the Run in the status of the Run, the user presumably already knows that. This error message should indicate there was no string value in the parameter.

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil
+		}
+
+		// Combine the Parse and Check phases CEL program compilation to produce an Ast and associated issues
+		ast, iss := env.Compile(param.Value.StringVal)
+		if iss.Err() != nil {
+			logger.Errorf("CEL expression %s could not be parsed when reconciling %s/%s", param.Name, o.Namespace, o.Name)
+			o.Status.MarkRunFailed("SyntaxError",
+				"CEL expression %s could not be parsed in Run %s/%s", param.Name, o.Namespace, o.Name, err)
```
ImJasonH, line 86: ditto on including name+namespace, but make sure to include the error message (and in the `logger.Errorf` on the previous line).

And same below, you can remove the references to `o.Name` and `o.Namespace`.

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	if o.IsDone() {
+		logger.Infof("Run %s/%s is done", o.Namespace, o.Name)
+		return nil
+	}
+
+	// Create a program environment configured with the standard library of CEL functions and macros
+	env, err := cel.NewEnv(cel.Declarations())
```
souleb, line 66: I think we should have all the cel code in a separate file. I would have a struct with methods that perform the different CEL steps. Would be easier to extend if we decide to add custom functions for example... I think :grin:

cel/pkg/reconciler/cel/celrun.go
souleb: As a user I would expect it to fail if any param expression is invalid. I think I assume a full validation before it runs

jerop: I've added an earlier validation to check that that at least one expression is provided and that the expression types are valid

To be consistent with `Tasks` which don't produce `Results` when they fail, I've modified so that we go with option 1 where we don't report any evaluation result if any expression is invalid or has any syntax/evaluation error -- and I agree that we can revisit those options based on user feedback

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Reconciler implements controller.Reconciler for Run resources.
+type Reconciler struct {
+}
+
+// Check that our Reconciler implements Interface
+var _ run.Interface = (*Reconciler)(nil)
+
+// ReconcileKind implements Interface.ReconcileKind.
+func (r *Reconciler) ReconcileKind(ctx context.Context, o *v1alpha1.Run) reconciler.Event {
```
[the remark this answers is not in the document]
jerop: done, thanks!

cel/pkg/reconciler/cel/celrun.go
jerop: For this custom task, we don't have a CEL CRD type because parameters can be used to provide all that we need (CEL expressions) -- we can define a CEL CRD type in the future if there are other use cases we want to solve for, but would prefer to keep it simple for now

cel/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+The `CEL` `Custom Task` can be specified within a `Pipeline`, as such:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  generateName: pr-
+spec:
+  pipelineSpec:
```
pritidesai, line 78: `pipelineSpec` is not needed here 🙃

bigkevmcd (approved):
Looks really good to me, thanks for taking this on.

cel/pkg/reconciler/cel/celrun.go
```diff
@@ -0,0 +1,162 @@
...
+		out, _, err := prg.Eval(map[string]interface{}{})
+		if err != nil {
+			logger.Errorf("CEL expression %s could not be evaluated when reconciling Run %s/%s: %v", param.Name, run.Namespace, run.Name, err)
+			run.Status.MarkRunFailed(ReasonEvaluationError,
+				"CEL expression %s could not be evaluated", param.Name, err)
+			return nil
+		}
+
+		// Evaluation of CEL expression was successful
+		logger.Infof("CEL expression %s evaluated successfully when reconciling Run %s/%s", param.Name, run.Namespace, run.Name)
```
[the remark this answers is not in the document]
ImJasonH: OTOH, that value might be large/spammy, and it might even be something sensitive like a secret (though users shouldn't pass secrets via results/params, but 🤷‍♂️ ). I'd like to bias toward not logging this result out-of-the-box.

pritidesai:
haha, I was still reviewing it 🙃  very excited for this 😍

jerop:
> haha, I was still reviewing it 🙃 very excited for this 😍

this is experimental so welcoming your feedback and we can modify as needed, could be easier to discuss in the TEP: https://github.com/tektoncd/community/pull/314

cel/README.md
```diff
@@ -0,0 +1,274 @@
...
+Events:                    <none>
+```
+
+For more information about monitoring `Run` in general, read [monitoring execution status](https://github.com/tektoncd/pipeline/blob/master/docs/runs.md#monitoring-execution-status).
+
+### Using the evaluation results
+
+A successful `Run` contains the `Results` of evaluating the CEL expressions under `status.results`, with the name of
+each evaluation `Result` matching the name of the corresponding CEL expression as provided in the `Parameters`.
+Users can reference the `Results` in subsequent `Tasks` using variable substitution, as such:
```
bobcatfish, line 223: hey @jerop sorry for the late comment, just wondering' what's the type and structure of the result? (e.g. json? or some CEL specific thing? maybe we can link to some CEL docs about this?)

PR #705, opened 2021-02-22
[PIP] Pipelines In Pipelines

jerop:
# Changes

This is an experimental project to provide support for `Pipelines` in
`Pipelines` to improve the composability and reusability of [Tekton Pipelines](https://github.com/tektoncd/pipeline). 
The functionality is provided by a controller that implements the 
`Custom Task` interface.

Its use cases include enabling defining and executing a set of `Tasks`
as a complete unit of execution and decoupling `Pipeline` branches
failures such that a `Task` failure in one branch does not stop
execution of unrelated `Pipeline` branches.

Pipelines in Pipelines is proposed in tektoncd/community#330
Issue is further described in tektoncd/pipeline#2134

Future work includes enabling passing in inputs through `Parameters`, 
producing outputs through `Results`, propagating timeouts, among others. 

/kind feature

# Submitter Checklist

review them:

/cc @afrittoli @bobcatfish @ImJasonH


pipelines-in-pipelines/config/100-namespace.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tekton-pip-run
```
afrittoli, line 18: Any special reason for running this in a different namespace by default?

pipelines-in-pipelines/config/config-logging.yaml
```diff
@@ -0,0 +1,53 @@
...
+      "development": false,
+      "sampling": {
+        "initial": 100,
+        "thereafter": 100
+      },
+      "outputPaths": ["stdout"],
+      "errorOutputPaths": ["stderr"],
+      "encoding": "json",
+      "encoderConfig": {
+        "timeKey": "",
```
afrittoli, line 37: NIT: I think it would be nice to have a timestamp key here so we can get timestamps in the logs.

pipelines-in-pipelines/go.mod
```diff
@@ -0,0 +1,33 @@
...
+
+// Pin k8s deps to v0.18.8
+replace (
+	k8s.io/api => k8s.io/api v0.18.8
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.18.8
+	k8s.io/apimachinery => k8s.io/apimachinery v0.18.8
+	k8s.io/client-go => k8s.io/client-go v0.18.8
+	k8s.io/code-generator => k8s.io/code-generator v0.18.8
+	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29
+)
```
afrittoli, line 33: Do we need these replace? 
It's not a problem if we do, just something that we'd need to maintain in future.

pipelines-in-pipelines/pkg/reconciler/pip/piprun.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (r *Reconciler) reconcile(ctx context.Context, run *v1alpha1.Run) error {
+	logger := logging.FromContext(ctx)
+
+	// confirm the run spec is valid
+	if err := validate(run); err != nil {
+		logger.Errorf("Run %s/%s is invalid because of %v", run.Namespace, run.Name, err)
+		run.Status.MarkRunFailed(ReasonRunFailedValidation,
+			"Run can't be run because it has an invalid spec - %v", err)
+		return nil
```
afrittoli, line 122: It would be nice to return a [permanent error](https://github.com/knative/eventing/blob/9b0599f33985020f445820ce6ee843d5ab701122/vendor/knative.dev/pkg/controller/controller.go#L425) instead.
The controller will not re-queue the key in case of permanent errors, but it will be aware about the queue (and an event will be emitted for it).

pipelines-in-pipelines/pkg/reconciler/pip/piprun_test.go
```diff
@@ -0,0 +1,381 @@
...
+			t.Errorf("Expected reason to be %q but was %q", expectedReason, condition.Reason)
+		}
+		if condition.Message != expectedMessage {
+			t.Errorf("Expected message to be %q but was %q", expectedMessage, condition.Message)
+		}
+	}
+	if run.Status.StartTime == nil {
+		t.Errorf("Expected Run start time to be set but it wasn't")
+	}
+	if expectedStatus == corev1.ConditionUnknown {
```
afrittoli, line 86: Nice :)

pipelines-in-pipelines/pkg/reconciler/pip/piprun_test.go
```diff
@@ -0,0 +1,381 @@
...
+		if run.Status.CompletionTime != nil {
+			t.Errorf("Expected Run completion time to not be set but it was")
+		}
+	} else if run.Status.CompletionTime == nil {
+		t.Errorf("Expected Run completion time to be set but it wasn't")
+	}
+}
+
+func checkEvents(fr *record.FakeRecorder, testName string, wantEvents []string) error {
```
afrittoli, line 95: This function is getting copied in a few places, I wonder if we should export it from a test packages somewhere... but it's ok to copy it here :)

afrittoli (commented):
Thanks for this 🎉 🙏 
It looks great, I added a couple of NITs / questions / comments, but nothing that cannot be addressed in a different patch.
/approve

afrittoli:
Perhaps we could merge https://github.com/tektoncd/experimental/pull/706 first so we could see the tests running?

pipelines-in-pipelines/config/config-logging.yaml
jerop: sounds good, added :)

pipelines-in-pipelines/go.mod
jerop: yes yes, otherwise I hit some issues like this https://github.com/kubernetes/client-go/issues/741 

we use the same deps in cel custom tasks, so can rework them separately

pipelines-in-pipelines/pkg/reconciler/pip/piprun.go
jerop: great idea, changed it to return permanent error :)

pipelines-in-pipelines/pkg/reconciler/pip/piprun_test.go
jerop: agreed, will need to do that because it's used to check events in all the custom tasks controllers

pipelines-in-pipelines/config/100-namespace.yaml
jerop: we run them in a separate namespace to make it clear what was created by Pipelines in Pipelines controller, and make it easy to delete all its resources as needed

jerop: if/when the feature is promoted from experimental to top level, we can switch to `tekton-pipelines` namespace at that time

jerop: /cc @ImJasonH

afrittoli:
Thanks for all the updates @jerop .
I don't see the changes from #706 in here?
Thank you 🙏

jerop:
> Thanks for all the updates @jerop .
> I don't see the changes from #706 in here?
> Thank you 🙏

just pushed it, thanks :)

afrittoli:
> > Thanks for all the updates @jerop .
> > I don't see the changes from #706 in here?
> > Thank you 🙏
> 
> just pushed it, thanks :)

Thank you! 
Unit tests are running for PIP now, it looks like there is a failure.

jerop:
> Unit tests are running for PIP now, it looks like there is a failure.

yes, needed to add the event from the permanent error I added, fixed now 

thanks for the reviews @afrittoli!

afrittoli (commented):
/lgtm

PR #717, opened 2021-03-04
[cel] parse variables in context

### LINKED ISSUE #716: Add ability to parse variables in expression from context for CEL
### Feature request

The CEL is a great feature, we plan to adopt in our project, but seems it can not support parse expression which include other variables, for example:
```
  params:
    - name: name
      value: "'hello'"
    - name: types
      value: "type(name)"
```
see the `name` has been defined in first `param` and expect to used in second `param`. 
I guess this is not required in `when express`, but useful in other scenario.

In our case, we will set some initialize variables in some runtime storage (maybe a runtime created CM), and in expression, we will use these variables.

vincent-pli:
fix #716 
# Changes

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cel/README.md
+++ b/cel/README.md
@@ -135,6 +135,37 @@ spec:
 
 For more information about specifying `Parameters`, read [specifying parameters](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#specifying-parameters).
 
+### Use context variables in CEL expression
+
+Besides use plain text in CEL expression we can also adopt `context variables`, means we can use `parameter` defined previously in subsequent `parameter`, like this:
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  generateName: celrun-with-context-
+spec:
+  ref:
+    apiVersion: cel.tekton.dev/v1alpha1
+    kind: CEL
+  params:
+    - name: name
+      value: "'hello'"
+    - name: types
+      value: "type(name)"
+    - name: expression-use-context
+      value: "name + ' is type of ' + types"
+```
+
+The expression: `"name + ' is type of ' + types"` adopt `name` and `types` which defined previously as part of the expression.
+so the result should be: "hello is type of string".
+
+*CATION*: 
+- Take care of the order of the `parameter`s, the subsequent one will depends on the previous ones.
+- To refer to previous `parameter`, use the name of `parameter` directly in the expression, and use single quotation marks for static string, I mean `'name'`.
+
+See [example](examples/celrun-with-context.yaml) here.
+
+
 ### Monitoring execution status
 
 As the `Run` executes, its `status` field accumulates information about the execution status of the `Run` in general.
--- a/cel/examples/celrun-with-context.yaml
+++ b/cel/examples/celrun-with-context.yaml
@@ -1,0 +1,15 @@
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
+  generateName: celrun-with-context-
+spec:
+  ref:
+    apiVersion: cel.tekton.dev/v1alpha1
+    kind: CEL
+  params:
+    - name: name
+      value: "'hello'"
+    - name: types
+      value: "type(name)"
+    - name: expression-use-context
+      value: "name + ' is type of ' + types"
--- a/cel/pkg/reconciler/cel/celrun.go
+++ b/cel/pkg/reconciler/cel/celrun.go
@@ -19,7 +19,9 @@ package cel
 import (
 	"context"
 	"fmt"
+
 	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/types"
 	"github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1alpha1/run"
 	"knative.dev/pkg/apis"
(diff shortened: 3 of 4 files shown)


afrittoli (commented):
Thanks @vincent-pli !
It's great to see this being used :)

A few small asks 🙏 

- could you add "[cel]" in beginning of the commit message
- could you add a unit test too for the changes you've done?

jerop (commented):
Thanks for the contribution to this project @vincent-pli! 

I have a couple of concerns about this change:
1. How can users distinguish between a static string and a context variable? In your example, a user can no longer use static strings `name` and `types` because they'll interpreted as variable strings that are substituted with `hello` and `string` respectively
1. In `Pipelines`, we don't substitute `Parameter` values with `Results` from the same `Task` so doing that here would be inconsistent and confusing to users 
1. What happens when the order of the expressions is rearranged?
   1. This change introduces dependencies between the `Parameters` (Expressions) which requires that users are careful about the ordering of the `Parameters` (Expressions) 
   1. Is the order of `Parameters` in the YAML maintained when marshalling & unmarshalling? 

One way to solve issue the first concern would be to express the context variables as variables using `$(...)` to distinguish them from the static strings, that is:

```yaml
# current proposal
apiVersion: tekton.dev/v1alpha1
kind: Run
metadata:
  generateName: celrun-with-context-
spec:
  ref:
    apiVersion: cel.tekton.dev/v1alpha1
    kind: CEL
  params:
    - name: name
      value: "'hello'"
    - name: types
      value: "type(name)"
    - name: expression-use-context
      value: "name + ' is type of ' + types"
---
# suggested proposal
apiVersion: tekton.dev/v1alpha1
kind: Run
metadata:
  generateName: celrun-with-context-
spec:
  ref:
    apiVersion: cel.tekton.dev/v1alpha1
    kind: CEL
  params:
    - name: name
      value: "'hello'"
    - name: types
      value: "type(name)"
    - name: expression-use-context
      value: "$(result.name) + ' is type of ' + $(result.types)"

```

An alternative approach I'd prefer would be to support substituting `Parameters` from the `Pipeline` and `Results` from other `Tasks` in `CEL Custom Tasks`, similarly to what we do in the rest of the `Pipeline`, something like:

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: example-pr-
spec:
  params:
    - name: name
      value: "hello"
  pipelineSpec:
    params:
      - name: name
        type: string
    tasks:
      - name: get-type
        taskRef:
          apiVersion: cel.tekton.dev/v1alpha1
          kind: CEL
        params:
          - name: type
            value: "$(params.name)"
      - name: use-type
        taskRef:
          apiVersion: cel.tekton.dev/v1alpha1
          kind: CEL
        params:
          - name: expression-use-type
            value: "$(params.name) is of type $(tasks.get-type.results.type)"
```

separately, could you please add documentation updates as well to describe the change?

vincent-pli:
@jerop thanks for comments.
I agree with you it will introduce dependencies in `Parameters` (order of list/array in yaml is preserved and `Parameters` is Array). 
The subsequent `parameter` adopt the `parameter` defined previously for expression, I think it's make sense. Actually it's exactly what I want. if no such ability we have to defined multiple `Task` for `CEL` one by one, and use `$(task.result.xxx)` in the last `expression`, it's complex for `pipeline` designer and will cause bad performance.

Basically, it's what CEL expect to, the `cel.NewEnv` and `prg.Eval` is design for it. Seems `CEL` let user define the expression with variables(a express template) firstly then do compile and verification then parse it with runtime parameters:
```
out, details, err := prg.Eval(map[string]interface{}{
    "name": "/groups/acme.co/documents/secret-stuff",
    "group": "acme.co"})
fmt.Println(out) // 'true'
```

So if we substitute expression before enter `CEL`,  I believe it could work but introduce complex and not necessary.
I think this project is not only serve for `when express` but for more cases and the new introduced functionality will not hurt the `when express` scenario.

The user should use quotation(single or double) marks for static string, I means "types" and "name", without quotation will treat as variable.

jerop:
@vincent-pli thanks for the detailed response!

> order of list/array in yaml is preserved

that's right, the parameters order is preserved since https://github.com/tektoncd/pipeline/pull/3024 (will need to add docs for that though) 

> it's what CEL expect to, the `cel.NewEnv` and `prg.Eval` is design for it

makes sense, we can try it out and iterate as needed given it's experimental

> the user should use quotation(single or double) marks for static string, I means "types" and "name", without quotation will treat as variable

please add documentation for this in the PR, I want to make sure that static vs variable strings is clear to users

vincent-pli:
@jerop thanks
Add some description in the `README`, please take a look.

jerop:
@vincent-pli thank you for adding the documentation and patience on this issue! 


1. **order of parameters/expressions**
there's an assumption here that the parameters/expressions would be evaluated in the order in which they were declared, but @bobcatfish clarified that while the order was made deterministic in https://github.com/tektoncd/pipeline/pull/3024 it's not necessarily the order in which the user declared them

1. **static vs variable strings**
the added docs are a great improvement but the distinction between the static and variable strings in the parameters/expressions isn't clear to me still  

1. **use cases**
thanks for providing the simple example -- it would really help if you could also provide a realworld CI/CD use case that would help us understand the issue better and find the best solution for it 

wondering what the other CEL owners think about this, /cc @bobcatfish @ImJasonH @pritidesai @bigkevmcd

vincent-pli:
@jerop 
For the order things, if we insist to it should no dependency between `param`s, we can turn to https://github.com/tektoncd/community/issues/403, but will lost some flexibility. @bobcatfish

PR #725, opened 2021-03-12
[task-loops] update pipelines version

pritidesai:
# Changes

Use the latest pipelines version `0.22.0` while moving away from `v0.20.0`.

Up until `v0.20.0`, pipelines had `pkg/system` to retrieve the namespace which was dropped in favor of `knative.dev/pkg/system`, see PR: https://github.com/tektoncd/pipeline/pull/3177

Applying changes to use `knative.dev/pkg/system`.

Finally, changes include updating dependencies using `hack/update-deps.sh`.

# Submitter Checklist

review them:

for more details._


jerop:
@pritidesai the `cannot find package "."` is caused by missing files in `vendor/`

it seems that this is caused by `gitignore` - thought that `!vendor/` would have caused them to be included but that's not the case and it's unclear why - so i force added the vendor directory in https://github.com/tektoncd/experimental/pull/784 and it fixed the issue 

shout out to @wlynch and @sbwsg for the fix! thank you so much 😀

jerop:
instead of the above fix, i deleted the vendor files in https://github.com/tektoncd/experimental/pull/795 to get rid of the dependency issues

PR #729, opened 2021-04-12
Adding experimental interceptors

jmcshane:
# Changes

Creating additional ClusterInterceptors using the interceptor-sdk in tektoncd/triggers#1042

This PR cannot be merged until the upstream one is released

# Submitter Checklist

review them:

for more details._


afrittoli:
Thanks for this! I think it's a great idea to start having this in experimental.
What would be really useful (as follow-ups to this PR):
- have some discussion with the triggers team / tekton community about what we want to do about custom interceptors. We have some now in `plumbing`, some new here, and I think it would be great to have mechanism to make it easy for Tekton users to discover and use them. We could have a repo for each, or a repo with all of them, we could document them in the official doc, provide config options in the operator to pull them in...
- setup nightly builds - to make it easier to install and maintain them. This should be pretty straight-forward, we can re-use the existing nightly-build infra

jmcshane:
One nice thing about #1042 is that it makes it easy to package up a subset of interceptors into a single deployment. We could have a single repo just for interceptors, but have those interceptors packaged into some sort of "groups" (as subdirectories of the `config/` directory). @wlynch discussed the idea of an extension "interface" in the last WG, and grouping interceptors around that concept may be helpful as well.

PR #730, opened 2021-04-19
[POC] Add support for `Run.RunSpec.Spec` to TaskLoop.

ScrapCodes:
This PR is not intended to be reviewed from the point of being merged immediately, as the corresponding changes to the `Tektoncd/pipeline` are still under discussion in TEP-0061 link https://github.com/tektoncd/community/pull/408.

This is available for both trying out the work done as part of TEP-0061 and also review the design of the feature itself. Are the APIs looking good?

e.g. Does it look wierd, when I call `Run.Spec.Spec.Spec` everywhere? Shall we think of renaming it to `Run.RunSpec.Spec.TaskSpec` for example.

 If we do so, what should be the future of `Run.RunSpec.Ref` ?

# Changes

# Submitter Checklist

review them:

for more details._


task-loops/pkg/reconciler/tasklooprun/tasklooprun.go
```diff
@@ -286,6 +296,15 @@ func (c *Reconciler) getTaskLoop(ctx context.Context, run *v1alpha1.Run) (*metav
 		}
 		taskLoopMeta = tl.ObjectMeta
 		taskLoopSpec = tl.Spec
+	} else if run.Spec.Spec != nil {
+		err := json.Unmarshal(run.Spec.Spec.Spec.Raw, &taskLoopSpec)
```
ScrapCodes, line 300: `run.Spec.Spec.Spec.Raw` Do you think this looks good?

PR #746, opened 2021-05-12
[notifiers] GitHub: Add support for Status API.

wlynch:
# Changes

Using the Check Run API requires a GitHub App, and access to App debug details
(i.e. webhook delivery statuses, etc) require admin priviledges on the owning
user/org. For Tekton dogfooding, this makes using GitHub Apps difficult because:

1. For an org-internal App, we would need to grant anyone that would
need to debug dogfooding (i.e. build captains) admin access to the
tektoncd org, or put the burden on tektoncd admins to investigate
if/when there are issues with webhook delivery.
2. For an app owned by a separate org, we would need to make the App
public, which could let any GitHub user install the App and
cause us to receive their webhooks.

As an alternative, this PR adds support to make the behavior configurable based
on the authentication creds provided to the controller -
If a GitHub App ID + Key are passed in the Check Run API will be used, else we
fall back to the Status API and assume an access token has been provided to the
controller.

# Submitter Checklist

review them:

for more details._


notifiers/github-app/README.md
```diff
@@ -32,3 +37,6 @@ For an example a full TaskRun with the initial annotations set, see
 | -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | GITHUB_APP_ID        | ID of your GitHub App. Can be found under https://github.com/settings/apps > Edit > General > About > App ID                                                           |
 | GITHUB_APP_KEY       | Path to the [private key of your GitHub App](https://docs.github.com/en/free-pro-team@latest/developers/apps/authenticating-with-github-apps#generating-a-private-key) |
+| GITHUB_TOKEN         | GitHub Personal Access Token (https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token). We strongly recommend using your own personal GitHub Account - use a bot user instead. |
```
dibyom, line 40:
```suggestion
| GITHUB_TOKEN         | GitHub Personal Access Token (https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token). We strongly recommend not using your own personal GitHub Account - use a bot user instead. |
```

dibyom (approved):

notifiers/github-app/pkg/controller/status.go
```diff
@@ -0,0 +1,67 @@
...
+		Description: description,
+		TargetURL:   github.String(dashboardURL(tr)),
+		Context:     github.String(tr.GetName()),
+	}
+	_, _, err = client.Repositories.CreateStatus(ctx, owner, repo, commit, status)
+	return err
+}
+
+func dashboardURL(tr *v1beta1.TaskRun) string {
+	// TODO: generalize host, object type.
```
dibyom, line 39: We could set an env var like DASHBOARD_HOST that this can function can use.

notifiers/github-app/README.md
wlynch: LOL. 🤣 

Good catch. :)

notifiers/github-app/pkg/controller/status.go
wlynch: Yup! That's the plan, but holding off for another PR for the time being.

PR #749, opened 2021-05-20
Bump plumbing to b245931

wlynch:
# Changes

Bump plumbing to https://github.com/tektoncd/plumbing/commit/b2459312cc6b5116a3fe728207b8f76234182260 to fix test script causing all tests to run.

Also disables go modules for webhooks-extension, since it is not
compatible and is failing tests.

# Submitter Checklist

review them:

- [-] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [-] Commit messages includes a project tag in the subject - e.g. [OCI], [hub], [results], [task-loops]

for more details._

### DIFF
--- a/webhooks-extension/test/presubmit-tests.sh
+++ b/webhooks-extension/test/presubmit-tests.sh
@@ -26,6 +26,9 @@
 export DISABLE_MD_LINTING=1
 export TEST_FOLDER=$(pwd)
 
+# This project still uses Godep - disable Go modules.
+export GO111MODULE=off
+
 source $(dirname $0)/../../vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh
 
 function get_yq() {
--- a/vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh
+++ b/vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh
@@ -53,7 +53,7 @@ GOFLAGS=${GOFLAGS:-}
 # Returns true if PR only contains the given file regexes.
 # Parameters: $1 - file regexes, space separated.
 function pr_only_contains() {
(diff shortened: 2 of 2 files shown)


webhooks-extension/test/presubmit-tests.sh
```diff
@@ -26,6 +26,9 @@
 export DISABLE_MD_LINTING=1
 export TEST_FOLDER=$(pwd)
 
+# TODO: move to Go modules.
```
bobcatfish, line 29: could you update this TODO with a bit more info about why this fails things and/or maybe link to an issue?

bobcatfish (commented):
/approve

webhooks-extension/test/presubmit-tests.sh
wlynch: Rephrased. It's not so much an issue moreso this package uses Godep, and as of Go 1.16 modules are on by default. This is really here for compatibility.

bobcatfish: i find that TODO type comments end up just kinda hanging around forever and often eventually lose meaning so i feel like if this is a TODO it would make sense to link to a github issue or something for context (e.g. like `TODO(#11111)`) or remove the "TODO" entirely and just leave something like what you just said, i.e. "this package uses Godep, and as of Go 1.16 modules are on by default"

> This is really here for compatibility.

I don't totally understand why you mean by that, compatibility with what? (maybe the comment could just provide more context)

wlynch: 🤦  My git push was stuck on ssh-agent password so it never pushed. *Now* it should be changed.

bobcatfish: ahhh haha sorry i had a feeling, should have messaged you

AlanGreene:
@wlynch fyi, I have a PR open to delete the webhooks-extension: https://github.com/tektoncd/experimental/pull/751

PR #754, opened 2021-06-02
[cloudevents] Add experimental controller for cloudevents

waveywaves:
# Changes

Currently in Tekton, sending cloudevents to the sink is done by the individual resource controllers (PipelineRun, TaskRun etc.) With this we would be extracting the cloudevents functionality to a controller of it's own thus reducing the coupling between resource controllers and emitting cloudevents. 

# Submitter Checklist

review them:

for more details._


cloudevents/README.md
```diff
@@ -0,0 +1,29 @@
...
+This will build and install the controller on your cluster, in the namespace
+`tekton-pipelines`.
+
+## Example
+
+## Uninstall
+
+```
+$ kubectl delete namespace tekton-pipelines
+namespace "tekton-pipelines" deleted
```
afrittoli, line 26: I think this is not a good idea. Tekton is installed in the `tekton-pipelines` namespace too, deleting the namespace is not a good option to uninstall.
```
ko delete -f config/
```

cloudevents/pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (t CDEventType) String() string {
+	return string(t)
+}
+
+type Reconciler struct {
+	cloudEventClient cloudevent.CEClient
+	tracker          tracker.Interface
+
+	taskRunLister listers.TaskRunLister
```
afrittoli, line 87: The `taskRunLister` is not used - and it's not passed to the `Reconciler` anyways when the `Reconciler` is created.

cloudevents/cmd/controller/main.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const pipelineRunControllerName = "cloudevents-pipelinerun-controller"
+
+func main() {
+	sharedmain.Main(pipelineRunControllerName, newPipelineRunController)
+}
+
+func newPipelineRunController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	c := &reconciler.Reconciler{}
```
afrittoli, line 36: The `cloudEventClient` is not passed to the `Reconciler` here?

cloudevents/pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	defer func() {
+		if pr.Status.StartTime == nil {
+			return
+		}
+		// Compute the time since the task started.
+		elapsed := time.Since(pr.Status.StartTime.Time)
+		// Snooze this resource until the timeout has elapsed.
+		c.snooze(pr, pr.GetTimeout(ctx)-elapsed)
```
afrittoli, line 135: The cloud event controller should not take care of implementing the timeout - that is done by the main controller.

afrittoli (commented):
Thanks for this @waveywaves !!
I left some comments on the code.
Could you add an OWNERS file to the folder? 
I think this is fine for an initial experiment - we'll need to start adding unit tests soon though!
We should be able to re-use a lot of ideas from pipeline own unit tests :)

cloudevents/config/controller.yaml
```diff
@@ -0,0 +1,237 @@
...
+  kind: ClusterRole
+  name: tekton-cloudevents-controller-cluster-access
+  apiGroup: rbac.authorization.k8s.io
+
+---
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging
```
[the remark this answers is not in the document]
waveywaves: I have moved it to the tekton-cloudevents namespace. It isn't conflicting now.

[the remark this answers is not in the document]
waveywaves: I tried changing the name and mounting it in the same location but seems that config-logging has to be the name for it.

cloudevents/config/controller.yaml
```diff
@@ -0,0 +1,237 @@
...
+metadata:
+  name: tekton-cloudevents-controller-cluster-access
+  labels:
+    app.kubernetes.io/component: tekton-cloudevents-controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-cloudevents
+rules:
+  # Controller needs cluster access to all Run CRs.
+  - apiGroups: ["tekton.dev"]
+    resources: ["runs"]
```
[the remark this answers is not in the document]
waveywaves: Updated to watch pipelineruns

cloudevents/pkg/reconciler/reconciler.go
waveywaves: removed it

cloudevents/cmd/controller/main.go
waveywaves: updated to consume the default cloudevent client injected in the context

cloudevents/pkg/reconciler/reconciler.go
waveywaves: removed the defer statement which takes care of the timeout

vdemeester (approved):

afrittoli (commented):
Thanks @waveywaves !!
Let's go ahead with this - I think next we'll need to:
- add a local config store so that we can configure the sink and do not depend on pipeline configs that we don't need here
- add tests to verify that we send events at the right times. It is going to be challenging to identify the right transitions - because, unlike the core controller, here we only read the status and do not mutate it. One possibility could be to look at k8s events as a source of truth. Another option could be to hold a state, which would enable plenty more features.

PR #755, opened 2021-06-02
[task-loop] Update 0.24.1

ScrapCodes:
# Changes

Task loop has not been updated to latest tektoncd version, since 0.21. In this PR, we are updating it to 0.24.1.

# Submitter Checklist

review them:

for more details._


ScrapCodes:
Interestingly, build and unit tests are passing locally. Any direction will be helpful.

pritidesai:
@ScrapCodes I tried updating it in the other PR #725 and ran into the same issues. I haven't been able to resolve it, will update you if I find something helpful.

jerop:
@ScrapCodes  the `cannot find package "."` dependency issue is caused by missing files in `vendor/`

it seems that this is caused by `gitignore` - thought that `!vendor/` would have caused them to be included but that's not the case and it's unclear why - so i force added the vendor directory in https://github.com/tektoncd/experimental/pull/784 and it fixed the issue 

shout out to @wlynch and @sbwsg for the fix! thank you so much 😀

jerop:
instead of the above fix, i deleted the vendor files in https://github.com/tektoncd/experimental/pull/795 to get rid of the dependency issues

PR #760, opened 2021-06-16
[cloudevents] use cde events for taskrun/pipelinerun start,finish

waveywaves:
# Changes

- extract reconciler/events/cloudevent library from tektoncd/pipeline to amke changes in the datatype which is being used in sending cloudevents
- replace existing TektonEventType based events with CDEEventType events
- Update TaskRun and PipelineRun to use Started and Finished events
- TaskRun and PipelineRun success/failure events don't exist yet so these have been replaced with cde "finished" events for time being

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
+++ b/cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
@@ -1,0 +1,162 @@
+package cloudevent
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	cdeevents "github.com/cdfoundation/sig-events/cde/sdk/go/pkg/cdf/events"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/reconciler/events/cloudevent"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+	"time"
+)
+
+// CEClient matches the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
+type CEClient cloudevents.Client
+
+// TektonCloudEventData type is used to marshal and unmarshal the payload of
+// a Tekton cloud event. It can include a TaskRun or a PipelineRun
+type TektonCloudEventData struct {
+	TaskRun     *v1beta1.TaskRun     `json:"taskRun,omitempty"`
+	PipelineRun *v1beta1.PipelineRun `json:"pipelineRun,omitempty"`
+}
+
+// newTektonCloudEventData returns a new instance of TektonCloudEventData
+func newTektonCloudEventData(runObject objectWithCondition) TektonCloudEventData {
+	tektonCloudEventData := TektonCloudEventData{}
+	switch v := runObject.(type) {
+	case *v1beta1.TaskRun:
+		tektonCloudEventData.TaskRun = v
+	case *v1beta1.PipelineRun:
+		tektonCloudEventData.PipelineRun = v
+	}
+	return tektonCloudEventData
+}
+
+func getEventType(runObject objectWithCondition) (*cdeevents.CDEventType, error) {
+	c := runObject.GetStatusCondition().GetCondition(apis.ConditionSucceeded)
+	if c == nil {
+		return nil, fmt.Errorf("no condition for ConditionSucceeded in %T", runObject)
+	}
+	var eventType cdeevents.CDEventType
+	switch {
+	// TODO : Update TaskRun/PipelineRun events from finished to success/fa
+	case c.IsUnknown():
+		switch runObject.(type) {
+		case *v1beta1.TaskRun:
+			switch c.Reason {
+			case v1beta1.TaskRunReasonStarted.String():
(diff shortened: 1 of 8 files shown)


cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -0,0 +1,162 @@
...
+	case c.IsUnknown():
+		switch runObject.(type) {
+		case *v1beta1.TaskRun:
+			switch c.Reason {
+			case v1beta1.TaskRunReasonStarted.String():
+				eventType = cdeevents.TaskRunStartedEventV1
+				//case v1beta1.TaskRunReasonRunning.String():
+				//	eventType = TaskRunRunningEventV1
+				//default:
+				//	eventType = TaskRunUnknownEventV1
```
afrittoli, line 59: TaskRuns do not have a queued event, I'm not sure what is the reason for that?
NIT: could we delete the commented code?

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+		case *v1beta1.PipelineRun:
+			switch c.Reason {
+			case v1beta1.PipelineRunReasonStarted.String():
+				eventType = cdeevents.PipelineRunStartedEventV1
+				//case v1beta1.PipelineRunReasonRunning.String():
+				//	eventType = PipelineRunRunningEventV1
+				//default:
+				//	eventType = PipelineRunUnknownEventV1
+			}
```
afrittoli, line 69: Since the controller is notified here when the `PipelineRun` is created (as opposed to when it's picked up by the other controller), we could in fact send the "Queued" event here, and send the "Started" event for the Tekton "Running" condition, which corresponds to the first TaskRun being created. wdyt?

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -0,0 +1,162 @@
...
+	// TODO: SelfLink is deprecated https://github.com/tektoncd/pipeline/issues/2676
+	source := runObject.GetObjectMeta().GetSelfLink()
+	if source == "" {
+		gvk := runObject.GetObjectKind().GroupVersionKind()
+		source = fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s/%s",
+			gvk.Group,
+			gvk.Version,
+			runObject.GetObjectMeta().GetNamespace(),
+			gvk.Kind,
+			runObject.GetObjectMeta().GetName())
```
afrittoli, line 106: Uhm, this does not seem to work correctly, I wonder if it's broken in pipeline too?

afrittoli: I get `/apis///namespaces/default//guarded-pr-tmxzd`, so a few fields are missing

afrittoli (commented):
Thanks for this, it's definitely in the right direction, and I was able to get some events generated. The folder structure is the same we have in pipeline, which I think might not be the best one for this controller, but we can shuffle things around later.
I'll add a commit with some unit tests on top before we merge this.

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// eventForObjectWithCondition creates a new event based for a objectWithCondition,
+// or return an error if not possible.
+func eventForObjectWithCondition(runObject objectWithCondition) (*cloudevents.Event, error) {
+	event := cloudevents.NewEvent()
+	event.SetID(uuid.New().String())
+	event.SetSubject(runObject.GetObjectMeta().GetName())
+	// TODO: SelfLink is deprecated https://github.com/tektoncd/pipeline/issues/2676
+	source := runObject.GetObjectMeta().GetSelfLink()
```
afrittoli, line 98: you should use the functions from the CDE SDK here

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
waveywaves: +1

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
waveywaves: initially tried out the types for prototyping purposes

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
waveywaves: That association makes a lot of sense
Started -> CDE Queued
Running -> CDE Started
considering there is no "Running" in the CDE events Started means that it has started Running.

cloudevents/pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -4,83 +4,99 @@ import (
...
 		case *v1beta1.TaskRun:
-			eventType = cdeevents.TaskRunFinishedEventV1 //TaskRunFailedEventV1
+			eventType.Type = cdeevents.TaskRunFinishedEventV1 //TaskRunFailedEventV1
 		case *v1beta1.PipelineRun:
-			eventType = cdeevents.PipelineRunFinishedEventV1 //PipelineRunFailedEventV1
+			eventType.Type = cdeevents.PipelineRunFinishedEventV1 //PipelineRunFailedEventV1
 		}
-	case c.IsTrue():
+	case c.IsFalse():
+		eventType.Status = StatusError
```
waveywaves, line 94: :sparkles:

PR #763, opened 2021-06-18
[cloudevents] Add cache for tracking state of cloudevents

waveywaves:
# Changes

- Inject LRU Cache to the context.Context
- Update cache if event is sent.
- Disallow sending duplicate events by checking the cache

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/cloudevents/config/config-defaults.yaml
+++ b/cloudevents/config/config-defaults.yaml
@@ -21,7 +21,6 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-cloudevents
 data:
-  default-cloud-events-sink: http://sockeye.default.10.108.51.172.nip.io/
   _example: |
     ################################
     #                              #
--- a/cloudevents/config/config-defaults.yaml
+++ b/cloudevents/config/config-defaults.yaml
@@ -21,6 +21,7 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-cloudevents
 data:
+  default-cloud-events-sink: http://sockeye.default.10.108.51.172.nip.io/
   _example: |
     ################################
     #                              #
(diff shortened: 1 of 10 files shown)


cloudevents/config/config-defaults.yaml
```diff
@@ -21,6 +21,7 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-cloudevents
 data:
+  default-cloud-events-sink: http://sockeye.default.10.108.51.172.nip.io/
```
afrittoli, line 24: Please let's not add this default in the repo 🙏

waveywaves: oh my bad, I forgot I added this, I'l remove it :bow:

cloudevents/pkg/reconciler/events/cache/cache.go
```diff
@@ -0,0 +1,26 @@
...
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+)
+
+// AddEventSentToCache adds the particular object to cache marking it as sent
+func AddEventSentToCache(ctx context.Context, event *cloudevents.Event) error {
+	cacheClient := Get(ctx)
+	if cacheClient == nil {
+		return errors.New("cache client is nil")
+	}
+	cacheClient.Add(event.String(), nil)
```
afrittoli, line 15: Thanks for adding this. 

I think we'll have to use something different for the key though, as duplicate events might not be exactly the same.
Even with the exact same pipelinerun, timestamps will be different. 

We need to define a function that extracts the part of an event that we consider enough to define two events as the same, and use that as a key.  Probably a combination for event type, resource type and name should be enough to send each event type once for each resource.

In future we might want to send more events, or refresher kind of events, when something changes in the pipeline run but the condition stays the same. We could use the same kind of conditions that are used in the cloud events controller in pipeline today.

afrittoli: Fixed

cloudevents/config/config-defaults.yaml
afrittoli: fixed

afrittoli (approved):

afrittoli (commented):
/lgtm

PR #768, opened 2021-06-25
[github notifier] Short circuit if resources do not contain matching annotations.

wlynch:
# Changes

Previously we would attempt to process the trigger, even if we didn't
find the information we needed, which would cause reconcile loops since
we'd report error. This changes the behavior to only process resources
that contain the GitHub annotations.

# Submitter Checklist

review them:

for more details._

/kind bug


dibyom (approved):
/lgtm

PR #784, opened 2021-08-26
[celeval] CEL Custom Tasks CRD

### LINKED ISSUE #716: Add ability to parse variables in expression from context for CEL

jerop:
# Changes

We introduced CEL Custom Tasks to experiment with using an expression language with Tekton Pipelines.

Given feedback from the past several months of usage, we have identified three main current challenges:
- CEL custom tasks do not take variables for the CEL environment. As such, users cannot evaluate CEL expressions given specific variables or in specific context. For example, as described in https://github.com/tektoncd/experimental/issues/716 and https://github.com/tektoncd/community/issues/403, a user needed to declare runtime storage variables in the CEL environment.
- CEL custom tasks are not a CRD thus making them unreusable across different Runs and PipelineRuns. Read more in https://github.com/tektoncd/community/pull/314#pullrequestreview-584219301.
- CEL custom tasks take the CEL expressions through Parameters which is misleading to some users. Read more in https://github.com/tektoncd/community/pull/314#pullrequestreview-584219301.

To address the above challenges, this change introduces a CRD for CEL Custom Tasks, which takes CEL expressions and CEL environment variables. With this change:
- CEL custom tasks now take variables for the CEL environment
- CEL custom tasks are reusable across different Runs and PipelineRuns
- CEL custom tasks take expressions through its own field

Closes tektoncd/community#403
Closes #716

# Submitter Checklist

review them:

/cc @vincent-pli @bobcatfish @imjasonh @pritidesai @bigkevmcd @vdemeester


jerop:
@vincent-pli @bobcatfish @imjasonh @pritidesai @bigkevmcd @vdemeester is ready for review, please take a look :)

cel/config/300-cel.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: cels.custom.tekton.dev
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-cel
+    pipeline.tekton.dev/release: "devel"
+    version: "devel"
```
wlynch, line 23: should we just use app.kubernetes.io/version instead?

cel/config/500-webhook.yaml
```diff
@@ -0,0 +1,115 @@
...
+  name: tekton-cel-webhook
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/name: webhook
+    app.kubernetes.io/component: webhook
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-cel
+    app.kubernetes.io/version: "devel"
+    pipeline.tekton.dev/release: "devel"
+    version: "devel"
```
wlynch, line 27: I noticed the usage of version/release labels is inconsistent - perhaps we should pull common labels like this into a kustomize config?

cel/pkg/apis/cel/v1alpha1/cel_types.go
```diff
@@ -0,0 +1,96 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Cel evaluates a CEL expression with given environment variables
+// +k8s:openapi-gen=true
+type Cel struct {
```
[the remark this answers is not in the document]
wlynch: It would probably also make sense to modify the CRD to match as well.

cel/test/e2e-cels.sh
```diff
@@ -0,0 +1,54 @@
...
+# Helper functions for E2E tests.
+
+source $(git rev-parse --show-toplevel)/vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
+
+function install_cel_crd() {
+  echo ">> Deploying Cel custom task"
+  ko resolve -f config/ \
+      | sed -e 's%"level": "info"%"level": "debug"%' \
+      | sed -e 's%loglevel.controller: "info"%loglevel.controller: "debug"%' \
+      | sed -e 's%loglevel.webhook: "info"%loglevel.webhook: "debug"%' \
```
wlynch, line 26: optional: instead of doing a sed replacement over all configs, an alternative that can be used is using a kustomize overlay for test deployments (see https://github.com/tektoncd/results/tree/main/test/e2e/kustomize an example)

cel/test/e2e-cels.sh
```diff
@@ -0,0 +1,54 @@
...
+  echo ">> Uninstalling Cel custom task"
+  ko delete --ignore-not-found=true -f ${REPO_ROOT_DIR}/cel/config/
+
+  # Make sure that everything is cleaned up in the current namespace.
+  delete_cel_resources
+}
+
+function delete_cel_resources() {
+  for res in run; do
+    kubectl delete --ignore-not-found=true ${res}.tekton.dev --all
```
wlynch, line 49: This has the side effect of deleting all custom task runs, not just CEL custom tasks. Is this okay, or do we need to add another condition to this delete?

cel/test/build_logs.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/pkg/test/logging"
+)
+
+// CollectPodLogs will get the logs for all containers in a Pod
+func CollectPodLogs(ctx context.Context, c *clients, podName, namespace string, logf logging.FormatLogger) {
```
wlynch, line 32: Since this is only for tests, does this need to be exported?

cel/test/init_test.go
```diff
@@ -0,0 +1,240 @@
...
+	"sync"
+	"testing"
+)
+
+var initMetrics sync.Once
+var skipRootUserTests = false
+
+func init() {
+	flag.BoolVar(&skipRootUserTests, "skipRootUserTests", false, "Skip tests that require root user")
+}
```
wlynch, line 48: Might have missed it, but is this being used?

cel/test/wait.go
```diff
@@ -0,0 +1,103 @@
+/*
```
wlynch, line 1: This file looks largely lifted from https://github.com/tektoncd/pipeline/blob/main/test/wait.go. This does seem like useful functionality we'd want to reuse across tests. I wonder if it would make sense to just depend on https://pkg.go.dev/github.com/tektoncd/pipeline/test directly instead of forking these.

wlynch (commented):
Looking good!

cel/.gitignore
```diff
@@ -53,6 +53,3 @@ cmd/*/kodata/source.tar.gz
 test/pullrequest/pullrequest-init
 /.bin/
 /bin/
-
-# allow vendor
-!vendor/
```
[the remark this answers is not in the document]
jerop: followed the pattern from task loops custom tasks -- will open a pr to remove them there as well

cel/pkg/apis/cel/v1alpha1/cel_types.go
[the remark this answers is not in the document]
jerop: Good catch, updated it CEL 👍🏾

cel/test/wait.go
jerop: added the missing function for polling for runs into that file in pipelines so that we can reuse it across different custom tasks -- https://github.com/tektoncd/pipeline/pull/4213

cel/pkg/reconciler/celrun/celrun.go
```diff
@@ -0,0 +1,278 @@
...
+		if err != nil {
+			return err
+		}
+		_, err = r.pipelineClientSet.TektonV1alpha1().Runs(run.Namespace).Patch(ctx, run.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+		return err
+	}
+	return nil
+}
+
+func getVariablesMap(variables []*v1beta1.Param) map[string]interface{} {
```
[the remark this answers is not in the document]
jerop: it has to be `map[string]interface{}` - otherwise the CEL expressions evaluation will fail with a syntax error

cel/pkg/reconciler/celrun/celrun.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return nil
+	}
+
+	storeCelSpec(status, celSpec)
+
+	propagateCelLabelsAndAnnotations(run, celMeta)
+
+	if err := celSpec.Validate(); err != nil {
+		logger.Errorf("Run %s/%s is invalid because of %s", run.Namespace, run.Name, err)
```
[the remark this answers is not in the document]
jerop: til, thanks!

cel/test/e2e-cels.sh
jerop: filtered them based on label -- thanks!

jerop:
thank you for the detailed review @wlynch! 😁

cel/README.md
```diff
@@ -5,25 +5,33 @@
 
 This is an experimental project that provides support for Common Expression Language (CEL) in Tekton Pipelines.
 The functionality is provided by a controller that implements the Custom Task interface. Its use cases include 
-evaluating complex expressions to be used in [`WhenExpressions`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#guard-task-execution-using-whenexpressions)
-in subsequent `Tasks` to guard their execution. 
+evaluating complex expressions to be used in [`when` expressions][when-expressions] in subsequent `Tasks` to guard 
+their execution.
```
bobcatfish, line 9: in some ways this probably goes without saying (it's true for all the experimental custom tasks) but i think it might be worth adding a word of caution expressing that we do have a desire to make this a top level feature, but at that point we'll need to re-examine the design via TEP and it might change

cel/README.md
```diff
@@ -37,231 +45,351 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
 
-- [`apiVersion`][kubernetes-overview] - Specifies the API version, `tekton.dev/v1alpha1`
-- [`kind`][kubernetes-overview] - Identifies this resource object as a `Run` object
-- [`metadata`][kubernetes-overview] - Specifies the metadata that uniquely identifies the `Run`, such as a `name`
-- [`spec`][kubernetes-overview] - Specifies the configuration for the `Run`
-- [`ref`][kubernetes-overview] - Specifies the `CEL` `Custom Task`
-  - [`apiVersion`][kubernetes-overview] - Specifies the API version, `cel.tekton.dev/v1alpha1`
-  - [`kind`][kubernetes-overview] - Identifies this resource object as a `CEL` object
-- [`params`](#specifying-cel-expressions) - Specifies the CEL expressions to be evaluated as parameters
+### Configuring a `ELl`
```
bobcatfish, line 57: typo maybe?

cel/README.md
```diff
@@ -37,231 +45,351 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
+
+A `CEL` definition supports the following fields:
```
bobcatfish, line 59: im probably nitpicking but im wondering about naming the CRD `CEL` in that it feels strange to me to refer to `a CEL`

ultimately this seems to be a collection of CEL expressions, am i understanding that right? if so do we want to call it something like `CELExpressions`?

I think this also suggests it might be worthwhile to be more clear about defining exactly what the purpose and scope of this CRD is - if we make it really vague, there's no indication of what belongs in there and what doesn't.

these are probably more so considerations for later TEPs but if we can sort through some of it now it'll be less headache for users if we make changes

cel/README.md
```diff
@@ -37,231 +45,351 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
+              script: echo ISINT!
+```
+
+For more information about `Parameters`, read [specifying `Parameters`][specifying-parameters].
+
+### Specifying CEL environment variables
+
+The CEL variables to be declared in the environment before evaluation by `CEL` are specified in `variables` field.
+
+They are specified as `Parameters`, that is, they are made up of `name` and `value` pairs which are Strings.
```
bobcatfish, line 198: maybe im just misunderstanding the example following this, but I don't see any parameters?

cel/README.md
```diff
@@ -37,231 +45,351 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
+            - name: echo
+              image: ubuntu
+              script: echo ISINT!
+```
+
+For more information about `Parameters`, read [specifying `Parameters`][specifying-parameters].
+
+### Specifying CEL environment variables
+
+The CEL variables to be declared in the environment before evaluation by `CEL` are specified in `variables` field.
```
bobcatfish, line 196: is it possible to explain more here about what CEL environment variables are and why these need to be treated differently from other variables/parameters?

(i think it might be that im just not following the example)

cel/README.md
jerop: > im probably nitpicking but im wondering about naming the CRD `CEL` in that it feels strange to me to refer to `a CEL`

given the pattern we have with `Task` --> `TaskRun`, `Pipelines` --> `PipelineRun`, so i thought it may make sense to call this `CEL` since we've been referring to the execution as `CELRun`

but i'm open to exploring other names for the CRD - maybe `CELEvaluation`?

> ultimately this seems to be a collection of CEL expressions, am i understanding that right? if so do we want to call it something like `CELExpressions`?

it's a collection of CEL expressions and environment variables

> I think this also suggests it might be worthwhile to be more clear about defining exactly what the purpose and scope of this CRD is - if we make it really vague, there's no indication of what belongs in there and what doesn't.
> 

the purpose of the CRD is to specify the CEL expressions to be evaluated given some (if any) environment variables

> these are probably more so considerations for later TEPs but if we can sort through some of it now it'll be less headache for users if we make changes

agreed that it makes sense to address them now

cel/config/201-clusterrole.yaml
```diff
@@ -15,29 +15,86 @@
...
+metadata:
+  name: tekton-cel-leader-election
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-cel
+rules:
+  # We uses leases for leaderelection
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
```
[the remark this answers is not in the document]
wlynch: Reopening - doesn't look like a change was made yet

cel/README.md
jerop: yes, fixed it :)

cel/README.md
jerop: 👍🏾 added the disclaimer

cel/README.md
jerop: meant to say `similarly to parameters` but i realize i was bringing up implementation details here so clarified it with user facing details

cel/README.md
jerop: yes yes, makes sense -- added a description of the evaluation environment and how you can bind variables to it (with reference to the CEL documentation that goes into further detail

cel/config/500-controller.yaml
```diff
@@ -15,52 +15,49 @@
...
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: CONFIG_LEADERELECTION_NAME
+          value: config-leader-election
+        - name: CONFIG_LOGGING_NAME
+          value: config-logging
+        - name: CONFIG_OBSERVABILITY_NAME
+          value: config-observability
+        - name: METRICS_DOMAIN
+          value: tekton.dev/pipeline
```
wlynch, line 60: Probably want to change this - I assume this is some sort of bucketing mechanism so we want to make sure we're not group metrics alongside the main pipeline controller.

cel/pkg/reconciler/celrun/celrun.go
```diff
@@ -0,0 +1,283 @@
...
+		CELMeta = cs.ObjectMeta
+		CELSpec = cs.Spec
+	} else {
+		// Run does not require name but for CEL it does
+		run.Status.MarkRunFailed(celv1alpha1.CELRunReasonCouldntGetCEL.String(),
+			"Missing spec.ref.name for Run %s/%s",
+			run.Namespace, run.Name)
+		return nil, nil, fmt.Errorf("missing spec.ref.name for Run %s", fmt.Sprintf("%s/%s", run.Namespace, run.Name))
+	}
+	return &CELMeta, &CELSpec, nil
```
wlynch, line 217: readability: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

e.g. you could rewrite this as:

```go
if ref == nil || name == "" {
  return nil, nil, err
}
cs, err := Get()
if err != nil {
  return nil, nil, err
}
return meta, spec, nil
```

cel/pkg/reconciler/celrun/celrun.go
```diff
@@ -0,0 +1,283 @@
...
+			},
+		}
+		patch, err := json.Marshal(mergePatch)
+		if err != nil {
+			return err
+		}
+		_, err = r.pipelineClientSet.TektonV1alpha1().Runs(run.Namespace).Patch(ctx, run.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+		return err
+	}
+	return nil
```
wlynch, line 266: readability: Similar vein to https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow, typically it's recommended to front-load early exit conditions where you can, instead of wrapping them in an if.

e.g. 

```go
r := Get()
if reflect.DeepEqual(labels) && reflect.DeepEqual(annotations) {
  // no change, nothing else to do
  return nil
}
// continue on with patch
```

cel/README.md
```diff
@@ -37,231 +48,357 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
 
-The `CEL` `Custom Task` is defined in a `Run`, which supports the following fields:
+- Required:
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version, `custom.tekton.dev/v1alpha1`.
+  - [`kind`][kubernetes-overview] - Identifies this resource object as a `CEL` object.
+  - [`metadata`][kubernetes-overview] - Specifies the metadata that uniquely identifies the `CEL`, such as a `name`.
+  - [`spec`][kubernetes-overview] - Specifies the configuration for the `CEL`.
+    - [`expressions`](#specifying-cel-expressions) - Specifies the CEL expressions to be evaluated
+- Optional:
+    - [`variables`](#specifying-cel-environment-variables) - Specifies the CEL environment variables
```
bobcatfish, line 71: nit pick: i think it would be clearer to list variables directly under spec e.g. something like:

```
    - [`expressions`](#specifying-cel-expressions) - Specifies the CEL expressions to be evaluated
    - [`variables`](#specifying-cel-environment-variables) -  (optional) Specifies the CEL environment variables 
```

with the current version its not obvious that variables is under spec? just my opinion tho XD

cel/README.md
```diff
@@ -37,231 +48,357 @@ kubectl apply --filename https://storage.cloud.google.com/tekton-releases-nightl
...
+apiVersion: custom.tekton.dev/v1alpha1
+kind: CEL
+metadata:
+  name: get-sev
+spec:
+  variables:
+    - name: priority
+      value: "high"
+    - name: alert_enable
+      value: "true"
```
bobcatfish, line 214: does anything special happen with the values of these variables? i assume they are themselves evaluated? e.g. looking at https://github.com/tektoncd/experimental/issues/716 it looks like the values of these variables could themselves be expressions, is that right?

if so i think it might be worth explaining that a bit in the docs - but also id like to understand more about what the line is between variables and expressions  - e.g. could i use the body of the expression below as the value of one of these variables? `"priority in ['high', 'normal'] ? 'sev-1' : 'sev-2'"` (and do we then hit the same problem of referring to other variables within these variables that was in the original issue?)

cel/README.md
jerop: makes sense, fixed it :)

cel/README.md
jerop: no, the values of the variables are not evaluated -- they just get defined in the evaluation environment as context variables

in https://github.com/tektoncd/experimental/issues/716, the author had requested that each expression is evaluated then set as an environment variable 

```yaml
params:
- name: name
  value: "'hello'"
- name: types
  value: "type(name)"
```

but I had concerns about that approach discussed in https://github.com/tektoncd/experimental/pull/717#pullrequestreview-606410727 and https://github.com/tektoncd/experimental/pull/717#issuecomment-817918600

here, we are explicit in specifying the variables and expressions separately -- so the example above would be written as such instead:

```yaml
variables:
- name: name
  value: "hello"
expressions:
- name: types
  value: "type(name)"
```

if needed, a user can reuse the result of a previous CELRun as a variable in a subsequent CELRun

bobcatfish: ah okay i think i finally understand - I was thinking of `variables` behaving the same way as `params` do today, but actually `variables` will be evaluated as CEL environment vars vs. `params`  which were just string replacements 👍

bobcatfish:
The last thing on my mind is renaming from `CEL` to something a bit more descriptive but other than that 👍  !

/approve

cel/test/init_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+}
+
+func initializeLogsAndMetrics(t *testing.T) {
+	initMetrics.Do(func() {
+		flag.Parse()
+		flag.Set("alsologtostderr", "true")
+		logging.InitializeLogger()
+
+		//if knativetest.Flags.EmitMetrics {
```
wlynch, line 113: remove?

wlynch (commented):
Aside from the RBAC feedback, everything else is minor nits. Should be LGTM after this!

jerop:
hoping to revisit this some time 🤞🏾 

closing for now

PR #795, opened 2021-09-03
[task-loops] remove vendor files and use go modules

jerop:
# Changes

today, task-loops project has its vendor files causing dependency
management issues as seen in:
- https://github.com/tektoncd/experimental/pull/755
- https://github.com/tektoncd/experimental/pull/725

in this change, we:
- remove those vendored files so that the project relies on go modules
- fix gitignore such that it stops allowing vendor, which will make it
easier to maintain the project (such as reviewing changes)

# Submitter Checklist

review them:


ScrapCodes:
Thanks @jerop , 
why did we had vendor in the first place?
Do we need to run code gen in the future? a doc?

P.S. `+0 −1,252,929 `  my browser crashed 🙃

afrittoli:
> Thanks @jerop ,
> why did we had vendor in the first place?
> Do we need to run code gen in the future? a doc?
> 
> P.S. `+0 −1,252,929 ` my browser crashed 🙃

This project was created following the structure used in `tektoncd/pipeline` - which includes `vendor`.
I believe that is because, even if we use go modules, the local `vendor` folder provides a local cache of the required modules which might be handy for CI / local dev - @vdemeester is that right?

The modules have to be downloaded anyways as part of the git-clone, so perhaps having a local `vendor` folder does not provide such a great benefit? 

The bad news is that even if we remove `vendor` now the size of a full clone won't change...

vdemeester:
> This project was created following the structure used in `tektoncd/pipeline` - which includes `vendor`.
> I believe that is because, even if we use go modules, the local `vendor` folder provides a local cache of the required modules which might be handy for CI / local dev - @vdemeester is that right?

For `tektoncd/pipeline` and other graduated tekton project, yes, that's one of the reason (CI, local/disconnected dev).

The only problem with this, in `experimental`, is that there is a bunch of repository in `tektoncd/experimental`, and having a `vendor` for each is gonna make the repository grow to no end. Hence I suggested in another PR to not commit `vendor` in the experimental repository. 

> The modules have to be downloaded anyways as part of the git-clone, so perhaps having a local `vendor` folder does not provide such a great benefit?

Not sure I follow this. The local `vendor` folder has one benefit, you don't go through the dependency resolution and downloading them each as `go build` does (aka it's faster). Another, handy one, in `tektoncd/pipeline` (and other repos such as `triggers` and `operator`) is that we can patch a dependency for fixing a temporary failure 🙃 (cf the k8s 1.22 patches).

But for `experimental`, I think it's a question of balance 😅.

> The bad news is that even if we remove `vendor` now the size of a full clone won't change...

Indeed 😅

PR #807, opened 2021-10-21
Use cmpopts.IgnoreFields instead of IgnoreType.

wlynch:
# Changes

From
https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreTypes:

IgnoreTypes returns an Option that ignores all values assignable to certain types,
which are specified by passing in a value of each type.

This means when we pass in apis.Condition{}.LastTransitionTime.Inner.Time,
we are actually ignoring all time.Time, not just that particular field.
Similarly for apis.Condition{}.Message, we are ignoring all strings.

This change uses IgnoreFields to correct the behavior. Luckily, this doesn't
impact the tests in a meaningful way.

See https://github.com/tektoncd/triggers/issues/1245 for issue where this was discovered.

/kind bug

# Submitter Checklist

review them:

for more details._


wlynch:
ping @afrittoli for review for task-loops / pipeline-loops 🙏

wlynch:
/assign @dibyom 

Could I get global approval for experimental? 🙏

PR #850, opened 2022-03-18
[pipeline/trusted-resources]Verify task

Yongxuanzhang:
# Changes

This commit will introduce the validation of task, after verifying the
taskrun, the refered task will also be fetched and verified.
The second change is that webhook will make a copy of object and exclude
system populated fields. And verify on the copied object only on create
operation.

Before this commit, the webhook can only verify on taskspec.

# Submitter Checklist

review them:

for more details._


pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -178,9 +135,49 @@ func unmarshal(t *testing.T, buf []byte) (*v1beta1.TaskRun, []byte) {
 	if err != nil {
 		t.Fatalf("error decoding signature: %v", err)
 	}
+
+	delete(tr.ObjectMeta.Annotations, trustedtask.SignatureAnnotation)
```
wlynch, line 139: Why do we need to delete the annotation here? (wouldn't it already be removed prior to signing?)

Yongxuanzhang:
will divide this pr

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -119,3 +118,25 @@ func Sign(ctx context.Context, tr *v1beta1.TaskRun, signer sigstore.Signer, writ
...
+	}
+
+	signedBuf, err := yaml.Marshal(task)
+	if err != nil {
+		return err
+	}
+
+	_, err = writer.Write(signedBuf)
+	return err
+}
```
wlynch, line 142: You should be able to combine these by using the [object interface](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Object)

pipeline/trusted-resources/pkg/trustedtask/verify.go
```diff
@@ -0,0 +1,76 @@
...
+	if annotations[kmsAnnotation] != "" {
+		// Fetch key from kms.
+		return kms.Get(ctx, annotations[kmsAnnotation], crypto.SHA256)
+	} else {
+		cosignPublicKeypath := secretPath
+		// Overwrite the path if set in configmap.
+		cm, err := k8sclient.CoreV1().ConfigMaps(system.Namespace()).Get(ctx, signingConfigMap, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
```
wlynch, line 49: This snuck in the last PR - but we'll want to move this out of the request path. We generally shouldn't need to fetch this per request - we can update the reconciler state when the config map is modified. @chuangw6 can help you get this set up for the reconciler.

PR #851, opened 2022-03-18
[pipeline/trusted-resources]Change sign and verify target 

Yongxuanzhang:
# Changes

In this commit, the  webhook will make a copy of taskrun and exclude
system populated fields. And verify on the copied object only on create
operation.

Before this commit, the webhook can only verify on taskspec.

# Submitter Checklist

review them:

for more details._


pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -101,125 +93,58 @@ func (tr *TrustedTaskRun) VerifyTask(
...
-	ctx context.Context,
-	taskspec interface{},
-	verifier signature.Verifier,
-	signature []byte,
-) (errs *apis.FieldError) {
-	ts, err := json.Marshal(taskspec)
-	if err != nil {
-		return apis.ErrGeneric(err.Error(), "taskSpec")
-	}
+func copyTaskRun(in *TrustedTaskRun) (v1beta1.TaskRun, []byte, error) {
```
wlynch, line 117: We probably want to rename this (+ add func comments) this is really doing multiple things: 1) copying certain fields 2) removing certain data 3) extracting the signature.

1+2 probably makes sense together, 3 you may want to split out.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
```diff
@@ -168,18 +166,19 @@ func TestVerifyResources_TaskSpec(t *testing.T) {
...
 	}
 
 }
 
-func TestVerifyResources_OCIBundle(t *testing.T) {
+func TestVerifyTaskRun_OCIBundle(t *testing.T) {
 	ctx := logging.WithLogger(context.Background(), zaptest.NewLogger(t).Sugar())
 
+
+
```
wlynch, line 181: remove?

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -101,125 +93,58 @@ func (tr *TrustedTaskRun) VerifyTask(
...
+	c.SetName(in.Name)
+	c.SetGenerateName(in.GenerateName)
+	c.SetNamespace(in.Namespace)
 
-	h := sha256.New()
-	h.Write(ts)
-
-	if err := verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader(h.Sum(nil))); err != nil {
-		return apis.ErrGeneric(err.Error(), "taskSpec")
+	// Question: do we include labels?
```
wlynch, line 124: I think it would make sense to include them.

pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -91,94 +77,68 @@ var (
...
-	u, err := url.Parse(s.URL)
+	signature, err := base64.StdEncoding.DecodeString(tr.Annotations[trustedtask.SignatureAnnotation])
 	if err != nil {
-		t.Fatalf("error parsing url: %v", err)
+		t.Fatalf("error decoding signature: %v", err)
 	}
 
-	// Push OCI bundle
-	pushOCIImage(t, u, ts)
+	delete(tr.ObjectMeta.Annotations, trustedtask.SignatureAnnotation)
```
wlynch, line 117: copied from the other PR - why do we need this delete here? wouldn't we want to keep the TaskRun unaltered?

wlynch (changes requested):

pipeline/trusted-resources/pkg/trustedtask/verify.go
```diff
@@ -0,0 +1,76 @@
...
+	"crypto/sha256"
+
+	cosignsignature "github.com/sigstore/cosign/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/system"
+	"knative.dev/pkg/webhook/json"
```
wlynch, line 32: Any reason to use this over `encoding/json`?

Yongxuanzhang: it should be encoding/json, I guess probably my ide somehow auto import the wrong pkg

pipeline/trusted-resources/examples/1-test-taskrun.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
   name: tr-1
   namespace: tekton-trusted-resources
 spec:
+  serviceAccountName: "tekton-trusted-resources-webhook"
```
wlynch, line 11: I'd avoid including this or at least don't match the webhook SA - want to avoid confusion of whether this needs to match the webhook SA.

pipeline/trusted-resources/cmd/sign/sign_test.go
Yongxuanzhang: it is used before verification, so we need to remove the signature from the taskrun then we can verify the whole taskrun

pipeline/trusted-resources/examples/1-test-taskrun.yaml
Yongxuanzhang: sure!

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: yes thanks!

pipeline/trusted-resources/pkg/trustedtask/verify.go
```diff
@@ -0,0 +1,76 @@
...
+	annotations map[string]string,
+	k8sclient kubernetes.Interface,
+) (signature.Verifier, error) {
+	if annotations[kmsAnnotation] != "" {
+		// Fetch key from kms.
+		return kms.Get(ctx, annotations[kmsAnnotation], crypto.SHA256)
+	} else {
+		cosignPublicKeypath := secretPath
+		// Overwrite the path if set in configmap.
+		cm, err := k8sclient.CoreV1().ConfigMaps(system.Namespace()).Get(ctx, signingConfigMap, metav1.GetOptions{})
```
Yongxuanzhang, line 46: another comment from previous pr, can we avoid calling this per request?

pipeline/trusted-resources/cmd/sign/sign_test.go
wlynch: This might be a sign that we should change the API a bit -
It's a bit weird that we're marshaling a TaskRun struct to unmarshal it to marshal it again. 🤔  This probably means we should refactor the code a bit to separate out the signing and file writing bits.

We should also think about where the struct normalization is happening, and try to centralize it if possible (e.g. why are we removing SignatureAnnotation here, but not any of the server generated fields)?

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -101,125 +92,58 @@ func (tr *TrustedTaskRun) VerifyTask(
...
-		Namespace:          system.Namespace(),
-		ServiceAccountName: serviceAccountName,
-	})
-	if err != nil {
-		return apis.ErrGeneric(err.Error()).ViaKey(bundle)
+	c.Labels = make(map[string]string)
+	for k, v := range in.Labels {
+		c.Labels[k] = v
 	}
+	delete(c.ObjectMeta.Labels, "app.kubernetes.io/managed-by")
```
wlynch, line 135: Why exclude this label? 

We should add comments to each field explaining why we're excluding them + add it to the README.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -101,125 +92,58 @@ func (tr *TrustedTaskRun) VerifyTask(
...
+		c.Annotations[k] = v
 	}
+	delete(c.ObjectMeta.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
 
-	if err := verifier.VerifySignature(bytes.NewReader(signature), bytes.NewReader([]byte(digest.String()))); err != nil {
-		return apis.ErrGeneric(err.Error()).ViaKey(bundle)
+	c.Spec = in.Spec
+	if c.Spec.Timeout != nil {
+		c.Spec.Timeout = nil
 	}
```
wlynch, line 146: I'm very hesitant about doing this - we should generally respect fields set in the user spec.

pipeline/trusted-resources/pkg/trustedtask/verify.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil, err
+		}
+		if cm.Data["signing-secret-path"] != "" {
+			cosignPublicKeypath = cm.Data["signing-secret-path"]
+		}
+		return cosignsignature.LoadPublicKey(ctx, cosignPublicKeypath)
+	}
+}
+
+func VerifyInterface(
```
wlynch, line 59: missing godoc

wlynch (changes requested):

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: Oh yes. I thought this is generated by system. This param has default value 1 hour, so if user doesn't specify this field, we will get value at the webook. Not sure how to deal with this situation. Maybe we should force user to set this?

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: it is a label generated by other component not from the user. Will document this

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: I think one solution is to add the default value in our signing cmd, if this is not defined by user then we will add the default value and document this.

pipeline/trusted-resources/cmd/sign/sign_test.go
wlynch: > In sign cmd: it is marshal-> unmarshal, and for testing since we generate the object directly we only do unmarshal once. But separating them also makes sense..

I can see that - I think I might have been getting mixed up with the signing library vs the signing cli here. SGTM to leave as-is.

> yes, previously I added it in the copyTaskRun() but you suggest we rename the func or split out the signature extraction. Maybe I should add it in the func and rename it?

Yeah that sounds reasonable. That comment came from seeing a mismatch of the name from the actual behavior - e.g. we're saying something is a copy, but it's actually removing certain fields, etc. On the surface the behavior seems unexpected without the additional context.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
wlynch: 🤔  We should think more about how we should handle mutating webhooks - these are just values set by the pipelines admission controller at runtime, but there could be others that mutate other values.

Yongxuanzhang: - ah yes, the defaulting mutating webhook will call `SetDefaults()` and add some defaults data. So we can also call this function to add the defaults values. But this is not feasible since you mentioned there may be other mutating wh that can change the data at admission time.

- Another way is try to validate before the mutating?

Yongxuanzhang:
will create another pr

PR #852, opened 2022-03-23
[pipeline/trusted-resources]Verify task OCI bundle

Yongxuanzhang:
# Changes

This commit will focus on the verification of task oci bundle, since
taskrun and task in cluster may be mutated, we will handle oci bundle
first. 

# Submitter Checklist

review them:

for more details._


pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -85,141 +81,79 @@ func (tr *TrustedTaskRun) Validate(ctx context.Context) (errs *apis.FieldError)
...
 	k8sclient kubernetes.Interface,
 	tektonClient versioned.Interface,
 ) (errs *apis.FieldError) {
 	logger := logging.FromContext(ctx)
-	logger.Info("Verifying Resources")
+	logger.Info("Verifying TaskRun")
 
-	if tr.ObjectMeta.Annotations == nil {
-		return apis.ErrMissingField("annotations")
+	if tr.Spec.TaskRef == nil {
```
wlynch, line 92: nit: add a comment explaining why we're ignoring this.

pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -91,94 +77,55 @@ var (
...
 
 }
 
-func TestSign_OCIBundle(t *testing.T) {
-	ctx := context.Background()
-
-	sv, err := trustedtask.GetSignerVerifier(password)
-	if err != nil {
-		t.Fatalf("error get signerverifier: %v", err)
+func unmarshalTask(t *testing.T, buf []byte) (*v1beta1.Task, []byte) {
```
wlynch, line 103: Since this is doing more than just unmarshalling - add a comment explaining the behavior.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -64,8 +55,12 @@ var (
 	_ apis.Validatable = (*TrustedTaskRun)(nil)
 )
 
-// Validate the TaskRun is tampered or not.
+// Validate the Taskref referred task is tampered or not. Only support OCI task bundle now.
 func (tr *TrustedTaskRun) Validate(ctx context.Context) (errs *apis.FieldError) {
+	if !apis.IsInCreate(ctx) {
+		return nil
+	}
```
wlynch, line 62: I'd break this change into another PR - this is a different change than just signing Tasks, and we should make sure to add tests to validate this behavior

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -85,141 +81,79 @@ func (tr *TrustedTaskRun) Validate(ctx context.Context) (errs *apis.FieldError)
...
-	}
+	// TODO: figure out how to config this, one benefit of using this is that we
+	// don't have to configure the feature map.
+	cfg := config.NewStore(logger)
+	cfg.OnConfigChanged(&corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: config.GetFeatureFlagsConfigName()},
+		Data: map[string]string{
+			"enable-tekton-oci-bundles": "true",
+		},
+	})
```
wlynch, line 114: I'd avoid pulling in the Pipelines config map here - we should keep these separate - we may need to make changes to the upstream fetcher to decouple the TaskResolver from the configmap. For now, you can just construct the config manually and call ToContext without the store.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -85,141 +81,79 @@ func (tr *TrustedTaskRun) Validate(ctx context.Context) (errs *apis.FieldError)
...
-	bundle string,
-	verifier signature.Verifier,
-	signature []byte,
-	k8sclient kubernetes.Interface,
-) (errs *apis.FieldError) {
+// prepareTask will convert the taskobject to task and extract the signature.
+func prepareTask(t v1beta1.TaskObject) (v1beta1.Task, []byte, error) {
+	task := v1beta1.Task{}
+	task.TypeMeta = metav1.TypeMeta{
+		APIVersion: "tekton.dev/v1beta1",
```
wlynch, line 147: Nothing to do for this PR - but something to think about for the future is how we might support v1alpha types as well :thinking:

wlynch (changes requested):
Looking good!

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: Oh thanks! I realized that I added the comment but add it in a wrong place (line 96), will move it here.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: np!

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: sure, I think that is doable

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: I'm just curious if it is ok to do it in code or via yaml config.
I tried to apply the featuremap yaml to enable this but it doesn't work, is it because I don't include `configmap.Watcher` in current webhook, so it will not watch the config map change?

Yongxuanzhang: maybe one improvement of doing this in code is that I restore the old `cfg.FeatureFlags.EnableTektonOCIBundles` after I used it here, no matter it is true or false

wlynch: (the new revision of this basically does the same as what I wrote above, so LGTM)

pipeline/trusted-resources/pkg/trustedtask/verify_test.go
```diff
@@ -0,0 +1,93 @@
...
+		t.Fatalf("Failed to get signed task %v", err)
+	}
+
+	tamperedTask := signedTask.DeepCopy()
+	tamperedTask.Name = "tampered"
+
+	tcs := []struct {
+		name         string
+		task         *v1beta1.Task
+		hasSignature bool
```
wlynch, line 51: this isn't used.

pipeline/trusted-resources/pkg/trustedtask/verify.go
```diff
@@ -0,0 +1,78 @@
...
+	k8sclient kubernetes.Interface,
+) (signature.Verifier, error) {
+	if annotations[kmsAnnotation] != "" {
+		// Fetch key from kms.
+		return kms.Get(ctx, annotations[kmsAnnotation], crypto.SHA256)
+	} else {
+		cosignPublicKeypath := secretPath
+		// TODO: refactor this to avoid make request to api server each time.
+		// Overwrite the path if set in configmap.
+		cm, err := k8sclient.CoreV1().ConfigMaps(system.Namespace()).Get(ctx, signingConfigMap, metav1.GetOptions{})
```
wlynch, line 47: nit: move the const definition into this file where its used.

wlynch (changes requested):
just minor things left, otherwise lgtm!

pipeline/trusted-resources/pkg/trustedtask/verify_test.go
Yongxuanzhang: thx!

wlynch (approved):
/lgtm

Yongxuanzhang:
/lgtm
try to add this label :D

PR #853, opened 2022-03-25
[pipeline/trusted-resources]Verify task from cluster at taskrun

Yongxuanzhang:
# Changes

This commit focus on verify the task which is applied in cluster and
referred in taskrun. In this commit we will exclude system populated
fields and keep fields from users. The reason we verify the task in
taskrun validation not the task's validation is that there is a stage
between task creation and taskrun creation, so we need to make sure the
task is not tampered.

Before this commit we can only verify task from oci bundle.

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
+++ b/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
@@ -19,6 +19,7 @@ package trustedtask
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
 	"os"
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
@@ -51,7 +52,7 @@ var (
 	_ apis.Validatable = (*TrustedTaskRun)(nil)
 )
 
-// Validate the Taskref referred task is tampered or not. Only support OCI task bundle now.
+// Validate the Taskref referred task is tampered or not.
 func (tr *TrustedTaskRun) Validate(ctx context.Context) (errs *apis.FieldError) {
 	// TODO: validate only on create operation.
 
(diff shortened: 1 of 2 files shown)


pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -136,8 +131,27 @@ func prepareTask(t v1beta1.TaskObject) (v1beta1.Task, []byte, error) {
...
 		Kind:       "Task"}
 	task.Spec = t.TaskSpec()
-	task.ObjectMeta = t.TaskMetadata()
 
+	// exclude the fields populated by system.
+	task.Name = t.TaskMetadata().Name
+	task.GenerateName = t.TaskMetadata().GenerateName
+	task.Namespace = t.TaskMetadata().Namespace
+
+	task.Labels = make(map[string]string)
```
wlynch, line 140: In particular, do we need to be sensitive to nil vs empty maps here?

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -136,8 +131,27 @@ func prepareTask(t v1beta1.TaskObject) (v1beta1.Task, []byte, error) {
 		APIVersion: "tekton.dev/v1beta1",
 		Kind:       "Task"}
 	task.Spec = t.TaskSpec()
-	task.ObjectMeta = t.TaskMetadata()
 
+	// exclude the fields populated by system.
```
wlynch, line 135: Probably worth adding a test for this func - in particular to make sure we're covering cases like empty/missing data (e.g. metadata, maps, etc.)

wlynch (changes requested):
For the most part looks good! :D

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
Yongxuanzhang: you mean check if t.TaskMetadata().Labels is nil ?

Yongxuanzhang: I guess it is unnecessary?  because I don't use labels in my test and examples and they can work 😂  maybe I'm miss something

Yongxuanzhang: I changed it, not sure if that makes sense

wlynch (approved):
/lgtm

PR #855, opened 2022-03-29
[pipeline/trusted-resources]Add config map watcher

Yongxuanzhang:
# Changes

This commit implements the CosignConfigs which is used to watch the
config map changes, by this we can avoid making requests to api server
to fetch the config map.

# Submitter Checklist

review them:

for more details._


pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
```diff
@@ -430,18 +448,22 @@ func getSignerFromFile(t *testing.T, ctx context.Context, k8sclient kubernetes.I
...
+		Data: map[string]string{signingSecretPath: filepath.Join(tmpDir, "cosign.pub")},
 	}
-	if _, err := k8sclient.CoreV1().ConfigMaps(system.Namespace()).Create(ctx, cfg, metav1.CreateOptions{}); err != nil && !errors.IsAlreadyExists(err) {
-		t.Fatal(err)
+
+	if _, err := k8sclient.CoreV1().ConfigMaps(nameSpace).Update(ctx, cm, metav1.UpdateOptions{}); err != nil {
+		t.Errorf("error updating configmap: %v", err)
 	}
+	// It should be updated by then...
+	time.Sleep(100 * time.Millisecond)
```
wlynch, line 464: While this should usually be true, this has the potential to fail / not happen in time. If you use the other ToContext method to set the config directly in the context, it should give you what you want without needing to sleep.

wlynch (changes requested):

pipeline/trusted-resources/pkg/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	cm "knative.dev/pkg/configmap"
+)
+
+type Config struct {
```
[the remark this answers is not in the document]
wlynch: My bad - we should add markdown user documentation for this.

pipeline/trusted-resources/pkg/config/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Config holds the collection of configurations that we attach to contexts.
+// Configmap named with "config-trusted-resources" should include "signing-secret-path" as key in the data
+type Config struct {
+	Cosign CosignConfigs
+}
+
+// CosignConfigs holds the collection of cosign related configs
+type CosignConfigs struct {
+	CosignKey string
```
wlynch, line 34: What's the expected format of this?

pipeline/trusted-resources/pkg/config/store_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	cm.Data = map[string]string{}
+	cm.Data[cosignKey] = "newPath"
+
+	if _, err := fakekubeclient.CoreV1().ConfigMaps(ns).Update(ctx, cm, metav1.UpdateOptions{}); err != nil {
+		t.Errorf("error updating configmap: %v", err)
+	}
+
+	// It should be updated by then...
+	time.Sleep(100 * time.Millisecond)
```
wlynch, line 65: Want to highlight this again - sleeps are a common cause of flaky tests, and how long we can wait before this causes non-deterministic failures is always a balance. Is there any way we can synchronously invoke the update? (e.g. https://pkg.go.dev/knative.dev/pkg@v0.0.0-20220404063442-74799947727e/configmap#ManualWatcher.OnChange or something similar?)

pipeline/trusted-resources/pkg/config/store_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Test that the values are set!
+	if diff := cmp.Diff("newPath", store.Load().Cosign.CosignKey); diff != "" {
+		t.Error(diff)
+	}
+
+	cfg := FromContext(store.ToContext(ctx))
+	if diff := cmp.Diff("newPath", cfg.Cosign.CosignKey); diff != "" {
+		t.Error(diff)
+	}
```
wlynch, line 75: There's not too much you gain by diffing strings - IIUC `cmp` only diffs on the field level. I'd change this to a diff against the fully defined struct for the config.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
```diff
@@ -395,27 +382,21 @@ func getSignedTask(unsigned *v1beta1.Task, signer signature.Signer) (*v1beta1.Ta
 }
 
 // Generate key files to tmpdir, set configMap and return signer
-func getSignerFromFile(t *testing.T, ctx context.Context, k8sclient kubernetes.Interface) (signature.Signer, error) {
+func getSignerFromFile(t *testing.T, ctx *context.Context, k8sclient kubernetes.Interface) (signature.Signer, error) {
```
wlynch, line 385: Generally contexts are not passed around by pointer - you generally don't want a sub-process modifying the context of a parent.

Typically something like this would be done by initializing this at the beginning of a test, similar to how you are setting up the test logger. Can we pull this into a common "initialize context" func instead?

wlynch (changes requested):

pipeline/trusted-resources/pkg/config/config.go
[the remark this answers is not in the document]
Yongxuanzhang: you mean document it in the readme?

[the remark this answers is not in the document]
wlynch: yup! that works.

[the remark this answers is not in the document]
Yongxuanzhang: got it! Thanks!

pipeline/trusted-resources/pkg/config/store_test.go
Yongxuanzhang: Oh this can work, thanks!

Yongxuanzhang: But this is manually triggering the callbacks I guess? And we don't need to make request via fake k8s client. Is it ok to do so?

pipeline/trusted-resources/pkg/config/store_test.go
Yongxuanzhang: sure, thanks!

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
Yongxuanzhang: Got it, thanks!

pipeline/trusted-resources/pkg/config/config.go
Yongxuanzhang: So we don't need this? I can remove this and put the CosignKey into Config

pipeline/trusted-resources/pkg/config/store.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"context"
+
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/reconciler"
+)
+
+type cfgKey struct{}
+
+// +k8s:deepcopy-gen=false
```
XinruZhang, line 28: Just out of curiosity :) why do we need to explicitly disable the deepcopy-gen?

pipeline/trusted-resources/pkg/config/config.go
XinruZhang: I think Billy means we need to add a comment on this field :)

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
```diff
@@ -116,24 +115,15 @@ func TestVerifyTaskRun_TaskRun(t *testing.T) {
 func TestVerifyTaskRun_OCIBundle(t *testing.T) {
 	ctx := logging.WithLogger(context.Background(), zaptest.NewLogger(t).Sugar())
 
-	// this is needed to enable oci bundle
```
XinruZhang, line 119: May I ask why this configuration is not required any more?

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
```diff
@@ -394,9 +386,27 @@ func getSignedTask(unsigned *v1beta1.Task, signer signature.Signer) (*v1beta1.Ta
...
+func setupContext(ctx context.Context, k8sclient kubernetes.Interface, secretpath string) context.Context {
+	cmw := informer.NewInformedWatcher(k8sclient, system.Namespace())
+	store := config.NewConfigStore(logging.FromContext(ctx).Named("config-store"))
+	store.WatchConfigs(cmw)
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: nameSpace,
+			Name:      signingConfigMap,
+		},
+		Data: map[string]string{"signing-secret-path": secretpath},
```
XinruZhang, line 398: we've already defined it as a const variable `cosignKey`?

pipeline/trusted-resources/pkg/config/store.go
Yongxuanzhang: This is from https://github.com/tektoncd/chains/blob/5586b90ef1c108019a6a91d71156367a449f281e/pkg/config/store.go#L28-L31

My guess is we don't want to generate deepcopy for this

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
Yongxuanzhang: yes, thanks!

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun_test.go
Yongxuanzhang: it is configured in code already, so we don't need it here. This is a legacy code and not so related to this PR. 
https://github.com/tektoncd/experimental/blob/29ab1898266c8958f62724c45fb36af6224b2425/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go#L98-L100

XinruZhang: SG

pipeline/trusted-resources/pkg/config/config.go
```diff
@@ -31,23 +31,24 @@ type Config struct {
 
 const (
 	// cosignKey is the name of consign configmap
```
XinruZhang, line 33: nit: CosignKey, so as the following two const vars

pipeline/trusted-resources/pkg/config/config.go
```diff
@@ -31,23 +31,24 @@ type Config struct {
...
-	cosignKey = "signing-secret-path"
-
+	CosignKey = "signing-secret-path"
+	// secretPath is the default path of cosign public key
+	SecretPath = "/etc/signing-secrets/cosign.pub"
```
XinruZhang, line 36: super nit: DefaultSecretPath

pipeline/trusted-resources/pkg/config/store.go
Yongxuanzhang: and this kind of store type contains all pointer fields, maybe that's one reason?

pipeline/trusted-resources/pkg/config/config.go
Yongxuanzhang: thanks I'm revising these 😂

pipeline/trusted-resources/pkg/config/config.go
Yongxuanzhang: gotcha!

pipeline/trusted-resources/pkg/config/config.go
```diff
@@ -0,0 +1,62 @@
...
+// Config holds the collection of configurations that we attach to contexts.
+// Configmap named with "config-trusted-resources" should include "signing-secret-path" as key in the data
+type Config struct {
+	// CosignKey defines the value of the key in configmap data
+	CosignKey string
+}
+
+const (
+	// CosignPubKey is the value of the key in configmap data
+	CosignPubKey = "signing-secret-path"
```
XinruZhang, line 35: Want to make sure I understand correctly: do you mean CosignPubKey is the name of the key in configmap data here? other than the value, i.e. data[CosignPubKey]?

If so, the comment is kind of misleading :), you may want to update it.

Yongxuanzhang: yes, data[CosignPubKey] is correct.Maybe "name of the key"

XinruZhang:
/approve

I know you might already be working on this but I'll still point it out just in case XD - remember to merge the two commits, the others LGTM!

PR #868, opened 2022-04-20
[pipeline/trusted-resources]Add pipeline signing

Yongxuanzhang:
# Changes
This commit adds the signing for pipeline and merge it into one unified
function. Before this commit we only support signing for task.

# Submitter Checklist

review them:

for more details._


pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -77,44 +86,81 @@ var (
...
 	}
-
 }
 
 // unmarshalTask will get the task from buffer extract the signature.
-func unmarshalTask(t *testing.T, buf []byte) (*v1beta1.Task, []byte) {
-	task := &v1beta1.Task{}
-	if err := yaml.Unmarshal(buf, &task); err != nil {
-		t.Fatalf("error unmarshalling buffer: %v", err)
+func unmarshal(t *testing.T, buf []byte, kind string) (interface{}, []byte) {
```
Yongxuanzhang, line 126: maybe this func can be cleaner

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -71,36 +73,49 @@ func main() {
...
-	if err := yaml.Unmarshal(tsBuf, &ts); err != nil {
+	var crd resourcesemantics.GenericCRD
+	switch *kind {
+	case "Task":
+		crd = &v1beta1.Task{}
+	case "Pipeline":
+		crd = &v1beta1.Pipeline{}
+	}
+	if err := yaml.Unmarshal(tsBuf, &crd); err != nil {
 		log.Fatalf("error unmarshalling taskrun: %v", err)
```
XinruZhang, line 89: Is the error message correct? should be something like`error unmarshalling Task/Pipeline`?

Yongxuanzhang: ah thanks!

pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -77,44 +86,81 @@ var (
 	}
 )
 
-func TestSignTask(t *testing.T) {
+func TestSign(t *testing.T) {
```
XinruZhang, line 89: Do we need to add tests that fail to verify? i.e. modify it after signing?

pipeline/trusted-resources/cmd/sign/sign_test.go
XinruZhang: unmarshalCRD?

pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -77,44 +86,81 @@ var (
...
+			target, signature := unmarshal(t, signed, tc.resource.GetObjectKind().GroupVersionKind().Kind)
+			if err := trustedtask.VerifyInterface(ctx, target, sv, signature); err != nil {
+				t.Fatalf("VerifyTaskOCIBundle get error: %v", err)
+			}
+		})
 	}
-
 }
 
 // unmarshalTask will get the task from buffer extract the signature.
```
XinruZhang, line 125: remember to update the comment

pipeline/trusted-resources/cmd/sign/sign_test.go
Yongxuanzhang: That is a good point! 
But that case is covered in another folder, so here we just need to make sure the signed files can pass verification. That's my opinion

pipeline/trusted-resources/cmd/sign/sign_test.go
Yongxuanzhang: Thanks!

pipeline/trusted-resources/cmd/sign/sign_test.go
Yongxuanzhang: sure!

Yongxuanzhang: previously I return interface{}, but it make more sense to return crd at this moment

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -71,36 +73,49 @@ func main() {
...
+	case "Task":
+		if crd.(*v1beta1.Task).Annotations == nil {
+			crd.(*v1beta1.Task).Annotations = map[string]string{}
+		}
+		crd.(*v1beta1.Task).Annotations[trustedtask.SignatureAnnotation] = sig
+	case "Pipeline":
+		if crd.(*v1beta1.Pipeline).Annotations == nil {
+			crd.(*v1beta1.Pipeline).Annotations = map[string]string{}
+		}
+		crd.(*v1beta1.Pipeline).Annotations[trustedtask.SignatureAnnotation] = sig
```
wlynch, line 115: Instead of using `GenericCRD`, consider using [`metav1.Object`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Object) instead. And use the Get/SetAnnotations methods.

Then this looks like this for *any* type!

```suggestion
	a := o.GetAnnotations()
	a[trustedtask.SignatureAnnotation] = sig
	o.SetAnnotations(a)
```

wlynch (changes requested):

pipeline/trusted-resources/cmd/sign/sign_test.go
```diff
@@ -130,6 +176,19 @@ func getTask() *v1beta1.Task {
...
+	return &v1beta1.Pipeline{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "tekton.dev/v1beta1",
+			Kind:       "Pipeline"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-Pipeline",
+			Namespace: namespace,
+		},
+		Spec: *pipelineSpec,
+	}
```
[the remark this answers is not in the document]
Yongxuanzhang: Oh Thanks!!!

pipeline/trusted-resources/cmd/sign/main.go
Yongxuanzhang: yeah this works! Thanks!

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -71,36 +73,49 @@ func main() {
...
-		log.Fatalf("error unmarshalling taskrun: %v", err)
+	var crd resourcesemantics.GenericCRD
+	switch *kind {
+	case "Task":
+		crd = &v1beta1.Task{}
+	case "Pipeline":
+		crd = &v1beta1.Pipeline{}
+	}
+	if err := yaml.Unmarshal(tsBuf, &crd); err != nil {
+		log.Fatalf("error unmarshalling Task/Pipeline: %v", err)
```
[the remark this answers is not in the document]
Yongxuanzhang: Thanks! 😄 

This reminds me of the mutating webhook issue, I wonder what if we call `crd.SetDefaults(ctx)`, would it work to avoid possible issues for mutating webhooks? 

One reason the mutating wh will change the values is because tekton needs the resource to have default values for some fields but the users may forgot to configure them. So if before we sign it we call the SetDefaults directly, will that address this issue?

@wlynch what do you think?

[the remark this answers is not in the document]
Yongxuanzhang: Haven't applied the suggested changes yet but would like some feedback on this before that. 😄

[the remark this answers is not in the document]
XinruZhang: To add more context: Yongxuan and I discussed about this topic a littble bit, our thought is we define a new function called `FormatTask|Pipeline` to be called before customers sign their task/pipeline. In that function, we format the task/pipeline inputs from customers. For example, set default values before signing. 

Standardizing customer inputs before signing sounds pretty reasonable. Meanwhile, most importantly, we don't need to worry about what if Pipeline Mutation Webhooks modify a signed Task/Pipeline, which in the end is an issue about Mutation Webhook ordering.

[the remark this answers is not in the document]
wlynch: Hmmmm - not sure!

One thing I'd be worried about is the crd.SetDefaults probably wouldn't be sufficient to capture everything - the Tekton Pipeline webhook happens to correspond to the Defaultable type, but there might be other webhooks that do not conform to that at all. We probably want to sign either the initial unmodified input or the final fully resolved object - if we can't guarantee the fully resolved object my gut feeling would be to stick with the original object? wdyt?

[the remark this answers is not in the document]
Yongxuanzhang: yes that's one issue, there is the possibility that other mutating wh change the tasks. But if the users don't apply new mutating wh, then it will help to complete the yaml files I guess? I feel like it is a partial fix solution to this issue and better than nothing? What will be the cons of it?

So I think for current solution, we can document this can also inform the users of possible issues:
1) the non-standard usage of task or pipeline yaml files. The tekton mutating wh will populate default values for them and thus fail the verification. -> this may be fixed if we call SetDefaults before signing or either logging it to users
2) if the user deploy other mutating wh to change the task or pipeline, then this solution may fail. -> We have no idea what wh user will deploy, so maybe just document this. (it is very hard to meet all users requirements I guess T_T)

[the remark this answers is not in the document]
XinruZhang: Exactly, we did touch the topic you mentioned. The main concern would be what exactly Pipeline webhooks modify.
1.@yongxuangzhang talked to Christie about this issue weeks ago, Christie's suggestion was we can avoid modifying stuff in webhooks. Maybe we need to bring this up in the community.
2.@yongxuanzhang also found that the SetDefault modification is inevasible, so we start to think what's the best solution.

Verification webhooks are invoked after mutation webhooks, so there's not an easy way to verify signatures of Task/Pipeline API objects that are created through K8s API Server. Wdyt?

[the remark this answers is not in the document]
Yongxuanzhang: I added a flag and users can choose whether use it or not. 😄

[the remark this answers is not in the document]
wlynch: sgtm. let's hold off on that for another PR though? (or at the very least update this PR to include the changes?)

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -71,40 +75,50 @@ func main() {
...
+	// get annotation
+	a := o.GetAnnotations()
+	if a == nil {
+		a = map[string]string{}
 	}
-	task.Annotations[trustedtask.SignatureAnnotation] = sig
 
-	signedBuf, err := yaml.Marshal(task)
+	// add signature
+	sig, err := trustedtask.SignInterface(signer, o)
```
wlynch, line 115: check error

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -32,18 +32,22 @@ import (
...
-	cosignKey  = flag.String("ck", "", "cosign private key path")
-	kmsKey     = flag.String("kms", "", "kms key path")
-	taskFile   = flag.String("ts", "", "YAML file path for tekton task")
-	targetDir  = flag.String("td", "", "Dir to save the signed files")
-	targetFile = flag.String("tf", "signed.yaml", "Filename of the signed file")
+	cosignKey    = flag.String("ck", "", "cosign private key path")
+	kmsKey       = flag.String("kms", "", "kms key path")
+	resourceFile = flag.String("rf", "", "YAML file path for tekton resources")
+	kind         = flag.String("kd", "Task", "The kind of the signed object")
+	format       = flag.Bool("fmt", true, "Whether we add missing values for the CRD")
```
wlynch, line 45: nit: I'd be more descriptive here - e.g. how would a user who's just using the tool without looking at the code would understand what will happen?

- is this just format only changes (i.e. fixing tabs/spaces?)
- what missing values are we adding?

```suggestion
	format       = flag.Bool("default", true, "Whether we add Tekton default values to the CRD before signing")
```

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -32,18 +32,22 @@ import (
...
 var (
-	cosignKey  = flag.String("ck", "", "cosign private key path")
-	kmsKey     = flag.String("kms", "", "kms key path")
-	taskFile   = flag.String("ts", "", "YAML file path for tekton task")
-	targetDir  = flag.String("td", "", "Dir to save the signed files")
-	targetFile = flag.String("tf", "signed.yaml", "Filename of the signed file")
+	cosignKey    = flag.String("ck", "", "cosign private key path")
+	kmsKey       = flag.String("kms", "", "kms key path")
+	resourceFile = flag.String("rf", "", "YAML file path for tekton resources")
+	kind         = flag.String("kd", "Task", "The kind of the signed object")
```
wlynch, line 44:
```suggestion
	kind         = flag.String("kd", "Task", "The kind of the signed object. Supported values: [Task, Pipeline]")
```

(we may also want to make this case insensitive in another PR)

pipeline/trusted-resources/cmd/sign/sign_test.go
XinruZhang: SG!

pipeline/trusted-resources/cmd/sign/main.go
XinruZhang: +1 for docs

pipeline/trusted-resources/cmd/sign/main.go
```diff
@@ -32,18 +32,22 @@ import (
...
+	format       = flag.Bool("fmt", true, "Whether we add missing values for the CRD")
+	targetDir    = flag.String("td", "", "Dir to save the signed files")
+	targetFile   = flag.String("tf", "signed.yaml", "Filename of the signed file")
 )
 
-// This is a demo of how to generate signed task files
+// This is a demo of how to generate signed task or pipeline files
 func main() {
```
XinruZhang, line 51: Super nit: the current code structure is okay for now, but if we want to move it out from the experimental repo in the future, we'd better to keep the implementation details outside of the main function, for example:

```
signer, err := getSigner(...)
crd, err := getCRD(resourceFile)
formatCRD(&crd)
signCRD(...)
```

XinruZhang (commented):
/lgtm

pipeline/trusted-resources/cmd/sign/main.go
Yongxuanzhang: thanks!

pipeline/trusted-resources/cmd/sign/main.go
Yongxuanzhang: Maybe rename this to setdefaults is more clear, format is not so accurate here

pipeline/trusted-resources/cmd/sign/main.go
Yongxuanzhang: OK! I will add a todo here

pipeline/trusted-resources/cmd/sign/main.go
Yongxuanzhang: yes! (I didn't do this because I want to avoid more unit tests. No I'm just kidding. 😄 )

pipeline/trusted-resources/cmd/sign/main.go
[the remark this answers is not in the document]
Yongxuanzhang: Sure! I changed the commit messages and pr description

wlynch (approved):
/lgtm

PR #872, opened 2022-04-27
Create common release resources for custom tasks

### LINKED ISSUE #805: Demonstrate best practices via experimental project release configuration
# Expected Behavior

In our dogfooding of Tekton we should try to set an example for best use of Tekton, which should include Pipelines and Tasks supporting our experimental releases.

For our nightly experimental release this should be:
* One set of Pipelines and Tasks which can be reused for each experimental custom task project, with the only thing needing to vary being parameters to these Pipelines and Tasks
* Even better would be to use the same Pipelines and Tasks used to release other Tekton projects vs. creating specific ones for experimental projects
* Even better than that would be that we publish these as Tekton Bundles so we can refer to them and use them without needing to apply them to the cluster first

# Actual Behavior

Release tasks and pipelines are copy pasted between each experimental task, which is unnecessary since they all need the same logic but with different parameters, and also leads to errors getting copy pasted between projects.

# Steps to Reproduce the Problem

Examples of errors propagated via copy paste:

* https://github.com/tektoncd/experimental/pull/803#discussion_r731310349
* https://github.com/tektoncd/experimental/pull/803#discussion_r731320517

lbernick:
# Changes
This commit moves the release pipeline and task that are copy-pasted in many places in experimental/
to a common location. Individual projects in experimental should be migrated to use this Pipeline and Task
with appropriate parameter values, instead of continuing to copy paste these definitions in multiple places.
This will be used first for the pipeline-in-pod project (the ColocatedPipelineRun custom task).

# Submitter Checklist

review them:

- n/a Commit messages includes a project tag in the subject - e.g. [OCI], [hub], [results], [task-loops]

for more details._

Part 1 of #805

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -1,15 +1,17 @@
-TODO(#805): Migrate to use common release pipeline
-
 # Tekton Resources
 
-This folder contains tekton resources used build and release the Task Loops custom task.
+This folder contains tekton resources used build and release custom tasks.
 The resources included in `kustomization.yaml` are installed nightly on the `dogfooding`
 cluster, and used to run nightly releases.
+
+Individual projects will be updated to use these common release resources
+([#805](https://github.com/tektoncd/experimental/issues/805)). Please don't
+add more copy-pasted release Pipelines for releasing Custom Tasks.
 
 ## Nightly releases
 
 [The release pipeline](release-pipeline.yaml) is
-[triggered nightly by Tekton](https://github.com/tektoncd/plumbing/tree/master/tekton/resources/nightly-release).
+[triggered nightly by Tekton](https://github.com/tektoncd/plumbing/tree/main/tekton/resources/nightly-release).
 
 This Pipeline uses:
 
--- a/tekton/publish-customtask.yaml
+++ b/tekton/publish-customtask.yaml
@@ -1,0 +1,187 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: publish-release
+spec:
+  params:
+  - name: package
+    description: package to release (e.g. github.com/<org>/<project>
+  - name: subfolder
+    description: folder within the package for which to publish a release
+  - name: versionTag
+    description: The vX.Y.Z version that the artifacts should be tagged with (including `v`)
+  - name: imageRegistry
+    description: The target image registry
+  - name: imageRegistryPath
+    description: The path (project) in the image registry
+  - name: imageRegistryRegions
+    description: The target image registry regions
+    default: "us eu asia"
+  - name: releaseAsLatest
+    description: Whether to tag and publish this release as Pipelines' latest
+  - name: platforms
+    description: Platforms to publish for the images (e.g. linux/amd64,linux/arm64)
+    default: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
+  - name: serviceAccountPath
+    description: The name of the service account path within the release-secret workspace
+  workspaces:
+    - name: source
+      description: The workspace where the repo has been cloned
+    - name: release-secret
+      description: The secret that contains a service account authorized to push to the imageRegistry and to the output bucket
+    - name: output
+      description: The release YAML will be written to this workspace
+  stepTemplate:
+    env:
+      - name: "GIT_ROOT"
+        value: "/workspace/go/src/$(params.package)"
+      - name: "PROJECT_ROOT"
+        value: "/workspace/go/src/$(params.package)/$(params.subfolder)"
+      - name: CONTAINER_REGISTRY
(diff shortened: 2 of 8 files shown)


lbernick:
/assign afrittoli

This will probably need https://github.com/tektoncd/plumbing/pull/1073-- I'm guessing the best way to do this would be to add the new /tekton folder first, then merge the changes to plumbing, then remove each project's individual /tekton folder?

afrittoli:
Thanks for starting this @lbernick - that's something I've had in the back of my mind for some time now, but didn't have time to look for a proper solution. Some of the projects in the experimental are custom task controllers, some are generic controllers and some are something else entirely, which means their publish task and release pipeline may look different.

For custom task specifically, have a template for releasing them would be useful also as part of https://github.com/tektoncd/community/blob/main/teps/0071-custom-task-sdk.md

A few issues on this PR:
- some projects have a controller and a webhook, some only a controller image, and this is not reflected in this change, so for instance task-loop will be broken by this
- some projects have special logic in case of missing "kodata" folders - we should include that in the generic task

More comments:
-  a generic tekton folder might be confusing as it does not apply to all projects, perhaps we could at least add some comments to the README to describe which projects are using this and what are the requirements for a projects to be able to use these release resources
- we could have a small tekton/README in each project to mention that they use the shared resources for release
- one small downside of with this approach is that owners of the project don't own anymore the release resources. This is already true however for shared resources in plumbing so perhaps it's ok
- if the publish task can be made generic enough, we could eventually get it in the catalog

lbernick:
> Thanks for starting this @lbernick - that's something I've had in the back of my mind for some time now, but didn't have time to look for a proper solution. Some of the projects in the experimental are custom task controllers, some are generic controllers and some are something else entirely, which means their publish task and release pipeline may look different.
> 

Got it-- I've renamed the files in /tekton to indicate that this release Task and Pipeline are specifically for releasing a custom task.

> * some projects have a controller and a webhook, some only a controller image, and this is not reflected in this change, so for instance task-loop will be broken by this

I updated the koparse step in the Task to loop over cmd/* instead of ["webhook" "controller"] if that's what you're referring to

> * some projects have special logic in case of missing "kodata" folders - we should include that in the generic task

This should be in there now

> * a generic tekton folder might be confusing as it does not apply to all projects, perhaps we could at least add some comments to the README to describe which projects are using this and what are the requirements for a projects to be able to use these release resources

Updated the readme to indicate that this is just for custom Tasks

> * we could have a small tekton/README in each project to mention that they use the shared resources for release

Updated each project's readme to link to the TODO of swapping them to the new release pipeline. This way we avoid breaking existing projects, and just test it out with ColocatedPipelineRun.

> * one small downside of with this approach is that owners of the project don't own anymore the release resources. This is already true however for shared resources in plumbing so perhaps it's ok

I think this is probably OK! I like the idea of a common Pipeline and Task for releasing custom tasks and each project can have its own TaskRun and PipelineRun or something.

> * if the publish task can be made generic enough, we could eventually get it in the catalog

SGTM! :)

lbernick:
@afrittoli @dibyom @jerop I'd like to use this for release automation for the ColocatedPipelineRun custom task, so I'd like to get it moving forward, mind taking a look?

afrittoli (approved):
Thank you Lee!

tekton/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+- [publish](publish.yaml)
+- [git-clone](https://hub.tekton.dev/tekton/task/git-clone)
+- [gcs-upload](https://hub.tekton.dev/tekton/task/gcs-upload) 
+- [golang-build](https://hub.tekton.dev/tekton/task/golang-build)
+- [golang-test](https://hub.tekton.dev/tekton/task/golang-test)
+
+### Service account and secrets
+
+In order to release, these Pipelines expects a service account JSON file to be
```
dibyom, line 26: Hmmm...at some point we should move these to workload identity. Maybe we should have an issue to track that?

tekton/publish-customtask.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+      # Change to directory with our .ko.yaml
+      cd ${PROJECT_ROOT}
+
+      # For each cmd/* directory, include a full gzipped tar of all source in
+      # vendor/. This is overkill. Some deps' licenses require the source to be
+      # included in the container image when they're used as a dependency.
+      # Rather than trying to determine which deps have this requirement (and
+      # probably get it wrong), we'll just targz up the whole vendor tree and
+      # include it. As of 9/20/2019, this amounts to about 11MB of additional
```
dibyom, line 109: The last sentence is almost certainly inaccurate - it was true for the pipelines repo at some point. I think we should just remove it

dibyom (approved):
One one minor thing involving removing a comment which I think can happen in a followup. So,

/lgtm

PR #876, opened 2022-05-05
Fix typos and for loop in CustomTask release Task

lbernick:
# Changes

This commit fixes a typo in the envvar "CONTAINER_REGISTRY_CREDENTIALS"
and fixes a bash for loop so that it does not evaluate to the literal string "cmd/*".

# Submitter Checklist

review them:

- n/a Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- n/a Commit messages includes a project tag in the subject - e.g. [OCI], [hub], [results], [task-loops]

for more details._


tekton/publish-customtask.yaml
```diff
@@ -140,9 +140,8 @@ spec:
         IMAGES_PATH=${IMAGES_PATH}/$(params.subfolder)
       fi
 
-      for d in cmd/*
-      do
-        IMAGES="${IMAGES} ${IMAGES_PATH}/${d}:$(params.versionTag)"
+      for d in controller; do
+        IMAGES="${IMAGES} ${IMAGES_PATH}/cmd/${d}:$(params.versionTag)"
       done
```
afrittoli, line 145: Some custom tasks define their own custom resource and thus have a webhook, so I don't think this will work in general.

tekton/publish-customtask.yaml
```diff
@@ -140,9 +140,8 @@ spec:
         IMAGES_PATH=${IMAGES_PATH}/$(params.subfolder)
       fi
 
-      for d in cmd/*
-      do
-        IMAGES="${IMAGES} ${IMAGES_PATH}/${d}:$(params.versionTag)"
+      for d in ${PROJECT_ROOT}/cmd/*; do
```
lbernick, line 143: I've tested this out on my cluster and it works. The issue was that if there is a `cmd` folder in the current directory, `cmd/*` is interpreted as a list, but otherwise it's interpreted as a literal string.

afrittoli (commented):
/approve
/lgtm

PR #880, opened 2022-05-13
[pipeline-in-pod] Support parallel tasks

lbernick:
# Changes
This commit adds support for parallel tasks in the same pod. It writes step output to
/tekton/run/<pipeline_task_name>/<step_number> and supports multiple waitFiles per container.
It also pins the image versions for the entrypoint binary and git-init, to avoid breaking the controller.

# Submitter Checklist

review them:

for more details._


pipeline-in-pod/pkg/reconciler/pipelineinpod/entrypoint.go
```diff
@@ -165,14 +183,54 @@ func orderContainers(commonExtraEntrypointArgs []string, ptcs []pipelineTaskCont
...
+}
+
+func getVolumesForStep(ptc pipelineTaskContainers, i int, runAfter []pipelineTaskContainers) ([]corev1.Volume, []corev1.VolumeMount) {
+	var volumes []corev1.Volume
+	s := ptc.containers[i]
+
+	// TODO (maybe) creds-init
+
+	// Add /tekton/run state volumes.
+	// Each step should only mount their own volume as RW,
```
dibyom, line 205: Nice!

dibyom:
Nice!
/approve

Just making sure I understand the volumeMount logic:
1. Each step mounts each volume for that pipeline task - volumes for other steps are mounted as RO while volumes for that particular step gets mounted as RW
2. Volumes from other pipeline tasks are not mounted at all so should not be visible to the step

Question: Would it be possible to share the same `emptyDir` or similar non persistent volume across pipelineTasks somehow with this implementation?

lbernick:
> Just making sure I understand the volumeMount logic:
> 
> 1. Each step mounts each volume for that pipeline task - volumes for other steps are mounted as RO while volumes for that particular step gets mounted as RW

Yes

> 2. Volumes from other pipeline tasks are not mounted at all so should not be visible to the step

Not exactly; the first step in each task also mounts volumes for the last steps of any other tasks it has a `runAfter` dependency on, so it can know when those steps have completed.

Unclear what the security implications are here-- we'd probably have to prevent non tekton code from writing to the /tekton/run volumes somehow? but to some extent the previous tasks need to communicate in some way that they have finished

> Question: Would it be possible to share the same `emptyDir` or similar non persistent volume across pipelineTasks somehow with this implementation?

I'm not sure exactly what you mean by this, are you talking about a workspace shared among pipeline tasks or are you talking about using one volume for all of the /tekton/run data or other tekton-internal volumes that are mounted?

lbernick:
> Just making sure I understand the volumeMount logic:
> 
> 1. Each step mounts each volume for that pipeline task - volumes for other steps are mounted as RO while volumes for that particular step gets mounted as RW

Yes

> 2. Volumes from other pipeline tasks are not mounted at all so should not be visible to the step

Not exactly; the first step in each task also mounts volumes for the last steps of any other tasks it has a `runAfter` dependency on, so it can know when those steps have completed.

Unclear what the security implications are here-- we'd probably have to prevent non tekton code from writing to the /tekton/run volumes somehow? but to some extent the previous tasks need to communicate in some way that they have finished

> Question: Would it be possible to share the same `emptyDir` or similar non persistent volume across pipelineTasks somehow with this implementation?

I'm not sure exactly what you mean by this, are you talking about a workspace shared among pipeline tasks or are you talking about using one volume for all of the /tekton/run data or other tekton-internal volumes that are mounted?

dibyom:
> Not exactly; the first step in each task also mounts volumes for the last steps of any other tasks it has a runAfter dependency on, so it can know when those steps have completed.

Got it, that makes sense. I agree re: the security implications. 

> I'm not sure exactly what you mean by this, are you talking about a workspace shared among pipeline tasks or are you talking about using one volume for all of the /tekton/run data or other tekton-internal volumes that are mounted?

The former - I'd like a workspace that I can use throughout the pipeline.

lbernick:
> > I'm not sure exactly what you mean by this, are you talking about a workspace shared among pipeline tasks or are you talking about using one volume for all of the /tekton/run data or other tekton-internal volumes that are mounted?
> 
> The former - I'd like a workspace that I can use throughout the pipeline.

This PR should only affect the /tekton/run volume mounts, not workspace volumes. This custom task should already support a workspace available for the whole pipeline.

lbernick:
@jerop @dibyom @vdemeester @afrittoli if any of you are able to LGTM that would unblock me for some work I'm trying to get done on this project! 🙏

vdemeester (approved):
/lgtm

Issue #873, opened 2022-04-29
[Trusted Task] Verify task and pipeline when apply them


Yongxuanzhang:
Right now we verify task and pipeline in taskrun and pipelinerun, it would be better if we can verify them when applying these resources. This can help users check if the resources they apply are valid. And also for each UPDATE it can also help to verify the resources.

Yongxuanzhang:
Yep! The key points of this issue is to understand the code in 
https://github.com/tektoncd/experimental/blob/main/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
https://github.com/tektoncd/experimental/blob/main/pipeline/trusted-resources/pkg/trustedtask/trustedpipelinerun.go
In these two files we verify the task and pipeline from taskrun and pipelinerun. 

This issue want to have 2 similar files: trustedtask.go and trustedpipeline.go
and then create TrustedTask and TrustedPipeline, to do similar work for verification.

austinzhao-go:
merged impls for `Task` and `Pipeline` verification.

here related PRs
https://github.com/tektoncd/experimental/pull/886
https://github.com/tektoncd/experimental/pull/888

thanks @Yongxuanzhang's guidance on this work. closed this issue at this moment.

PR #886, opened 2022-05-31
[pipeline/trusted-resources] Verify Task and Pipeline for Trusted Task

austinzhao-go:
# Changes
The related issue: https://github.com/tektoncd/experimental/issues/873

Add `Task` and `Pipeline` Verification in Trusted Task repo
 

# TODO
- [x] Add tests for "Verify Task"
- [x] Add impl for "Verify Pipeline"

# Submitter Checklist

review them:

- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing) - Docs will follow with current user guides

for more details._


pipeline/trusted-resources/pkg/trustedtask/trustedpipeline.go
```diff
@@ -0,0 +1,66 @@
...
+func (p *TrustedPipeline) Validate(ctx context.Context) (errs *apis.FieldError) {
+	k8sClient := kubeclient.Get(ctx)
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return apis.ErrGeneric(err.Error())
+	}
+	tektonClient, err := versioned.NewForConfig(config)
+	if err != nil {
+		return apis.ErrGeneric(err.Error())
+	}
```
wlynch, line 29: If we're using injection clients for k8s, should we do the same for Tekton? https://pkg.go.dev/github.com/tektoncd/pipeline@v0.36.0/pkg/client/injection/client

pipeline/trusted-resources/pkg/trustedtask/trustedtask_test.go
```diff
@@ -0,0 +1,164 @@
...
+func init() {
+	os.Setenv("SYSTEM_NAMESPACE", nameSpace)
+	os.Setenv("WEBHOOK_SERVICEACCOUNT_NAME", serviceAccount)
+}
+
+func Test_verifyTask(t *testing.T) {
+	ctx := logging.WithLogger(context.Background(), zaptest.NewLogger(t).Sugar())
+	k8sClient := fakek8s.NewSimpleClientset(sa)
+	tektonClient := faketekton.NewSimpleClientset()
+	signer, secretpath, _ := getSignerFromFile(t, ctx, k8sClient)
```
wlynch, line 28: It looks like we're ignoring the error everytime. Looking at the impl it looks like we never return an error, so can we removing this from the return vals?

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
```diff
@@ -0,0 +1,86 @@
...
+	if !ok {
+		return nil, fmt.Errorf("signature is missing")
+	}
+
+	signature, err := base64.StdEncoding.DecodeString(sig)
+	if err != nil {
+		return nil, err
+	}
+
+	delete(metadata.Annotations, SignatureAnnotation)
```
wlynch, line 83: This is going to modify the incoming object. Is this intentional? If so update the godoc, else let's remove this.

wlynch (changes requested):
Looks good for the most part! Just a few small things.

pipeline/trusted-resources/pkg/trustedtask/trustedpipeline.go
austinzhao-go: didn't realize this diff. thanks! 
let me update with `client.Get()`.

pipeline/trusted-resources/pkg/trustedtask/trustedtask_test.go
austinzhao-go: good catch! think perhaps I can add the err on the return "load", so populating the errors if any out.

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
austinzhao-go: bad me. think I should add a comment for this line - "remove the signature to compute the 'expected signature over the original/unsigned task"

refactored from these line: https://github.com/austinzhao-go/experimental/blob/feature/verify-task-pipeline-when-applying/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go#L165 

but think now I found the problem - this delete will permanently delete the sig from original task, and  I should use a copy to do this deletion and move for the verification.

pipeline/trusted-resources/pkg/trustedtask/trustedtask_test.go
austinzhao-go: considering this helper function is under a test file. the error could be thrown in the func, so no need to pass to the return vals. 

agree with the original feedback. updated to remove the error in a follow-up PR
https://github.com/tektoncd/experimental/pull/888

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
```diff
@@ -0,0 +1,82 @@
...
+func (t *TrustedTask) Validate(ctx context.Context) (errs *apis.FieldError) {
+	k8sClient := kubeclient.Get(ctx)
+	tektonClient := client.Get(ctx)
+
+	if err := t.verifyTask(ctx, k8sClient, tektonClient); err != nil {
+		return err
+	}
+
+	return nil
+}
```
wlynch, line 32: Non blocking for this PR, but this is pretty trivial now - it may make sense to collapse this down to 1 func.

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
```diff
@@ -0,0 +1,82 @@
...
+	k8sClient kubernetes.Interface,
+	tektonClient versioned.Interface) (errs *apis.FieldError) {
+	logger := logging.FromContext(ctx)
+	logger.Info("Verifying Task")
+
+	task := t.Task
+
+	cfg := config.FromContextOrDefaults(ctx)
+	cfg.FeatureFlags.EnableTektonOCIBundles = true
+	ctx = config.ToContext(ctx, cfg)
```
wlynch, line 45: Ideally it'd be great if we didn't need to inject this into the context, but if this is always going to be set for every type, consider setting this for the whole controller here - https://github.com/tektoncd/experimental/blob/b94e00f1de513635d831c4ce239cd7c8f171ac60/pipeline/trusted-resources/cmd/webhook_trusted_resource/main.go#L65

wlynch (approved):
/lgtm

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
austinzhao-go: thanks Billy. so here mean we could arrange `verifyTask` into `Validate`, so having a single func right?

pipeline/trusted-resources/pkg/trustedtask/trustedtask.go
austinzhao-go: didn't realize this. thanks this hint! 

mark this as a follow up action. see if need certain flexibility or not here for following work. if always, will update in webhook. 
cc @Yongxuanzhang

Issue #870, opened 2022-04-21
[trusted-task] Add config for users to determine


Yongxuanzhang:
As a user of trusted-task admission webhook, we want to have a config to control whether we want to turn on or turn off the verification.
This may include several levels:
- Fail the taskrun if verification failed
- (Optional)Log in webhook and mark the taskrun failed verification(maybe in annotation) 
- Skip the verification

PR #887, opened 2022-06-02
[pipeline/trusted-resources]Add config option to skip TaskRun verification

Aleromerog:
# Changes
Add  optional config for skip TaskRun verification in trusted TaskRun to let the user
determine  verification.

Fix: #870 

# Submitter Checklist

review them:

- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [X] Commit messages includes a project tag in the subject - e.g. [OCI], [hub], [results], [task-loops]

for more details._

### DIFF
--- a/pipeline/trusted-resources/README.md
+++ b/pipeline/trusted-resources/README.md
@@ -118,4 +118,6 @@ metadata:
     app.kubernetes.io/part-of: admissioncontrol
 data:
   cosign-pubkey-path: "/etc/signing-secrets/cosign.pub"
+  pass-taskrun-when-fail-verification: true
+
 ```
--- a/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
+++ b/pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 
+	trustedconfig "github.com/tektoncd/experimental/pipelines/trusted-resources/pkg/config"
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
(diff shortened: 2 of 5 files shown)


pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -95,16 +102,19 @@ func (tr *TrustedTaskRun) verifyTaskRun(
...
 
 	// TODO: figure out how to config this.
+
 	cfg := config.FromContextOrDefaults(ctx)
 	cfg.FeatureFlags.EnableTektonOCIBundles = true
 	ctx = config.ToContext(ctx, cfg)
 
 	fn, err := resources.GetTaskFunc(ctx, k8sclient, tektonClient, tr.Spec.TaskRef, tr.Namespace, serviceAccountName)
 	if err != nil {
+		logger.Error(err.Error())
```
Yongxuanzhang, line 112: It should be ok if we don't log here, the returned errs will be logged eventually.

pipeline/trusted-resources/pkg/trustedtask/trustedtaskrun.go
```diff
@@ -81,7 +82,13 @@ func (tr *TrustedTaskRun) verifyTaskRun(
 	k8sclient kubernetes.Interface,
 	tektonClient versioned.Interface,
 ) (errs *apis.FieldError) {
+	trustedConfig := trustedconfig.FromContextOrDefaults(ctx)
 	logger := logging.FromContext(ctx)
+	if trustedConfig.SkipValidation {
+		logger.Info("Skipping TaskRun verification.")
+		return nil
+	}
```
Yongxuanzhang, line 90: Sorry my bad. I realized that it may doesn't make sense to directly skip it, since we can remove the webhook to do the same thing. (But configmap is more convenient though)

Maybe we should support these:
1) Config is true: If fail the verification, we add one field in annotation, logger the error information and return nil.
2) Config is false: Return err.

Yongxuanzhang (commented):
Thanks for working on this issue! Maybe we can discuss later offline about the direction of this issue, it is my fault that haven't thought about it throughly

pipeline/trusted-resources/README.md
```diff
@@ -118,4 +118,6 @@ metadata:
     app.kubernetes.io/part-of: admissioncontrol
 data:
   cosign-pubkey-path: "/etc/signing-secrets/cosign.pub"
+  skip-taskrun-validation: true
```
Yongxuanzhang, line 121: Just realized this is not updated

Aleromerog: Nice catch! just updated it

Yongxuanzhang (approved):

PR #905, opened 2022-10-03
[concurrency] Implement webhook and reconciler

lbernick:
# Changes
This commit adds a mutating admission webhook and reconciler for canceling concurrent PipelineRuns. The mutating admission webhook patches all PipelineRuns on creation with a Pending status. The reconciler handles only Pending PipelineRuns. It queries for all ConcurrencyControls in the same namespace, cancels any other matching PipelineRuns, and starts the PipelineRun it's reconciling.

# Submitter Checklist

review them:

for more details._


dibyom (commented):
Need to look at the reonciler code a bit more carefully to ensure that 1. if we are making multiple update calls for the pr and 2. if there are multiple calls, the second one does not overwrite the first one

But otherwise, this looks great!!

dibyom:
> Need to look at the reonciler code a bit more carefully to ensure that 1. if we are making multiple update calls for the pr and 2. if there are multiple calls, the second one does not overwrite the first one

I think we should set the SkipStatusUpdates Reconciler option that disables Knative from updating the PipellineRun: https://pkg.go.dev/knative.dev/pkg@v0.0.0-20221005164529-bc93b0add9a6/controller#Options

We don't really modify the pipelinerun status in this reconciler so we should not run into the issue that we run into in pipelines [issue](https://github.com/tektoncd/pipeline/issues/5146) due to multiple updates but this should still prevent the extra logic for status updates from even running just in case we do update the status down the road

dibyom (approved):

lbernick:
> I think we should set the SkipStatusUpdates Reconciler option that disables Knative from updating the PipellineRun: https://pkg.go.dev/knative.dev/pkg@v0.0.0-20221005164529-bc93b0add9a6/controller#Options
> 
> We don't really modify the pipelinerun status in this reconciler so we should not run into the issue that we run into in pipelines due to multiple updates but this should still prevent the extra logic for status updates from even running just in case we do update the status down the road

Good to know! Done.

PR #912, opened 2022-10-13
[concurrency] Merge mutating admission webhooks

lbernick:
# Changes
With two separate mutating admission webhooks (one for concurrencycontrol defaults and one for patching pipelineruns as pending), one seems to be ignored. This commit merges them into a single mutating admission webhook that applies both sets of configurations.

# Submitter Checklist

review them:

- n/a Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/concurrency/cmd/webhook/main.go
+++ b/concurrency/cmd/webhook/main.go
@@ -83,8 +83,12 @@ func newValidationAdmissionController(ctx context.Context, cmw configmap.Watcher
 	)
 }
 
-func newConcurrencyControlDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+// TODO: This sets defaults based on PipelineRun.SetDefaults imported from Pipelines.
+// We will need to write our own admission webhook that only uses the callback
+// defined above.
+func newDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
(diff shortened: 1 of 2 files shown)


concurrency/cmd/webhook/main.go
```diff
@@ -83,45 +83,21 @@ func newValidationAdmissionController(ctx context.Context, cmw configmap.Watcher
...
 }
 
-func newConcurrencyControlDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
-	var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
-		concurrencyControlKind: &v1alpha1.ConcurrencyControl{},
-	}
-	return defaulting.NewAdmissionController(ctx,
-
-		// Name of the resource webhook.
-		"webhook.concurrency.custom.tekton.dev",
```
dibyom, line 93: Is this because the 2 webhooks have the same name i.e. this one and the the below pr defaulting one? Can we try renaming one of them to something else (with the corresponding new yaml for the new webhook)?

lbernick: I'm not sure what the reason for the error is. Dibyo and I chatted offline-- decided to merge as there isn't really a reason to prefer 2 webhooks

PR #913, opened 2022-10-13
[concurrency] Add namespace allowlist

lbernick:
# Changes
This commit adds the ability to configure the behavior of the concurrency controller and webhook via a new configmap "concurrency-config" in the "tekton-concurrency" namespace. It allows the user to specify a list of namespaces where the concurrency controller and webhook should be allowed to modify PipelineRuns, and defaults to all namespaces.

Alternatively, a NamespaceSelector could be applied to the MutatingAdmissionWebhook that patches PipelineRuns as pending. However, this would not prevent the concurrency controller from operating on PipelineRuns in other namespaces.

/cc @afrittoli 

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/concurrency/README.md
+++ b/concurrency/README.md
@@ -37,3 +37,9 @@ Supported strategies are "Cancel", "GracefullyCancel", and "GracefullyStop"
 (corresponding to canceling, gracefully canceling, and gracefully stopping a PipelineRun, respectively).
 The default strategy is "GracefullyCancel".
 If multiple ConcurrencyControls with different strategies apply to the same PipelineRun, concurrency controls will fail.
+
+### Configuration
+
+To restrict the concurrency webhook and controller to only modify PipelineRuns in a subset of namespaces,
+edit the "allowed-namespaces" field of the [concurrency configMap](./config/concurrency-config.yaml) to be
+a comma-separated list of these namespaces, for example: `allowed-namespaces: "default,another-namespace"`.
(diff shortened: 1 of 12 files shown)


concurrency/README.md
```diff
@@ -36,4 +36,9 @@ The first PipelineRun should be canceled, and the second one should execute norm
...
 (corresponding to canceling, gracefully canceling, and gracefully stopping a PipelineRun, respectively).
 The default strategy is "GracefullyCancel".
-If multiple ConcurrencyControls with different strategies apply to the same PipelineRun, concurrency controls will fail.
\ No newline at end of file
+If multiple ConcurrencyControls with different strategies apply to the same PipelineRun, concurrency controls will fail.
+
+### Configuration
+
+To restrict the concurrency webhook and controller to only modify PipelineRuns in a subset of namespaces,
+edit the "allowed-namespaces" field of the [concurrency configMap](./config/concurrency-config.yaml).
```
jerop, line 44: do you mind noting that the list of namespaces is a comma-separated string with all namespaces

would this be a list of strings instead of one string with commas?

jerop (approved):

concurrency/README.md
lbernick: Updated the readme. For a list instead of a comma separated string-- based on [this example](https://kubernetes.io/docs/concepts/configuration/configmap/#configmaps-and-pods), it seems like a comma separated string is the way to specify a list in a configmap, and it's also not clear that one way is easier than the other

